PR #7, opened 2024-05-22
feat(ws): initial commit for backend

ederign:
Scaffolding of the backend. 

In this PR:
- basic Dockerfile
- basic Makefile
- Scaffold of App and first sample endpoint (http://localhost:4000/api/v1/healthcheck/)
- REST API basic infrastructure and error handling


workspaces/backend/integrations/k8s.go
```diff
@@ -0,0 +1,104 @@
...
+	case config.Production:
+		//probably fetch from Bearer Token
+		return nil, errors.New(fmt.Sprintf("%s is not implemented yet", config.Production))
+	default:
+		return nil, errors.New("environment is not defined")
+	}
+	certData, err := base64.StdEncoding.DecodeString(clientCert)
+	if err != nil {
+		log.Fatalf("Failed to decode client certificate: %v", err)
+	}
```
thesuperzapper, line 67: Lets remove the flag for "dev" vs "production" vs "staging".

The go library provides a number of hard-coded ways to retrieve credneitals from:

- Inside Cluster: https://github.com/kubernetes/client-go/blob/master/examples/in-cluster-client-configuration/main.go#L41
- Outside Cluster: https://github.com/kubernetes/client-go/blob/master/examples/out-of-cluster-client-configuration/main.go#L45


I think the client auth flow should be:

1. Try default location (see https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/doc.go)
1. If that fails, assume we are in cluser and use `rest.InClusterConfig()`
2. If that fails, crash.

thesuperzapper (changes requested):

workspaces/backend/go.mod
```diff
@@ -0,0 +1,35 @@
+module github.com/kubeflow/notebooks/workspaces/backend
+
+go 1.22.2
+
+require (
+	github.com/julienschmidt/httprouter v1.3.0
```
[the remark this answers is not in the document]
ederign: I'm open to exploring new routers if necessary, but I'm based on this flow chart here: https://www.alexedwards.net/blog/which-go-router-should-i-use

workspaces/backend/integrations/k8s.go
ederign: I've removed all code and move this to another issue.

workspaces/backend/go.mod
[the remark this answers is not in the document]
jiridanek: The package does not have transitive dependencies! I think it is not very dangerous to use it https://github.com/julienschmidt/httprouter/blob/master/go.mod

ederign:
@thesuperzapper @jiridanek I've updated my PR as we discussed.

I'll open a new issue to discuss the k8 client.

.gitignore
```diff
@@ -37,4 +37,7 @@ Sessionx.vim
 .DS_Store
 .AppleDouble
 .LSOverride
-._*
\ No newline at end of file
+._*
+
+## backend
+workspaces/backend/bin
```
jiridanek, line 43: maybe create a separate .gitignore for this in workspaces/backend?

workspaces/backend/Makefile
```diff
@@ -0,0 +1,33 @@
+CONTAINER_TOOL ?= docker
+IMG ?= nbv2-backend:latest
+
+.PHONY: all
+all: build
+
+.PHONY: help
+help: ## Display this help.
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
```
jiridanek, line 9: Just fyi, make version 4.4.1 will natively support a new `--print-targets` option, which makes this solution unnecessary, if you can have new enough make

jiridanek (approved):

.gitignore
ederign: We had this, but @thesuperzapper asked to remove it on the last review.

jiridanek: I understand he asked to just remove the last few lines from that (because they would duplicate content of the top level .gitignore)

thesuperzapper (approved):
@ederign thanks, lets merge this and work from here!

PR #10, opened 2024-05-24
ci(ws): add integration test that deploys workspace-controller onto kind cluster

jiridanek:
This is a recreation of https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/nb_controller_intergration_test.yaml.

The main difference is that I don't (yet) do multiarch dockerx build.

## Future work

Next step, would be helpful to wait not only for controller readiness, but for some sample CRD instances to have been reconciled by it. Ofc doing multiarch build early would not be bad either.

Having a similar test that uses OpenShift Local to run and podman to build would not go amiss either, in my view. This would force me to investigate the SCC thing https://docs.openshift.com/container-platform/4.15/authentication/managing-security-context-constraints.html

### DIFF
--- a/.github/workflows/workspaces-controller_integration-tests.yaml
+++ b/.github/workflows/workspaces-controller_integration-tests.yaml
@@ -22,41 +22,20 @@ jobs:
 
       - uses: actions/checkout@v4
 
-      # https://docs.k0sproject.io/v1.30.0+k0s.0/k0s-in-docker/#start-k0s
-      - name: Start k0s
-        run: |
-          docker run -d --name k0s --hostname k0s --privileged -v /mnt/k0s:/var/lib/k0s -p 6443:6443 --cgroupns=host \
-            docker.io/k0sproject/k0s:v1.30.0-k0s.0 -- k0s controller --enable-worker
-
       - name: Build Notebook Controller Image
         run: make docker-build
         env:
           IMG: "${{ matrix.img }}"
         working-directory: workspaces/controller
 
-      # https://docs.k0sproject.io/v1.30.0+k0s.0/airgap-install/#1-create-your-own-image-bundle-optional
-      - name: Deploy image bundle into k0s
-        run: |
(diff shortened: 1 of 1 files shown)


ederign:
@jiridanek, would you mind signing off on the commit?

ederign:
Also, is there a way to see this action running?

jiridanek:
> Also, is there a way to see this action running?

Yeah, click the circle and then Details

image

jiridanek:
Uh, the link to action run, it disappeared from above! So, here you go https://github.com/jiridanek/notebooks-v2/actions/runs/9284952173/job/25548433943

.github/workflows/workspaces-controller_integration-tests.yaml
```diff
@@ -0,0 +1,86 @@
+name: "workspaces/controller: integration-tests"
+
+permissions: {}
+
+on:
+  push:
+    branches: [ "main", "notebooks-v2", "v*-branch" ]
+  pull_request:
+    paths: [ "workspaces/controller/**" ]
```
jiridanek, line 9: right, so the action runs on PR only if the files here change, but not when the action itself changes!

ederign:
> > Also, is there a way to see this action running?
> 
> Yeah, click the circle and then Details
> 
> image

It's probably a permission issue, but I cannot see the action of the check run here. I checked on your fork.

/lgtm

jiridanek:
As per @thesuperzapper 's suggestion, I switched from k0s to using kind, which made the GHA yaml somewhat shorter.

Run can be seen in action at https://github.com/jiridanek/notebooks-v2/actions/runs/9602575963/job/26483718596

thesuperzapper:
@jiridanek I am going to close this in favor of https://github.com/kubeflow/notebooks/pull/11.

After https://github.com/kubeflow/notebooks/pull/30, the `make test-e2e` effectively does the same thing as this PR, so would be redundant.

/close

PR #11, opened 2024-05-24
ci(ws): run `make test-e2e` on GitHub Actions

jiridanek:
This is something that works differently in new Kubebuilder, compared to what's in kubeflow/kubeflow notebook controller.

Previously, what they called e2e tests was running in envtest, but currently they expect to run in externally-provided kind. So, let's run this on CI as well, so we don't have dead code.

### DIFF
--- a/.github/workflows/workspaces-controller_unit-tests.yaml
+++ b/.github/workflows/workspaces-controller_unit-tests.yaml
@@ -36,6 +36,8 @@ jobs:
 
       - uses: helm/kind-action@v1
         with:
+          version: v0.23.0
+          node_image: kindest/node:v1.25.16@sha256:5da57dfc290ac3599e775e63b8b6c49c0c85d3fec771cd7d55b45fae14b38d3b
           cluster_name: kind
 
       - name: Set up Go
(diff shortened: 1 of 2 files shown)


ederign (approved):

.github/workflows/workspaces-controller_unit-tests.yaml
```diff
@@ -27,3 +27,23 @@ jobs:
...
         working-directory: workspaces/controller
+
+  "e2e-tests":
+    runs-on: ubuntu-latest
+    steps:
+
+      - uses: actions/checkout@v4
+
+      - uses: helm/kind-action@v1
+        with:
```
jiridanek, line 38: specify kind version, 1.25; 1.28

jiridanek:
As per @thesuperzapper 's suggestion, I put in explicit kind version. Here's the CI running https://github.com/jiridanek/notebooks-v2/actions/runs/9602645980/job/26483942380

thesuperzapper:
@jiridanek now that we have actual end-to-end tests (added in https://github.com/kubeflow/notebooks/pull/30), I have rebased this PR and cleaned it up a little in https://github.com/kubeflow/notebooks/pull/11/commits/3e8beea2bc92c46064ad22c84744ef51a7fb0e78 and  https://github.com/kubeflow/notebooks/pull/11/commits/74089217f7eae491e1120c6bb1db03ea05d77f36.

We should be good to merge now, so that tests run automatically on every PR for the controller.

/lgtm
/approve

Issue #12, opened 2024-05-29
How to init the Kubernetes client for the workspace backend


ederign:
### Checks

- [X] I have searched the [existing issues](https://github.com/kubeflow/notebooks/issues).

### Motivation

The goal of this issue is to develop: 
- backend k8s client (wrapper of kubernetes/client-go/) 

### Implementation

```
func NewKubernetesClient() (*KubernetesClient, error) {

	clientSet, err := newClientSet()
	if err != nil {
		return nil, fmt.Errorf("failed to create Kubernetes clientset: %w", err)
	}

	return &KubernetesClient{ClientSet: clientSet}, nil
}

func getRestConfig() (*restclient.Config, error) {
	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
	configOverrides := &clientcmd.ConfigOverrides{}
	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
	return kubeConfig.ClientConfig()
}

func newClientSet() (*kubernetes.Clientset, error) {
	restConfig, err := getRestConfig()
	if err != nil {
		return nil, err
	}
	return kubernetes.NewForConfig(restConfig)
}

```

### Are you willing & able to help?

- [X] I am able to submit a PR!
- [X] I can help test the feature!

ederign:
FYI: I'm currently working on this. (I don't have 'assign' privileges yet).

ederign:
Inside cluster:

```
> kubectl port-forward service/my-service-nb 8080:80
Forwarding from 127.0.0.1:8080 -> 4001
Forwarding from [::1]:8080 -> 4001
Handling connection for 8080
Handling connection for 8080
```
<img width="493" alt="image" src="https://github.com/kubeflow/notebooks/assets/531351/cc6543c7-cf06-4fe9-9ecf-35435fea95fb">


Outside cluster, you can either use .kube config OR export KUBECONFIG=pathtokubeconfig.yaml

ederign:
Also, I've moved the Kubernetes client  to app startup. Seems like the Kubernetes client is designed to handle long-lived connections and can manage its own timeouts and retries.

However, on future, we can improve it with proper configuration of timeouts and other settings 

```
config.WrapTransport = func(rt http.RoundTripper) http.RoundTripper {
        return &http.Transport{
            Proxy:                 http.ProxyFromEnvironment,
[...]

PR #15, opened 2024-06-05
feat(ws): add kubernetes client to backend

ederign:
closes https://github.com/kubeflow/notebooks/issues/12

- Creation of Kubernetes Client
- Wrapping the client on application


thesuperzapper:
@ederign you might need to force push so the bot rechecks

ederign:
@thesuperzapper done!

workspaces/backend/cmd/main.go
```diff
@@ -34,7 +34,10 @@ func main() {
 
 	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
 
-	app := application.NewApp(cfg, logger)
+	app, err := application.NewApp(cfg, logger)
+	if err != nil {
+		logger.Error(err.Error())
```
thesuperzapper, line 39: This should probably exit with non-zero code rather than continuing.

```suggestion
		logger.Error(err.Error())
		os.Exit(1)
```

workspaces/backend/Dockerfile
```diff
@@ -16,6 +16,7 @@ COPY cmd/ cmd/
 COPY api/ api/
 COPY config/ config/
 COPY data/ data/
+COPY integrations/ integrations/
```
thesuperzapper, line 19: It seems a bit strange to not just copy the entire source folder.

Especially since we are using a build stage, so there is no additional size added to the final container.

If you are worried about copying stuff, just add a `.dockerignore` as we have for the controller:
- https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/controller/.dockerignore

thesuperzapper (changes requested):

workspaces/backend/cmd/main.go
ederign: Yeah that was a mistake.

workspaces/backend/Dockerfile
ederign: Agree. I just tried to follow the same pattern of here: https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/controller/Dockerfile#L15-L17

Works for me in both ways.

ederign: I've also added docker ignore file just in case.

thesuperzapper:
@ederign thanks, lets move forward so we can use this in the next PR.

/approve
/lgtm

PR #21, opened 2024-07-03
fix(ws): use correct `nodeSelector` type and add CRD validation tests

thesuperzapper:

- Uses the correct type for `spec.podTemplate.options.podConfig.values[].spec.nodeSelector`
- Ensures that the `spec` of `podConfig` and `imageConfig` values does not change
- Updates a few types to use the `+listType:="map"` because they should not support multiple elements with the same key:
     - `spec.podTemplate.volumes.data[]` (keyed on `name`)
     - `spec.podTemplate.extraEnv[]` (keyed on `name`)

I have also added tests for some of the more complicated `+kubebuilder:validation:XValidation` / `x-kubernetes-validations` we are doing, so we dont have to worry about regressions in the future.

### DIFF
--- a/workspaces/controller/api/v1beta1/workspace_types.go
+++ b/workspaces/controller/api/v1beta1/workspace_types.go
@@ -41,7 +41,7 @@ type WorkspaceSpec struct {
 	//+kubebuilder:validation:MaxLength:=63
 	//+kubebuilder:validation:Pattern:=^[a-z0-9][-a-z0-9]*[a-z0-9]$
 	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Workspace 'kind' is immutable"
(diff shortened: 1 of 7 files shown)


thesuperzapper:
@jiridanek this should be a simple review so we can unblock the reconciliation work by @Adembc (mainly the issue he had was related to the type of `nodeSelector`.

jiridanek (approved):

jiridanek:
/lgtm

@thesuperzapper sorry about the approve; our (@openshift's) instance of prow understands github approve as a lgtm; guess different configuration

thesuperzapper:
/approve

@Adembc don't forget to rebase your PR on `notebooks-v2` after this merges.

PR #29, opened 2024-07-23
test(ws): add tests for workspace and workspaceKind controller loops

Adembc:
This PR is still a work in progress, but I’d appreciate a review before completing all tests.

The tests covered in this PR include:

- Verifying that resources like Service and StatefulSet are created when a Workspace is created.
- Ensuring that reconciled resources are deleted and properly recreated.
- Checking that updated reconciled resources are reverted as expected.
- Validating that Go templates in WorkspaceKind's spec.podTemplate.extraEnv[].value function correctly:
   -  Succeed with a valid portID.
   -  Return an empty string for an invalid portID.
   -  Set the Workspace to an error state for invalid template formats (e.g., using single quotes for the portID string).

### DIFF
--- a/workspaces/controller/internal/controller/suite_test.go
+++ b/workspaces/controller/internal/controller/suite_test.go
@@ -169,6 +169,11 @@ func NewExampleWorkspace1(name string, namespace string, workspaceKind string) *
 	}
 }
 
+func NewExampleWorkspace2(name string, namespace string, workspaceKind string) *kubefloworgv1beta1.Workspace {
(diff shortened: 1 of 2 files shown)


thesuperzapper:
@Adembc I have not reviewed this PR, but I have rebased it so the tests will run.

andyatmiami:
/close

given the significant time and changes that have occurred since this PR was initially opened - going to close it.

can always re-open/revisit later if we feel its warranted.

PR #30, opened 2024-07-25
test(ws): add e2e tests

Adembc:

- Remove the `+kubebuilder:validation:XValidation:rule="self == oldSelf"` annotations for image config and pod config.
- Comment out probes as they prevent pod creation when empty:
  ```
  probes:
    # startupProbe: {}
    # livenessProbe: {}
    # readinessProbe: {}
  ```
- Fix typo in extraEnv `NB_PREFIX` value from "juptyerlab" to "jupyterlab".
- Remove installation and uninstallation of Prometheus Operator and Cert Manager in e2e tests - before and after each test.
- Create the `default-editor` service account and delete it after testing.
- Create `home` PVC and `data` PVC, and delete them afterwards.
- Apply some `WorkspaceKind` samples.
- Apply some `Workspace` samples.
- Check if the workspace pod is running.
- Test if the pod is exposed correctly through the service and path generated from the `NB_PREFIX` environment variable.

### DIFF
--- a/workspaces/controller/api/v1beta1/workspacekind_types.go
+++ b/workspaces/controller/api/v1beta1/workspacekind_types.go
@@ -312,7 +312,6 @@ type ImageConfigValue struct {
 	Redirect *OptionRedirect `json:"redirect,omitempty"`
 
 	// the spec of the image config
-	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="ImageConfig 'spec' is immutable"
 	Spec ImageConfigSpec `json:"spec"`
 }
 
@@ -396,7 +395,7 @@ type PodConfigValue struct {
 	Redirect *OptionRedirect `json:"redirect,omitempty"`
 
 	// the spec of the pod config
-	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="PodConfig 'spec' is immutable"
+
 	Spec PodConfigSpec `json:"spec"`
 }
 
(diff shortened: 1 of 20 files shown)


thesuperzapper:
@Adembc I have done the following changes in https://github.com/kubeflow/notebooks/pull/30/commits/61c85fa0c88149706f7a255f69ad6b4ed6723552

- Auto detect the current kind cluster name for the `kind load docker-image ...` command.

thesuperzapper:
@Adembc I have done the following changes in https://github.com/kubeflow/notebooks/pull/30/commits/16886cededb63db5f815f554fc85de69089597e5

- Removed the tests which were failing because `+kubebuilder:validation:XValidation:rule="self == oldSelf"` was removed
- Made the default image `ghcr.io/kubeflow/notebooks/workspace-controller:latest` (as this will be the image name once we release)
- Grouped the samples under `./workspaces/controller/config/samples` better to separate the "common" ones like `ServiceAccount/default-editor` and `PVC/workspace-home-pvc` from our actual custom CRDs.
- Introduced a system to get Warning type Events from the StatefulSet and use them to pouplate the Workspace `status.stateMessage` when the Pod is missing.
     - note, this required adding a new Kubebuilder index on the `.involvedObject.uid` field of Events
- Introduced new Workspace `status.stateMessage` information for cases when the Pod is unschedulable (e.g. due to no node meeting the CPU/RAM request)
- Refactored the `./workspaces/controller/test/e2e/e2e_test.go`:
    - to not run tests in the `default` namespace
    - to check that the Workspace `status.state` is Running
    - to properly run a `curl` pod that checks the service is up
- Removed the now unused methods from `./workspaces/controller/test/utils/utils.go`:
     - `InstallPrometheusOperator()`
     - `UninstallPrometheusOperator()`
     - `UninstallCertManager()`
     - `InstallCertManager()`

PR #34, opened 2024-08-15
feat(ws): add validation webhooks for Workspace and WorkspaceKind

Adembc:
This PR makes the following changes:
- Re-add cert manager install, It was removed in this [PR](https://github.com/kubeflow/notebooks/pull/30)
### Workspace Validations:
- Reject creation of Workspaces with invalid `spec.kind` (must reference an existing WorkspaceKind).

### WorkspaceKind Validations:
- Reject updates to `spec.podTemplate.options.imageConfig.values[].spec`
- Reject updates to `spec.podTemplate.options.podConfig.values[].spec` 
- Prevent cycles in options.
- Only allow removal of option IDs that are not in use.
- Ensure a default option exists.
- Ensure that ports in `spec.podTemplate.options.imageConfig.values[].spec.ports[].port` are unique.
- Ensure `spec.podTemplate.extraEnv[].value` is a valid Go template.

### Tests:
- Add unit tests for validating Workspace creation.
- Add unit tests for validating WorkspaceKind creation.
- Add unit tests for validating updates to WorkspaceKind.


thesuperzapper:
@Adembc thanks for your work on this and the rest of GSOC!

I did quite a bit of refactoring in the following commits:
- https://github.com/kubeflow/notebooks/pull/34/commits/ce5ea50940ef455ba8be5b5aa37707cf0a81c3c2
- https://github.com/kubeflow/notebooks/pull/34/commits/f6ca24bdd2b0e8595fc143d0b8928fb4b834ba92
- https://github.com/kubeflow/notebooks/pull/34/commits/4487a2c2388f41b257ec3db124e55eb3da040e62

The WorkspaceKind webhook needed a lot of work to make it behave correctly (and detect redirect cycles in a semi-efficient way), and I also made the tests specifically check for each kind of error and added some "success" cases, because otherwise everything might just be failing.

I think this PR should be ready for an initial merge now, there should only be two things left to have a fully working controller, I will raise issues so we can track that work publicly:
1. Implement the "culling controller"
2. Adding Istio VirtualService reconciliation to the workspace controller.

/lgtm
/approve

PR #39, opened 2024-09-02
test(ws): enhance e2e test setup and cleanup

Adembc:
This PR introduces the following changes:

- Removes the prompt for end-to-end (E2E) test execution.
- Adds checks to verify if the CertManager and Prometheus operators are already installed. This prevents reinstallation and avoids unnecessary deletion.
- Added support for skipping CertManager installation with the environment variable `CERT_MANAGER_INSTALL_SKIP`.
- Added support for skipping Prometheus installation with the environment variable `PROMETHEUS_INSTALL_SKIP`.
- Enhances the E2E test setup and cleanup.

These changes follow [discussions](https://kubernetes.slack.com/archives/CAR30FCJZ/p1724837864980059) with the Kubebuilder maintainer, who has already merged similar updates into Kubebuilder in this [PR](https://github.com/kubernetes-sigs/kubebuilder/pull/4106/files).

### Note:
- `PROMETHEUS_INSTALL_SKIP` is currently set to `true` by default because the Prometheus operator is not required at this time. However, it might be needed in the future, so the option remains configurable.

### DIFF
--- a/workspaces/controller/Makefile
+++ b/workspaces/controller/Makefile
@@ -77,6 +77,7 @@ test-e2e:manifests generate fmt vet ## Run the e2e tests. Expected an isolated e
 		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
 		exit 1; \
 	}
+	# TODO: set PROMETHEUS_INSTALL_SKIP to `false` or remove `PROMETHEUS_INSTALL_SKIP=true` if we start using Prometheus
 	PROMETHEUS_INSTALL_SKIP=true go test ./test/e2e/ -v -ginkgo.v
 
 .PHONY: lint
--- a/workspaces/controller/Makefile
+++ b/workspaces/controller/Makefile
@@ -68,7 +68,7 @@ test: manifests generate fmt vet envtest ## Run tests.
 # - PROMETHEUS_INSTALL_SKIP=true
 # - CERT_MANAGER_INSTALL_SKIP=true
 .PHONY: test-e2e
-test-e2e:manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
+test-e2e: manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
 	@command -v kind >/dev/null 2>&1 || { \
 		echo "Kind is not installed. Please install Kind manually."; \
 		exit 1; \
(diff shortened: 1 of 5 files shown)


workspaces/controller/test/utils/utils.go
```diff
@@ -84,6 +133,39 @@ func InstallCertManager() error {
...
+		"certificates.cert-manager.io",
+		"issuers.cert-manager.io",
+		"clusterissuers.cert-manager.io",
+		"certificaterequests.cert-manager.io",
+		"orders.acme.cert-manager.io",
+		"challenges.acme.cert-manager.io",
+	}
+
+	// Execute the kubectl command to get all CRDs
+	cmd := exec.Command("kubectl", "get", "crds")
```
thesuperzapper, line 150: To avoid having the `kubectl get` column headings, we can use `kubectl get crds -o name`

workspaces/controller/Makefile
```diff
@@ -63,11 +63,23 @@ vet: ## Run go vet against code.
...
+test-e2e: ## Run the e2e tests. Expected an isolated environment using Kind.
+	@command -v kind >/dev/null 2>&1 || { \
+		echo "Kind is not installed. Please install Kind manually."; \
+		exit 1; \
+	}
+	@kind get clusters | grep -q 'kind' || { \
+		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
+		exit 1; \
+	}
+	PROMETHEUS_INSTALL_SKIP=true go test ./test/e2e/ -v -ginkgo.v
```
thesuperzapper, line 81: Add a basic TODO which explains that we are not installing Prometheus because we are not using it right now,

E.g.:

```suggestion
        # TODO: set PROMETHEUS_INSTALL_SKIP to `false` if we start using Prometheus 
	PROMETHEUS_INSTALL_SKIP=true go test ./test/e2e/ -v -ginkgo.v
```

thesuperzapper:
@Adembc thanks for this, after we added the "should not error" check for the applying of the controller CRDs, we were getting errors when we re-ran the e2e tests quickly.

This was because the `ValidationWebhookConfigurations` were not able to get their CA injected by cert-manager, because even though the cert-manager pods were ready, they had not taken the lease (so were not actually managing the cluster yet).

To fix this, I had to automatically delete the leases used by cert-manager so that it would take ownership of the CA injection faster.

see https://github.com/kubeflow/notebooks/pull/39/commits/ac8455a4456c23ea8029ee2f299db3c0ed9127a9

PR #43, opened 2024-09-11
feat(ws): use controller-runtime for backend

thesuperzapper:
This PR sets up the Workspace Backend API to use `controller-runtime` rather than ClientSets to access the Kubernetes API.

This has some significant benefits:

1. We can use the local `envtest` for mocking API responses locally (like we already do in the controller tests)
2. Rather than constantly querying the API, it will use a "watch cache" based approach, so that we only list the resources once (e.g. Workspaces), and then watch for changes.
3. For extremely fast lookup of complicated queries like "list all workspaces owned by workspace kind xxxx", we can use indexes on the cache, like we [already do](https://github.com/kubeflow/notebooks/blob/bc4e4454bc1c921549cb9df5865e52a61fe07946/workspaces/controller/internal/helper/index.go#L50-L74) for the controller.

The way this is implemented is a little hacky, because we are directly adding a new "Runnable" to the `controller-runtime` manager object, and this runnable encapsulates our internal `net/http` server.

Other than that, this PR makes the following changes:

1. Loads the Workspace/WorkspaceKind types using a `replace github.com/kubeflow/notebooks/workspaces/controller => ../controller` redirect in the `go.mod`
2. Adds `controller-runtime` to `go.mod`
1. Moves all non-api go code under `internal/`
4. Makes the `/api/v1/healthcheck/` path return a list of Workspace names (lol, we can revert this before merging).
5. Renames the local variable for `App` from `app` to `a`


thesuperzapper:
Also, the tests won't pass until we either remove the dependency on the client being up (in the health-check test), or do exactly what we have done in the controller, and set up the controller-manager:

https://github.com/kubeflow/notebooks/blob/bc4e4454bc1c921549cb9df5865e52a61fe07946/workspaces/controller/internal/controller/suite_test.go#L98-L105

We also will need to add the `github.com/onsi/gomega` testing framework as a dependency of the backend.

ederign:
Rebased and updated here: https://github.com/kubeflow/notebooks/pull/44

ederign (approved):

PR #137, opened 2024-11-28
chore: update go linting + update dependencies for controller

thesuperzapper:
This PR makes the following changes:

1. Updates to dependencies used on the controller:

      - Go 1.21 -> 1.22
      - controller-runtime: 0.17.3 -> 0.19.1
      - golangci-lint: 1.57.2 -> 1.61.0
      - EnvTest: 1.29 -> 1.31

2. Updates to dependencies used by backend:

      - controller-runtime: 0.17.3 -> 0.19.1
      - golangci-lint: 1.57.2 -> 1.61.0
      - EnvTest: 1.29 -> 1.31

3. Improved the linting for all go projects (controller, backend):

     - We now enforce comment formats
     - We now enforce license headers
     - We now do lots of code smell checks
     - We now have a spell checker


thesuperzapper:
@ederign can you please LGTM when you are ready, this is an important update, and we need to get other PRs to rebase on it so they get the new linting.

/approve

workspaces/controller/internal/webhook/workspacekind_webhook_test.go
```diff
@@ -96,7 +96,6 @@ var _ = Describe("WorkspaceKind Webhook", func() {
 		}
 
 		for _, tc := range testCases {
-			tc := tc // Create a new instance of tc to avoid capturing the loop variable.
```
ederign, line 99: @thesuperzapper I want to double-check with you that this doesn't break anything. Based on the comment, the variable looks pretty intentional :)

workspaces/controller/config/rbac/role.yaml
```diff
@@ -4,37 +4,11 @@ kind: ClusterRole
 metadata:
   name: manager-role
 rules:
-- apiGroups:
```
ederign, line 7: Also, are sure about changing the roles?

ederign (changes requested):
@thesuperzapper thank you so much for this important PR. Just want to double check a few things before +1.

ederign (approved):

Issue #69, opened 2024-10-28
Notebooks 2.0 // Controller // Replace `reflect.DeepEqual` with `equality.Semantic.DeepEqual`


thesuperzapper:
## What's the problem?

We currently use `reflect.DeepEqual` to do a lot of comparisons in the Notebooks 2.0 controller.

As a result of this, we have to take special care when comparing Kubernetes API objects for a number of reasons:

1. a `nil` map is not equal to an empty map
2. a `nil` slice is not equal to an empty slice
3. resource requests need to be carefully compared (as they can be specified in a lot of forms)
4. time quantities need to be carefully compared (as they can be specified in a lot of forms)

Therefore, we need to use helper method like [`NormalizePodConfigSpec`](https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/controller/internal/helper/helper.go#L106-L168) and other cases to handle this complexity.

## What am I proposing?

We can replace this complexity by using Kubernetes built in API object semantic comparison:

```go
import "k8s.io/apimachinery/pkg/api/equality"

...

if equality.Semantic.DeepEqual(object1, object1) {
  return false, nil
}
```

## Next Steps

1. replace all instances of `reflect.DeepEqual` with `equality.Semantic.DeepEqual` 
     - ___NOTE:__ this should ONLY be changed for comparisons between Kubernetes API objects, or parts of them, not other things we might be comparing._
3. remove any special methods we are using to compare/mutate the objects before applying the equality check

Adembc:
Good catch! Really interesting stuff.

Adembc:
Could you please assign this issue to me?
@thesuperzapper

PR #152, opened 2024-12-04
improve(ws): replace reflect.DeepEqual with equality.Semantic.DeepEqual

Adembc:
Closes https://github.com/kubeflow/notebooks/issues/69


thesuperzapper:
@Adembc one thing we might need to check is that we aren't constructing any explicit empty values for fields with `omitempty`.

Although that issue is more about the status field, and we might actually have this issue already because we use `reflect.DeepEqual` to check if we should update the status (so if we have explicitly constructed one of the fields as an empty list, like `RedirectChain` it will always update the status every reconcile).

Issue #149, opened 2024-12-04
Notebooks 2.0 // Frontend // Workspaces table // Workspace Data Vol column


paulovmr:
Modify the "Data Vol" column to display all data PVCs on mouse hover. Figure out the best PatternFly 6 component to use for better visualization.

Issue #150, opened 2024-12-04
Notebooks 2.0 // Frontend // Workspaces table // Workspace Connect column


paulovmr:
Add the "Connect" column at the end of the table, before the actions column, to display a list of endpoints coming from the backend service on mouse hover. Figure out the best PatternFly 6 component to use for better visualization.

To be defined:
* What backend service will return this information?

Yael-F:
Hey @paulovmr 
Can you please assign this task to me?

ederign:
This issue is still planned for Notebooks 2.0.

PR #160, opened 2024-12-12
feat(ws): Add popup to Workspace data vol column in workspaces table

ElayAharoni:
Fixes: #149

added popup with the worspace data vol info, that will show when hovering on the data vol column field.

please inform me if anything needs to be changes in matter of functionality or styling.

added a picture of the change:

image

### DIFF
--- a/workspaces/frontend/src/app/pages/Workspaces/DataVolumesList.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/DataVolumesList.tsx
@@ -1,0 +1,74 @@
+import {
+  ClipboardCopy,
+  ClipboardCopyVariant,
+  Content,
+  Flex,
+  FlexItem,
+  List,
+  ListItem,
+  Stack,
+  StackItem,
+  Tooltip,
+} from '@patternfly/react-core';
+import { DatabaseIcon, LockedIcon } from '@patternfly/react-icons';
+import * as React from 'react';
+import { Workspace } from '~/shared/types';
+
+interface DataVolumesListProps {
+  workspace: Workspace;
+}
+
+export const DataVolumesList: React.FC<DataVolumesListProps> = ({ workspace }) => {
+  const workspaceDataVol = workspace.podTemplate.volumes.data;
+
+  const singleDataVolRenderer = (data: {
+    pvcName: string;
+    mountPath: string;
+    readOnly: boolean;
+  }) => (
+    <Flex
+      gap={{ default: 'gapSm' }}
+      alignItems={{ default: 'alignItemsFlexStart' }}
+      flexWrap={{ default: 'nowrap' }}
+    >
+      <FlexItem>
+        <DatabaseIcon />
+      </FlexItem>
+      <FlexItem>
+        <Content>
+          {data.pvcName}
+          {data.readOnly && (
+            <Tooltip content="Data is readonly">
+              <LockedIcon style={{ marginLeft: '5px' }} />
+            </Tooltip>
+          )}
+        </Content>
+        <Stack>
+          <StackItem>
+            <Content component="small">
+              Mount path:{' '}
+              <ClipboardCopy variant={ClipboardCopyVariant.inlineCompact}>
+                {data.mountPath}
+              </ClipboardCopy>
+            </Content>
+          </StackItem>
+        </Stack>
+      </FlexItem>
+    </Flex>
+  );
+
+  return (
+    <Stack hasGutter>
+      <StackItem>
+        <strong data-testid="notebook-storage-bar-title">Cluster storage</strong>
+      </StackItem>
+      <StackItem>
+        <List isPlain>
+          {workspaceDataVol.map((data, index) => (
+            <ListItem key={`data-vol-${index}`}>{singleDataVolRenderer(data)}</ListItem>
+          ))}
+        </List>
+      </StackItem>
+    </Stack>
+  );
+};
--- a/workspaces/frontend/src/app/pages/Workspaces/Workspaces.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Workspaces.tsx
@@ -35,7 +35,9 @@ import {
   IActions,
 } from '@patternfly/react-table';
 import { FilterIcon } from '@patternfly/react-icons';
-import { Workspace, WorkspaceState } from 'shared/types';
+import { Workspace, WorkspacesColumnNames, WorkspaceState } from '~/shared/types';
+
+import { ExpandedWorkspaceRow } from '~/app/pages/Workspaces/ExpandedWorkspaceRow';
 import { formatRam } from 'shared/utilities/WorkspaceResources';
 
 export const Workspaces: React.FunctionComponent = () => {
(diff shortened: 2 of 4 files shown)


paulovmr:
Hi @ElayAharoni , thanks for the PR! For this task, I asked Eugenia from UX to give her input on the best component to use to show those data volumes. One thing that would need to change, is that we need to show not only the first but all elements from the `workspace.podTemplate.volumes.data` array, so there might be a need for an alternative component than the `Popover`.

paulovmr (commented):
Hi @ElayAharoni , thanks for the PR! For this task, I asked @jenny-s51 from UX to give her input on the best component to use to show those data volumes. One thing that would need to change, is that we need to show not only the first but all elements from the `workspace.podTemplate.volumes.data` array, so there might be a need for an alternative component than the `Popover`.

paulovmr:
Hi @jenny-s51 , here is the PR about the data volumes column.

jenny-s51:
Hi @paulovmr , thank you for tagging me! https://github.com/kubeflow/notebooks/issues/150 is a similar ticket as well. Per our recent conversation this morning we agreed that the expandable table variant is an effective approach and component to use for these cases - since we want to show additional data for multiple columns in this table view, using the PF[ expandable table variant](https://www.patternfly.org/components/table/react-demos#expandcollapse-all) offers more flexibility and provides a visual cue to the user that there is more data within the table view. WDYT @ElayAharoni ?

jenny-s51:
Yes, that sounds like a good plan, @ElayAharoni.

ElayAharoni:
hi @paulovmr , @jenny-s51.
i have made the changes, to show the data via expendable table row.
attached screenshot for reference:

image

paulovmr:
IMO the idea of adding a expandable row would eliminate the need for a normal column for that field, since the expandable part would be presented as a separate row that adds more information, and not an extension to the columns, like seen in [here](https://www.patternfly.org/components/table#expandable). @jenny-s51 do you agree?

jenny-s51 (commented):
This is a good callout @paulovmr - if we are showing additional data for multiple columns, the data should be presented within the span of the column it is associated with. 

Here is a good example of how this is currently done in the RHOAI Dashboard, for example the "Container size" column has Limits and Memory within the same column span (see [NotebookTableRow.tsx](https://github.com/jenny-s51/odh-dashboard/blob/main/frontend/src/pages/projects/screens/detail/notebooks/NotebookTableRow.tsx#L218-L243)).
Screenshot 2024-12-18 at 5 22 17 PM

We also want to make sure to wrap the content of each child row cell in ExpandableRowContent to apply the correct styles. @ElayAharoni

ElayAharoni:
hi @jenny-s51 @paulovmr.
i have made the requested fixes.
attached new snapshot:
image

ElayAharoni:
hi @paulovmr , @jenny-s51.
after the discussions we had this is the current fix.
please review when you have time.
attached latest snapshot: 
image

workspaces/frontend/src/app/pages/Workspaces/ExpandedDataVol.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+interface ExpandedDataVolProps {
+  workspace: Workspace;
+}
+
+export const ExpandedDataVol: React.FC<ExpandedDataVolProps> = ({ workspace }) => {
+  const workspaceDataVol = workspace.podTemplate.volumes.data;
+  return (
+    <>
+      <Content component="h3">Data Volumes:</Content>
```
paulovmr, line 14: @ElayAharoni IMO the List component from PatternFly might be a better fit for this kind of information on the UI. Please consider replacing the `Content` usage for the `List` component.

workspaces/frontend/src/app/pages/Workspaces/ExpandedDataVol.tsx
```diff
@@ -0,0 +1,37 @@
...
+      {workspaceDataVol.map((data, index) => (
+        <Content key={`data-vol-${index}`} component="blockquote">
+          <div>
```
paulovmr, line 17: In general, it is best to avoid the usage of plain HTML components in the middle of PatternFly components to avoid breaking future CSS rules.

workspaces/frontend/src/app/pages/Workspaces/ExpandedDataVol.tsx
```diff
@@ -0,0 +1,37 @@
+import { Content } from '@patternfly/react-core';
+import { LockedIcon } from '@patternfly/react-icons';
+import * as React from 'react';
+import { Workspace } from '~/shared/types';
+
+interface ExpandedDataVolProps {
+  workspace: Workspace;
+}
+
+export const ExpandedDataVol: React.FC<ExpandedDataVolProps> = ({ workspace }) => {
```
paulovmr, line 10: Please consider renaming this component (and also other associated declarations).
```suggestion
export const DataVolumesList: React.FC<ExpandedDataVolProps> = ({ workspace }) => {
```

paulovmr (changes requested):

ElayAharoni:
@paulovmr thanks for your comments i will fix them

workspaces/frontend/src/app/pages/Workspaces/DataVolumesList.tsx
```diff
@@ -0,0 +1,74 @@
...
+  ListItem,
+  Stack,
+  StackItem,
+  Tooltip,
+} from '@patternfly/react-core';
+import { DatabaseIcon, LockedIcon } from '@patternfly/react-icons';
+import * as React from 'react';
+import { Workspace } from '~/shared/types';
+
+interface ExpandedDataVolProps {
```
paulovmr, line 17:
```suggestion
interface DataVolumesListProps {
```

workspaces/frontend/src/app/pages/Workspaces/DataVolumesList.tsx
```diff
@@ -0,0 +1,74 @@
...
+  workspace: Workspace;
+}
+
+export const DataVolumesList: React.FC<ExpandedDataVolProps> = ({ workspace }) => {
```
paulovmr, line 21:
```suggestion
export const DataVolumesList: React.FC<DataVolumesListProps> = ({ workspace }) => {
```

paulovmr (changes requested):
Hi @ElayAharoni , left a few minor changes inline. One thing that is important as well is to remove the `Data Vol` column from the table, now that the information has moved to the expandable row. Also please add the new type `WorkspacesColumnNames` to the variable `columnNames`.

workspaces/frontend/src/app/pages/Workspaces/Workspaces.tsx
```diff
@@ -438,26 +456,34 @@ export const Workspaces: React.FunctionComponent = () => {
...
-            <Th sort={getSortParams(6)}>{columnNames.dataVol}</Th>
-            <Th sort={getSortParams(7)} info={{ tooltip: 'Workspace CPU usage' }}>
+            <Th sort={getSortParams(6)} info={{ tooltip: 'Workspace CPU usage' }}>
               {columnNames.cpu}
             </Th>
             <Th sort={getSortParams(8)} info={{ tooltip: 'Workspace memory usage' }}>
               {columnNames.ram}
             </Th>
             <Th sort={getSortParams(9)}>{columnNames.lastActivity}</Th>
+            <Th sort={getSortParams(6)}>{columnNames.lastActivity}</Th>
```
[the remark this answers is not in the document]
ElayAharoni: yeah you are right, i have fixed and pused for the final time i hope

paulovmr (approved):

ederign (approved):

Issue #163, opened 2024-12-22
Notebooks 2.0 // Frontend // Fetch API Infra


asaadbalum:
Bring up Model-Registry's API mechanism to the NB2 Fronted.
Note: Bring up the mechanism, and adjust it to fetch the Frontend usage.

ederign:
Done by https://github.com/kubeflow/notebooks/pull/167

ederign:
@YosiElias feel free to close it!

PR #167, opened 2024-12-24
feat(ws): Fetch API Infra

YosiElias:
fix: #163


paulovmr (approved):
@YosiElias LGTM!

@ederign This PR copies a few utilities from `model-registry`, that will later be extracted to a common dependency.

PR #169, opened 2024-12-24
feat(ws): Notebooks 2.0 // FrontEnd // Add CPU and RAM columns

### LINKED ISSUE #151: Workspace CPU and RAM columns
Add the "CPU" and "RAM" columns after the "Data Vol" column, to display CPU percentage and RAM amount usages.

To be defined:
* What backend service will return this information?

senanz:
Add new columns CPU and RAM that used by the workspace, fixes #151


workspaces/frontend/package.json
```diff
@@ -86,6 +86,9 @@
     "webpack-merge": "^5.10.0"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-svg-core": "^6.7.2",
```
paulovmr, line 89: I believe we can use patternfly icons directly here, without other dependencies. Please consider following [this example](https://www.patternfly.org/components/icon#status-colors), which uses `@patternfly/react-icons` resources, which is already added.

workspaces/frontend/src/app/pages/Workspaces/Workspaces.tsx
```diff
@@ -35,7 +35,10 @@ import {
   IActions,
 } from '@patternfly/react-table';
 import { FilterIcon } from '@patternfly/react-icons';
-import { Workspace, WorkspaceState } from '~/shared/types';
+import { Workspace, WorkspaceState } from 'shared/types';
+import {formatRam} from "shared/utilities/WorkSpaceResources";
```
paulovmr, line 39: I think `WorkSpaceResources` is missing on this PR. Please also consider renaming `WorkSpace` to `Workspace`, here and on other parts of this PR.

workspaces/frontend/src/app/pages/Workspaces/Workspaces.tsx
```diff
@@ -304,7 +315,11 @@ export const Workspaces: React.FunctionComponent = () => {
           (dataVol) =>
             dataVol.pvcName.search(searchValueInput) >= 0 ||
             dataVol.mountPath.search(searchValueInput) >= 0,
-        ))
+        )) ||
```
paulovmr, line 318: I don't think CPU and RAM should belong to filters.

workspaces/frontend/src/shared/types.ts
```diff
@@ -52,6 +52,8 @@ export interface Workspace {
   paused: boolean;
   deferUpdates: boolean;
   kind: string;
+  CPU: number;
```
paulovmr, line 55:
```suggestion
  cpu: number;
```

workspaces/frontend/src/shared/types.ts
```diff
@@ -52,6 +52,8 @@ export interface Workspace {
...
+  RAM: number;
```
paulovmr, line 56:
```suggestion
  ram: number;
```

paulovmr (changes requested):

workspaces/frontend/package.json
senanz: Hi @paulovmr , Since i'm adding test infromation about the field i can use react-icon bu need to import and use dependcies for other toolib to add the infromation and present it when hover on the icon.
Please let me know if you prefer the suggested on the one that already implmented

paulovmr: Hi @senanz , I believe for this case specifically there is an OOTB way to do that with patternfly table headers, please check it [here](https://www.patternfly.org/components/table#custom-row-wrapper-header-tooltips--popovers). You could have the tooltip by setting the `info` property in the `Th` component, like `<Th info={{ tooltip: 'More information ' }}></Th>`.

workspaces/frontend/src/app/pages/Workspaces/Workspaces.tsx
senanz: You Right, will be changed.

workspaces/frontend/src/app/pages/Workspaces/Workspaces.tsx
senanz: Agree.

workspaces/frontend/src/shared/types.ts
senanz: @paulovmr , I intended to write them in capital letter as they are abbreviation

workspaces/frontend/src/shared/types.ts
senanz: @paulovmr , I intended to write them in capital letter as they are abbreviation

workspaces/frontend/package.json
senanz: Thanks for that, will be changed

workspaces/frontend/src/shared/types.ts
paulovmr: @senanz As a convention and not a rule, I would say the lower case is preferred in this case. Also to keep it the same as the workspace CRDs, like [here](https://github.com/thesuperzapper/kubeflow-notebooks-v2-design/blob/de9663a0cf9d3e61858694d0a10b8cddd0a8404e/crds/workspace-kind.yaml#L334-L335).

paulovmr (changes requested):

workspaces/frontend/package-lock.json
```diff
@@ -21397,4 +21397,4 @@
       }
     }
   }
-}
+}
```
paulovmr, line 21400: @senanz This change also needs to be reverted, otherwise the `npm install` will change it back as before locally for everyone.

workspaces/frontend/src/app/pages/Workspaces/Workspaces.tsx
```diff
@@ -129,6 +134,8 @@ export const Workspaces: React.FunctionComponent = () => {
     state: 'State',
     homeVol: 'Home Vol',
     dataVol: 'Data Vol',
+    cpu: 'cpu',
```
paulovmr, line 137: For the column name, it should be uppercase
```suggestion
    cpu: 'CPU',
```

workspaces/frontend/src/app/pages/Workspaces/Workspaces.tsx
```diff
@@ -129,6 +134,8 @@ export const Workspaces: React.FunctionComponent = () => {
...
+    ram: 'ram',
```
paulovmr, line 138:
```suggestion
    ram: 'Memory',
```

workspaces/frontend/package-lock.json
senanz: Fixed.

workspaces/frontend/src/app/pages/Workspaces/Workspaces.tsx
senanz: Fixed

workspaces/frontend/src/app/pages/Workspaces/Workspaces.tsx
senanz: Fixed

paulovmr (approved):

ederign (approved):

Issue #148, opened 2024-12-04
Notebooks 2.0 // Frontend // Workspaces table // Workspace Kind column


paulovmr:
Modify the "Kind" column to show an icon representing the WorkspaceKind instead of its displayName. The icon URL will come from [here](https://github.com/thesuperzapper/kubeflow-notebooks-v2-design/blob/de9663a0cf9d3e61858694d0a10b8cddd0a8404e/crds/workspace-kind.yaml#L31).

liavweiss:
PR was merged, please close the ticket.

PR #177, opened 2025-01-09
feat(ws): Notebooks 2.0 // Frontend // Workspaces table // Workspace Kind column #148

liavweiss:
This pr also contains the implementation of fetch data from workspacekinds API, to be able to take the Icon URL. 

In case we have an Icon:
image

In case we don't have an Icon(broken URL or Icon does not exist):
image


node_modules/.package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
paulovmr, line 1: This file should not be created. It was probably created because of an `npm` command being ran in the repository root.

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
paulovmr, line 1: This file should not be created. It was probably created because of an `npm` command being ran in the repository root.

workspaces/frontend/src/shared/types.ts
```diff
@@ -10,11 +10,60 @@ export interface WorkspaceKind {
...
+      annotations: {
+        myWorkspaceKindAnnotation: string
+      }
+    },
+    volumeMounts: {
+            home: string
+    },
+    options: {
+      imageConfig: {
+        default: string,
```
paulovmr, line 32:
```suggestion
        spawner: {
          default: string,
        },
```

workspaces/frontend/src/shared/types.ts
```diff
@@ -10,11 +10,60 @@ export interface WorkspaceKind {
...
+        values: [
+          {
+            id: string,
+            displayName: string,
+            labels: {
+              pythonVersion: string
+            },
+            hidden: true,
```
paulovmr, line 40:
```suggestion
            spawner: {
              displayName: string,
              description: string,
              labels: [
                {
                  key: string,
                  value: string;
                }
              ],
              hidden: true,
            }
```

workspaces/frontend/src/shared/types.ts
```diff
@@ -10,11 +10,60 @@ export interface WorkspaceKind {
...
+              message: {
+                text: string,
+                level: string
+              }
+            }
+          },
+        ]
+      },
+      podConfig: {
+        default: string,
```
paulovmr, line 52:
```suggestion
        spawner: {
          default: string,
        },
```

workspaces/frontend/src/shared/types.ts
```diff
@@ -10,11 +10,60 @@ export interface WorkspaceKind {
...
+        values: [
+          {
+          id: string,
```
paulovmr, line 55: This part is also not matching the CR structure [here](https://github.com/thesuperzapper/kubeflow-notebooks-v2-design/blob/de9663a0cf9d3e61858694d0a10b8cddd0a8404e/crds/workspace-kind.yaml#L317).

paulovmr (changes requested):
Thanks for the PR @liavweiss ! I've made a few comments inline, also please consider using the camel case `[W|w]orkspaceKinds` instead of `[W|w]orkspacekinds` for variable names where applicable.

paulovmr (approved):

ederign (changes requested):
@liavweiss thanks for the PR. Can we have a default logo (even just a placeholder) when the logo is not available? I think it would be a much better experience for the user then showing the 'broken' image.

liavweiss:
After Eder comment and a discussion with Paulo and Jenny we agreed to have a default Icon in case we don't have icon for the specific kind
image
in case we have an icon it will be the same as the first pic in this page.

ederign (approved):

ederign:
@liavweiss thank you for accommodating!

PR #193, opened 2025-02-04
fix(ws): handle empty pod metadata in controller

thesuperzapper:
This PR fixes an issue where the controller did not handle the possibility that `spec.podTemplate.podMetadata` could be empty on the WorkspaceKind, and `spec.podTemplate.podMetadata` could be empty on the Workspace.

It was simply that `labels.Merge()` is not the right way to do this, because it requires that all arguments be non-nil maps.


thesuperzapper:
@ederign here is another small PR for the controller.

ederign (approved):

PR #202, opened 2025-02-11
feat(ws): add auth to backend

thesuperzapper:
This PR:

1. adds authentication to the backend
2. Removes the `helper.GetKubeconfig` and replaces it with `ctrl.GetConfig()`
      - This allows us to set `CLIENT_QPS` and `CLIENT_BURST` as configs (because the defaults are very low 20/30)

Authentication is implemented by reading the `USERID_HEADER` and `GROUPS_HEADER` in each request (see `NewRequestAuthenticator()`).

Authorization is implemented with SubjectAccessReviews (see `NewRequestAuthorizer()`). The `NewRequestAuthorizer()` method uses the `authorizerfactory.DelegatingAuthorizerConfig` from `k8s.io/apiserver/pkg/authorization/authorizerfactory` and has a 10-second cache to not check the user is still authorized more than once every 10 seconds.

To define what auth a specific handler requires, we have a new `a.requireAuth()` method, which takes a new `[]ResourcePolicy`  list, that can easily be constructed using `NewResourcePolicy()` in a clear way.


ederign:
@thesuperzapper, impressive PR. This approach is indeed smart! :)

As an FUP, I'll add the authorization mocks for the front end (in the standalone). After 1.10 release, I"ll also upgrade Model Registry auth for this.

Great work

ederign (approved):

PR #209, opened 2025-02-13
feat(ws): Add GET workspace yaml endpoint

mohamedch7:


workspaces/backend/api/workspace_yaml_handler.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2024.
```
ederign, line 1: just for the sake of aesthetics, could you please add a block comment instead of line comment?

workspaces/backend/api/workspace_yaml_handler.go
```diff
@@ -0,0 +1,67 @@
...
+func (a *App) GetWorkspaceYAMLHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	namespace := ps.ByName(NamespacePathParam)
+	workspaceName := ps.ByName(ResourceNamePathParam)
+
+	if namespace == "" || workspaceName == "" {
+		a.serverErrorResponse(w, r, fmt.Errorf("namespace or workspace name is empty"))
+		return
+	}
+
+	workspace, err := a.repositories.Workspace.GetWorkspace(r.Context(), namespace, workspaceName)
```
ederign, line 43: @mohamedch7 I would like to confirm with @paulovmr what we are doing here.

As far as I understand, what we are planning for the UI is having a  "RAW" YAML representation of the workspace and not the models.Workspace object (that we translate)

@paulovmr can you please confirm that what we want is a way to see the RAW Workspace YAML there?

ederign (changes requested):
We will probably need to change the call to that repository.

workspaces/backend/api/workspace_yaml_handler.go
paulovmr: Yes @ederign , we do want the raw YAML here.

mohamedch7: @ederign @paulovmr 
I've changed the endpoint to get raw YAML
image

Issue #208, opened 2025-02-12
Notebooks 2.0 // Backend // Workspace Logs API //  Fetch Workspace container logs


Yael-F:
### Summary
Develop an API endpoint to retrieve logs from a container running in a workspace. 
This API should support retrieving both the current and 'previous' log and provide a real-time log view option.

**Key features:**
1. Retrieve logs of a container in a workspace.
2. Fetch real-time logs from a container.
3. Support fetching logs from a previously terminated container run.
4. Limit the size of the fetched logs.
5. RBAC - only authorized users can view the logs
6. No sensitive data is exposed.

### Suggested Implementation:

1. **API Endpoints:**
   - **For log pagination and old log retrieval:**
   **PATH:** GET `/api/v1/workspaces/{namespace}/{name}/details/log`
   **Query Parameters:**
      - `container `(required): The name of the container for which logs are requested.
      - `previous `(optional): Fetch logs from the previously terminated container.
      - `offset` (optional): The starting point for pagination. It indicates the number of lines to skip (default: `end_of_file - limit` or 0, whichever is larger).
      - `limit` (optional): The number of lines to retrieve (default: 100, max limit: 1,000).
      
      **Response:**
`{
"logs": [],
"pagination": {
"offsetFrom": start_index,
"offsetTo": end_index,
"endOfFile": bool 
}`

   - **For log streaming:**
**PATH:** GET `/api/v1/workspaces/{namespace}/{name}/details/log/stream`
**Query Parameters:**
      - `container `(required): The name of the container for which logs are requested.
      - `sinceTime` (optional): Retrieve logs after a specified timestamp.

      **Response:**
`{
"logs": [],
"referenceTimestamp": last_timestamp_sent
}`

2. **Real time log retrieval:**
Implement Server-Sent Events (SSE)/WS for streaming.

3. **Add 'previous' query parameter**
Adding query parameter to indicate previous log retrieval.
The pod container list data should include for each container if 'previous log' exists, so the option is disabled for containers that doesn't have previous log. in case of API called for container with no previous log, the API returns:
HTTP 400 Bad Request → "No previous logs available for this container." 

4. **Size of the log**
Limit the response to a maximum of 1,000 lines of log data.

5. **RBAC**
Only workspace owner and contributors can view the logs. 

6. **Sensitive data**
By default, no sensitive data will be filtered.
Possible enhancement:
[...]

Issue #172, opened 2024-12-27
Notebooks 2.0 // Frontend // Workspaces details // Logs tab


paulovmr:
Implement the logs tab using data coming from the backend service to be implemented.

Yael-F:
Hi @paulovmr 
I’m sharing a screenshot example here to validate the requirements. Should we also include 'live streaming' of container logs, or is the example provided sufficient?

![Image](https://github.com/user-attachments/assets/9205f5d3-62c0-408e-b7d7-a6d05181d3ce)

PR #210, opened 2025-02-14
feat(ws): add `podTemplatePod` to Workspace status, for backend logs

thesuperzapper:
related: https://github.com/kubeflow/notebooks/issues/208
related: https://github.com/kubeflow/notebooks/issues/172

Currently, it would have been impossible for the backend to know which pod to read logs from as part of https://github.com/kubeflow/notebooks/issues/208, so this PR adds `podTemplatePod` to the status of Workspaces.

This new status field will look something like (when the Pod exists):

```yaml
status:
  podTemplatePod:
    name: ws-jupyterlab-workspace-56mlf-0
    containers:
      - name: main
    initContainers: []
```

This new status field will look something like (when the Pod does not exist):

```yaml
status:
  podTemplatePod:
    name: ""
```

---

This PR also:

- improves the `generateWorkspaceStatus` method by factoring out a `generateWorkspaceState` to generate the state/stateMessage
- ensures we only do an UPDATE call when the status actually changes on WS and WSK (reduce API load)
- factors out a common `workspacePodTemplateContainerName` variable, rather than using `"main"` everywhere


thesuperzapper:
@ederign this should be ready to review/lgtm

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -204,15 +219,9 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-		workspace.Status.PendingRestart = false
-		if err := r.Status().Update(ctx, workspace); err != nil {
-			if apierrors.IsConflict(err) {
-				log.V(2).Info("update conflict while updating Workspace status, will requeue")
-				return ctrl.Result{Requeue: true}, nil
-			}
-			log.Error(err, "unable to update Workspace status")
-			return ctrl.Result{}, err
-		}
+
```
ederign, line 222: I'm new for controller, but why we don't need workspace.Status.PendingRestart = false anymore?

ederign (commented):
I'm learning how controller works, so just left a few comments to clarify. @etirelli @harshad16 @andyatmiami maybe one of you can take a look on it?

workspaces/controller/internal/controller/workspace_controller.go
thesuperzapper: @ederign a few lines up I realized we were not always correctly setting it.

Now I build the status by setting it to false at the beginning and then only setting it to true when I detect an option has a redirect.

workspaces/controller/config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ resources:
 - manager.yaml
 images:
 - name: controller
-  newName: ghcr.io/kubeflow/notebooks/workspace-controller
+  newName: controller
```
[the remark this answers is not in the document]
thesuperzapper: Fixed in https://github.com/kubeflow/notebooks/pull/210/commits/689c8283c5cbe57d9ad7e7bec151fd62d185e2cd

workspaces/controller/internal/controller/workspace_controller.go
ederign: Thank you for the clarification!

ederign (approved):

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -961,32 +1027,32 @@ func (r *WorkspaceReconciler) generateWorkspaceStatus(ctx context.Context, log l
...
-			status.StateMessage = stateMsgRunning
-			return status, nil
+			state = kubefloworgv1beta1.WorkspaceStateRunning
+			stateMessage = stateMsgRunning
+			return state, stateMessage, nil
 		}
 
 		// get container status
 		// https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states
 		var containerStatus corev1.ContainerStatus
```
harshad16, line 1053: Is there initcontainerstatus maintained some other place ?

harshad16 (commented):
still getting upto speed with codeblock
lgtm in general

just one question

workspaces/controller/internal/controller/workspace_controller.go
thesuperzapper: @harshad16 you are correct that the initContainer status is stored under `status.InitContainerStatuses`, but in this case, we are looking for the status of the `"main"` container, so we can get info about it.

But you are correct that we might have a problem if the Pod is failing because an init-container (or in fact some other container is in an error state), as we would fall through to the Unknown State.

Do you want to pick this up as another task after we merge?

thesuperzapper:
I am merging this for now, but as @harshad16 said in https://github.com/kubeflow/notebooks/pull/210#discussion_r1956742897, there is still some work to do on the state extraction.

thesuperzapper:
I am merging this for now, but as @harshad16 said in https://github.com/kubeflow/notebooks/pull/210#discussion_r1956742897, there is still some work to do on the state extraction.

/approve

workspaces/controller/internal/controller/workspace_controller.go
harshad16: Yes absolutely,  i can pick up the task for including the initcontainer check.

PR #188, opened 2025-01-30
test(ws): Add workspaces tests

yehudit1987:
This PR resolve issue #142

PR #213, opened 2025-02-16
feat(ws): add http service paths to WS get on backend

yehudit1987:
This PR is for adding missing ports field to workspace model.

### DIFF
--- a/workspaces/backend/internal/models/workspaces/funcs.go
+++ b/workspaces/backend/internal/models/workspaces/funcs.go
@@ -18,6 +18,7 @@ package workspaces
 
 import (
 	"fmt"
+	"path"
 
 	kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"
 	"k8s.io/utils/ptr"
@@ -124,16 +125,17 @@ func NewWorkspaceModelFromWorkspace(ws *kubefloworgv1beta1.Workspace, wsk *kubef
 			//       https://github.com/kubeflow/notebooks/issues/38
 			LastProbe: nil,
 		},
+		Services: buildServicesList(ws, wsk),
 	}
 	return workspaceModel
 }
 
-func buildPortsList(ws *kubefloworgv1beta1.Workspace, wsk *kubefloworgv1beta1.WorkspaceKind) []ImagePort {
-	var ports []ImagePort
+func buildServicesList(ws *kubefloworgv1beta1.Workspace, wsk *kubefloworgv1beta1.WorkspaceKind) []Service {
+	//nolint:prealloc
+	var services []Service
 
-	// Return an empty list if wsk is nil.
 	if !wskExists(wsk) {
-		return ports
+		return services
 	}
 
 	// Get the image configuration from the WorkspaceKind's PodTemplate options.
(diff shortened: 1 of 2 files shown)


workspaces/backend/internal/models/workspaces/types.go
```diff
@@ -84,6 +84,13 @@ type ImageConfig struct {
 	Current       OptionInfo     `json:"current"`
 	Desired       *OptionInfo    `json:"desired,omitempty"`
 	RedirectChain []RedirectStep `json:"redirect_chain,omitempty"`
+	Ports         []ImagePort    `json:"ports,omitempty"`
+}
+
+type ImagePort struct {
```
thesuperzapper, line 90: @yehudit1987 it might be more useful for the frontend if we do something more abstract like "services" which initially only includes "http" type services.

That makes it a bit easier to populate the front-end "connect" box, because rather than constructing the HTTP path in the front end we could just include the `httpPath` (relative to the UI base, not absolute) rather than constructing it in the front end.

Also, stuff like "port number" don't need to be included because this is hidden from the end user.

yehudit1987: Hey @thesuperzapper, thanks for the review! I totally agree with your suggestion—it makes more sense. I initially followed the frontend mock, but once the PR is approved, I'll update the frontend accordingly. Also, I’ve moved the location in the struct; hope that makes sense to you as well!

workspaces/backend/internal/models/workspaces/funcs.go
```diff
@@ -124,10 +125,46 @@ func NewWorkspaceModelFromWorkspace(ws *kubefloworgv1beta1.Workspace, wsk *kubef
...
+	}
+
+	// Get the image configuration from the WorkspaceKind's PodTemplate options.
+	imageConfig := wsk.Spec.PodTemplate.Options.ImageConfig
+
+	for _, val := range imageConfig.Values {
+		if len(val.Spec.Ports) == 0 {
+			continue
+		}
+		firstPort := val.Spec.Ports[0]
```
ederign, line 148:
```
// ports that the container listens on
	//   - if multiple ports are defined, the user will see multiple "Connect" buttons
	//     in a dropdown menu on the Workspace overview page
	// +kubebuilder:validation:MinItems:=1
	// +listType:="map"
	// +listMapKey:="id"
	Ports []ImagePort `json:"ports"`	
```

Shall we support here multiple ports here?

ederign: Or we deliberatelly are going to always use the first port here?

workspaces/backend/internal/models/workspaces/funcs.go
```diff
@@ -124,10 +125,46 @@ func NewWorkspaceModelFromWorkspace(ws *kubefloworgv1beta1.Workspace, wsk *kubef
...
+		portStr := fmt.Sprintf("%d", firstPort.Port)
+		displayName := firstPort.DisplayName
+		if displayName == "" {
+			displayName = val.Id
+		}
+		basePath := "/workspace"
```
ederign, line 154: Just out of my curiosity, where this /workspace path comes from?

ederign (changes requested):
@yehudit1987 thank you for the PR. Just two small comments there.

workspaces/backend/internal/models/workspaces/funcs.go
ederign: @thesuperzapper maybe you can chime in on this one. ^

I'm good to support just one for now and do a FUP PR on it.

yehudit1987: @ederign  I think it's a mistake (mine). Will fix it soon.

ederign: No worries @yehudit1987 ! Thank you for the PR

thesuperzapper:
@ederign @yehudit1987 I have made significant changes in https://github.com/kubeflow/notebooks/pull/213/commits/4e91e38bb6916390a7b788bb812bb270752557c8

I am happy to merge it now so we can unblock https://github.com/kubeflow/notebooks/pull/188 (which will need to be rebased to remove the backend updates).

PR #218, opened 2025-02-24
improve(ws): add controller-image parameter to e2e tests

etirelli:
Adding support for a command line parameter to set which image to use for e2e tests.

To run the tests with a given image, use the IMG parameter:

```sh
make test-e2e IMG=<some-registry>/workspace-controller:tag
```

Please note that if not defined, the default for e2e tests is now:

```makefile
IMG ?= controller:latest
```


workspaces/controller/Makefile
```diff
@@ -1,5 +1,6 @@
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
```
etirelli, line 2: I was not sure if I should change the default here to `ghcr.io/kubeflow/notebooks/workspace-controller:latest` to stay consistent with the previous e2e test behaviour or even create a separate parameter for it. Let me know please if you prefer that I do that.

PR #230, opened 2025-03-12
feat(ws): add `pendingRestart` field to workspace model

yehudit1987:
This PR is for adding missing field PendingRestart to workspace model for frontend use.


ederign (changes requested):
Hi @yehudit1987 can you please clarify why we need this for the frontend?

My understanding is https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/internal/models/workspaces/funcs.go#L54-L68

extract all the state and populate it on [state](https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/internal/models/workspaces/funcs.go#L107).

ederign (approved):

Issue #239, opened 2025-03-25
Allow mounting Secrets on Workspaces


thesuperzapper:
There has been a lot of discussion on how we can enable users to store credentials in Kubernetes secrets, and then make them available for usage in their Workspace Pods.

## Goal

Avoid users storing sensitive things within their notebooks (e.g. db/git credentials), and let them use Kubernetes secrets instead.

## Exclusions

- Not allow users to set arbitrary env-vars, only allow mounting secrets as files with projected volume mounts.

## Features

- We will let users mount keys from Secrets as files on the Workspace container.
- We will let users create new secrets (or mount existing secrets) when spawning/updating a Workspace.
- We will let users mount existing secrets when spawning/updating a Workspace.
- We will provide a "Secrets" page, similar to the existing "Volumes" page (which also needs to be re-implemented for NB2.0)

### STEP 1: Update Controller + Workspace CRD

We will need to add `spec.podTemplate.volumes.secrets[]` to the Workspace CRD:

```yaml
spec:
  podTemplate:
    volumes:
      secrets:
        - secretName: "workspace-secret"
          mountPath: "/secrets/my-secret"
          defaultMode: 420 # same as 0644 in octal
```

When we implement the reconciliation in the controller, we __MUST__ use  remember that Kubernetes [does not update secrets mounted with `subPath`](https://kubernetes.io/docs/concepts/storage/volumes/#secret) (except on container restarts), so its best to mount the whole secret as a folder, or possibly extend this to specific `items` in the future (but still in a folder for the secret).

### STEP 2: Extend Workspace Backend

We then need to:

- extend the Workspace GET/LIST/CREATE/UPDATE API to support the new secrets field
- implement API paths for listing Secrets in the namespace
    - __WARNING:__ we need to be careful when using controller-runtime with Secrets, because controller-runtime will try and cache all secrets in the cluster (to do its list-watch approach), this is a problem in very large clusters because they might have gigabytes of secrets and all would need to be in memory.
         - In other projects, a couple of approaches been taken.
         - OPTION 1: use a metadata-only cache, which does not cache the `data` part of the secret
         - OPTION 2: disable the list-watch cache for all secrets (fixes memory usage at the cost of Kube API calls, can be done in combination with option 1)
         - OPTION 3: use some kind of partial cache which only caches secrets with a specific label (probably will not work in our case)

The create workspace api might look like this after the updates:

```bash
# POST /api/v1/workspaces/{namespace}
[...]

PR #240, opened 2025-03-25
feat(ws): add `spec.podTemplate.volumes.secrets[]` to Workspace

thesuperzapper:
related https://github.com/kubeflow/notebooks/issues/239

This PR adds `spec.podTemplate.volumes.secrets[]` to the Workspace CRD, which lets users can mount secrets as volumes to their workspaces.

The Workspace CRD now has fields like this:

```yaml
spec:
  podTemplate:
    volumes:
      secrets:
        - secretName: "workspace-secret"
          mountPath: "/secrets/my-secret"
          defaultMode: 420 # same as 0644 in octal
```

Because errors relating to missing secrets (which cause the pod to get stuck in "pending" state) are only surfaced through Pod events, we now extract Pod events when determining the Workspace state and stateMessage.

Additionally, because those events might not exist when we first reconcile the Workspace (and none of the current WS reconcile triggers would cause a reconciliation when a new event is raised), we simply always requeue every 15 seconds when a Workspace's Pod is in the Pending state (and hopefully pick up a Warning even which will break us out of this loop).


vdksystem:
@thesuperzapper, why don't we check if secret exists as part of reconciliation process? Then we can set `ConfigError` status or similar right away. 
Or may be we have validation webhook which will prevent user from creating workspace with wrong config.
Also I wold add backoff to delayed reconciliation.

thesuperzapper:
> @thesuperzapper, why don't we check if secret exists as part of reconciliation process? Then we can set `ConfigError` status or similar right away. Or may be we have validation webhook which will prevent user from creating workspace with wrong config.

@vdksystem I think its safer to create the pod and see if it succeeds (e.g. eventual consistency). There are also some strange situations where a secret is only created when a pod tries to mount it. 

I also want to avoid the validation webhook rejecting "potentially" valid configs (especially because a secret could be deleted after the Workspace, which would lead to an "update" being rejected on a previously acceptable manifest).

>  Also I wold add backoff to delayed reconciliation.

I think the only sensible way to achieve this would be to make the reconciliation loop "fail" in the case that we have a pending Pod with no events (that is, return non-nill err, but still update the Workspace state to Pending before doing that). 

This would kick us into the exponential failure backoff of controller-runtime (which is probably way to aggressive, as discussed in https://github.com/kubeflow/notebooks/issues/245). However, this would also be quite noisy in the controller logs, because each of the failed reconciles will show as an error even though we are really "waiting" for Pod events to be created.

Another approach is not returning an error, but storing a "retry count" in the status of the Workspace. However, this will cause a whole bunch of updates to the Workspace resource during the retry (which will also trigger other resources to reconcile, e.g. the WorkspaceKind used by the Workspace).

Not sure which is "best", do you have any thoughts?

vdksystem:
@thesuperzapper agree with eventual consistency. Optional - we can implement kinda soft check. Not sure if it make sense to adjust status, but at least emit warning event.

With retries, we can use `creationTimestamp` and implement backoff logic around time since created (e.g. retyInterval = 15s * minutes since created).

I think all of that would make sense only at scale, can be ignored for now.

yashpal2104:
Does this PR require help?

andyatmiami (commented):
@thesuperzapper 

pretty much everything looks good here but wanted to get one clarification on behavior I am observing (albeit for now I don't think this needs to block anything and its _working as designed_ )

Scenario:

1. Create `workspace` referencing a non-existent secret
2. `workspace` enters `Error` state as expected
3. create the (missing) referenced secret
4. `workspace` enters `Running` state
5. delete the secret
6. `workspace` remains in running state while `MountVolume.Setup` events start occurring
7. `pause` the `workspace` (really to just trigger a reconcile loop)
8. `workspace` enters `Paused` state
9. `unpause` the `workspace`
10. `workspace` enters `Error` state

This **IS** essentially expected k8s behavior for (for instance) deployments that reference a secret that is deleted **after** is mounted.  

Just want confirmation this is as you would expect - and/or if we should track some type of "tech debt" item to "improve" (?) somehow the `status` reporting on `Workspace` as there does seem to be an Event available we could provide more information to user on
- now... what that "state" should be if we wanted to expose this is a wholly different question... as it would be a "future `Error` is looming if/when workload restarts" 😎

Issue #298, opened 2025-04-30
API to start, stop and restart a workspace


ederign:
TBD

ederign:
Shall we use a single update endpoint?

andyatmiami:
/close not-planned

we are separating out the ability to start/stop from "restart" - so closing this issue in favor of more granular tracking.

PR #340, opened 2025-05-14
feat(ws): add workspace pause actions backend API

andyatmiami:
related: https://github.com/kubeflow/notebooks/issues/298

- Added PauseActionWorkspaceHandler to handle pausing or unpausing a given workspace
- Introduced single new route for starting and pausing workspaces in the API.
    - `api/v1/workspaces/{namespace}/{name}/actions/pause`
			- pausing or unpausing operation is specified in the request payload
- Created a new WorkspaceActionPauseEnvelope type for successful responses.
- Leveraging JSONPatch / client.RawPatch to ensure Workspace in "valid state" before attempting action
    - for `start`: `spec.paused` must be `true`, and `status.state` must be `Paused`
    - for `pause`: `spec.paused` must be `false`
        - note: I would love to have a `status.state` check here of `status.state != Paused`, but that type of comparison is not supported in [JSONPatch](https://datatracker.ietf.org/doc/html/rfc6902#section-4.6)
- Added tests for the new API, including success and error cases.
- Updated README/OpenAPI documentation to include the new endpoints.

---

As an interesting "edge case" worth calling out, the following payload is currently honored by the API:
```
{
  "data": {}
}
```

Given the `WorkspaceActionPause` struct is simply `{"paused": true|false}`, the "empty" Envelope presented above deserializes the JSON using the zero value of `bool` (which is `false`).

Our validation today is always performed against the **deserialized** object, and as such impossible to distinguish the following cases w.r.t actual JSON request payload:
```
{
  "data": {}
}
```

vs

```
{
  "data": {
    "paused": false
  }
}
```

The effort and (relative) complexity to prevent this and return a `422` in this scenario was not deemed "worth it" for the time being.  As a result, a test case has been added for this specific scenario to at minimum document this "strange" behavior.
- Clients, however, should **NOT** rely on this behavior and always provide a fully defined `WorkspaceActionPause` JSON object to ensure future compatibility.

Rendered Swagger doc:

image

### DIFF
--- a/workspaces/backend/api/workspace_actions_handler.go
+++ b/workspaces/backend/api/workspace_actions_handler.go
@@ -1,0 +1,128 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/auth"
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/helper"
+	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/workspaces/actions"
+	repository "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/workspaces"
+)
+
+type WorkspaceActionPauseEnvelope Envelope[*models.WorkspaceActionPause]
+
+// PauseActionWorkspaceHandler handles setting the paused state of a workspace.
+//
+//	@Summary		Pause or unpause a workspace
+//	@Description	Pauses or unpauses a workspace, stopping or resuming all associated pods.
+//	@Tags			workspaces
+//	@Accept			json
+//	@Produce		json
+//	@Param			namespace		path		string							true	"Namespace of the workspace"	extensions(x-example=default)
+//	@Param			workspaceName	path		string							true	"Name of the workspace"			extensions(x-example=my-workspace)
+//	@Param			body			body		WorkspaceActionPauseEnvelope	true	"Intended pause state of the workspace"
+//	@Success		200				{object}	WorkspaceActionPauseEnvelope	"Successful action. Returns the current pause state."
+//	@Failure		400				{object}	ErrorEnvelope					"Bad Request."
+//	@Failure		401				{object}	ErrorEnvelope					"Unauthorized. Authentication is required."
+//	@Failure		403				{object}	ErrorEnvelope					"Forbidden. User does not have permission to access the workspace."
+//	@Failure		404				{object}	ErrorEnvelope					"Not Found. Workspace does not exist."
+//	@Failure		413				{object}	ErrorEnvelope					"Request Entity Too Large. The request body is too large."
+//	@Failure		415				{object}	ErrorEnvelope					"Unsupported Media Type. Content-Type header is not correct."
+//	@Failure		422				{object}	ErrorEnvelope					"Unprocessable Entity. Workspace is not in appropriate state."
+//	@Failure		500				{object}	ErrorEnvelope					"Internal server error. An unexpected error occurred on the server."
+//	@Router			/workspaces/{namespace}/{workspaceName}/actions/pause [post]
+func (a *App) PauseActionWorkspaceHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	namespace := ps.ByName(NamespacePathParam)
+	workspaceName := ps.ByName(ResourceNamePathParam)
+
+	var valErrs field.ErrorList
+	valErrs = append(valErrs, helper.ValidateFieldIsDNS1123Subdomain(field.NewPath(NamespacePathParam), namespace)...)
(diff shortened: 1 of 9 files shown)


harshad16 (commented):
Great work :100: 
Tested with cli:
Screenshot from 2025-05-20 16-50-35

workspaces/backend/api/workspace_actions_handler.go
```diff
@@ -0,0 +1,142 @@
...
+			a.notFoundResponse(w, r)
+			return
+		}
+		a.serverErrorResponse(w, r, err)
+		return
+	}
+
+	// Get the workspace to check its paused state
+	workspace, err := a.repositories.Workspace.GetWorkspace(r.Context(), namespace, workspaceName)
+	if err != nil {
```
ederign, line 83: On this call, we are not checking anything, are we?

ederign: Andy, thanks for the clarification on the chat. You are right about it. Just ignore these comments.

andyatmiami: So, for transparency here.. I was slightly incorrect in my comments above...

It seems the `RawPatch` command leveraged in the application code **is itself** asynchronously applied by k8s... so there is a race condition present here in this version of the code...

At the time we call this `GetWorkspace` function - we **cannot* guarantee that the `Patch` has been applied.. and as such the `.spec.paused` attribute might not have been updated yet...

I'm honestly not sure how to handle this...

Blocking and waiting seems kinda gross (to me) - but I also feel _dirty_ hard-coding the `.data.paused` value we return in the PausedState response struct...

workspaces/backend/api/workspace_actions_handler.go
```diff
@@ -0,0 +1,142 @@
...
+		a.serverErrorResponse(w, r, err)
+		return
+	}
+
+	// Return 200 OK with pause state
+	err = a.WriteJSON(w, http.StatusOK, PauseStateEnvelope{
```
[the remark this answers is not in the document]
andyatmiami: see comment above: https://github.com/kubeflow/notebooks/pull/340#discussion_r2114155301

`202` might be the safer option - but still need to then answer/determine if/how we want to return the `paused` attribute in the response

workspaces/backend/api/workspace_actions_handler.go
ederign: We should definitely not block it. For the UI, maybe we can use the status to notify user that something is being processed. 

I would not hardcode the value, because, for instance, on start, things can fail, and if we hardcode, it could be inconsistent.

Why don't return 202 and something like
{data:
  "status": "accepted",
  "message": "bla",
  "
}

jiridanek: > It seems the RawPatch command leveraged in the application code is itself asynchronously applied by k8s... so there is a race condition present here in this version of the code...

how do you know that

> At the time we call this GetWorkspace function - we cannot guarantee that the Patch has been applied.. and as such the .spec.paused attribute might not have been updated yet...

on success,

```go
if err := r.client.Patch(ctx, workspace, client.RawPatch(types.JSONPatchType, patchBytes)); err != nil {
```

modifies `workspace` with the result of the patching; you don't need to `Get` the workspace afterwards and risk getting inconsistent results due to client side caching if that's enabled; and when `Patch` call succeeded, you are guaranteed that the patch was not rejected, no?

* https://github.com/kubernetes-sigs/controller-runtime/issues/343#issuecomment-469439104

harshad16 (approved):
The new changes looks great, with return payload 
Tested it with running system:
Screenshot from 2025-06-18 06-14-10
Works as intended.

great work :100: 

/lgtm

thesuperzapper (changes requested):
I left a few comments (we were on a call during the review so @andyatmiami should be aware of what to do and we were in agreement on the changes).

andyatmiami:
### Verifying behavior

Check existing state
```
➜ backend/ git:(feat/workspace-start-api) $ kubectl get workspaces -A
NAMESPACE   NAME                         STATE
default     jupyterlab-scipy-workspace   Running
default     jupyterlab-workspace         Running
```

Pause running `jupyterlab-workspace` workspace
```
➜ backend/ git:(feat/workspace-start-api) $ curl -X 'POST' \
  'http://localhost:4000/api/v1/workspaces/default/jupyterlab-workspace/actions/pause' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "data": {
    "paused": true
  }
}'

{
	"data": {
		"paused": true
	}
}
```

Verify `jupyterlab-workspace` now paused
```
➜ backend/ git:(feat/workspace-start-api) $ kubectl get workspaces -A
NAMESPACE   NAME                         STATE
default     jupyterlab-scipy-workspace   Running
default     jupyterlab-workspace         Paused
```

Start (unpause) paused `jupyterlab-workspace` workspace
```
➜ backend/ git:(feat/workspace-start-api) $ curl -X 'POST' \
  'http://localhost:4000/api/v1/workspaces/default/jupyterlab-workspace/actions/pause' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "data": {
    "paused": false
  }
}'

{
	"data": {
		"paused": false
	}
}
```

Verify `jupyterlab-workspace` now running (unpaused)
```
➜ backend/ git:(feat/workspace-start-api) $ kubectl get workspaces -A
NAMESPACE   NAME                         STATE
default     jupyterlab-scipy-workspace   Running
default     jupyterlab-workspace         Running
```

Attempt to start (unpause) running `jupyterlab-workspace` workspace
```
➜ backend/ git:(feat/workspace-start-api) $ curl -X 'POST' \
  'http://localhost:4000/api/v1/workspaces/default/jupyterlab-workspace/actions/pause' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "data": {
    "paused": false
  }
}'
{
	"error": {
		"code": "422",
		"message": "workspace is in an invalid state for this operation",
		"cause": {}
	}
}
```

Attempt to start (unpause) running `jupyterlab-workspace` workspace exercising "empty" Envelope edge case
```
➜ backend/ git:(feat/workspace-start-api) $ curl -X 'POST' \
  'http://localhost:4000/api/v1/workspaces/default/jupyterlab-workspace/actions/pause' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "data": {}
}'               
{
	"error": {
		"code": "422",
		"message": "workspace is in an invalid state for this operation",
		"cause": {}
	}
}
```

Verify state
```
➜ backend/ git:(feat/workspace-start-api) $ kubectl get workspaces -A
NAMESPACE   NAME                         STATE
default     jupyterlab-scipy-workspace   Running
default     jupyterlab-workspace         Running
```

Issue #318, opened 2025-05-09
[TASK] Persist last used namespace


paulovmr:
### Checks

- [x] I have searched the [existing issues](https://github.com/kubeflow/notebooks/issues).
- [x] My request is related to one of the components in the [`kubeflow/notebooks`](https://github.com/kubeflow/notebooks) repository.

### Motivation

Every time the user access the application, a random namespace is selected, and they have to manually select the one they want to see.

### Implementation

Save the last used namespace in a cookie, so the same one gets loaded once the user returns to the application.

### Are you willing & able to help?

- [x] I am able to submit a PR!
- [x] I can help test the feature!

thaorell:
I can take this issue

PR #341, opened 2025-05-14
feat(ws): Persist last used namespace across refreshes and tabs

thaorell:
closes: #[318](https://github.com/kubeflow/notebooks/issues/318)

Previous behaviors: Always use the first namespace from fetch as selected namespace
Features: 

1. Persist selected namespace across refreshes. If no such namespace exists, select the one at index 0 of the `namespaces` array
2. If namespace is changed on one tab, that change will reflect on other tabs. That will ensure consistency and won't confuse users on what the "last" selected namespace was.

### DIFF
--- a/workspaces/frontend/src/app/context/NamespaceContextProvider.tsx
+++ b/workspaces/frontend/src/app/context/NamespaceContextProvider.tsx
@@ -1,11 +1,16 @@
 import React, { useState, useContext, ReactNode, useMemo, useCallback } from 'react';
 import useMount from '~/app/hooks/useMount';
 import useNamespaces from '~/app/hooks/useNamespaces';
+import { useStorage } from './BrowserStorageContext';
+
+const storageKey = 'kubeflow.notebooks.namespace.lastUsed';
 
 interface NamespaceContextType {
   namespaces: string[];
   selectedNamespace: string;
   setSelectedNamespace: (namespace: string) => void;
+  lastUsedNamespace: string;
+  updateLastUsedNamespace: (value: string) => void;
 }
 
 const NamespaceContext = React.createContext<NamespaceContextType | undefined>(undefined);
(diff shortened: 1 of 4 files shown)


workspaces/frontend/src/app/context/NamespaceContextProvider.tsx
```diff
@@ -18,6 +19,8 @@ export const useNamespaceContext = (): NamespaceContextType => {
   return context;
 };
 
+const storageKey = 'kubeflow.notebooks.last.used.namespace';
```
paulovmr, line 22:
```suggestion
const storageKey = 'kubeflow.notebooks.namespace.lastUsed';
```

paulovmr: Please consider making this constant available through this context and use it in the `NamespaceSelector.tsx` file instead of duplicating the `storageKey` string there.

workspaces/frontend/src/app/context/NamespaceContextProvider.tsx
```diff
@@ -26,20 +29,24 @@ export const NamespaceContextProvider: React.FC<NamespaceContextProviderProps> =
   const [namespaces, setNamespaces] = useState<string[]>([]);
   const [selectedNamespace, setSelectedNamespace] = useState<string>('');
   const [namespacesData, loaded, loadError] = useNamespaces();
+  const [lastUsedNameSpace, setLastUsedNamespace] = useStorage<string>(storageKey, '');
```
paulovmr, line 32:
```suggestion
  const [lastUsedNamespace, setLastUsedNamespace] = useStorage<string>(storageKey, '');
```
Please also change its usages.

paulovmr (changes requested):

paulovmr (approved):
/lgtm

Issue #335, opened 2025-05-13
Notebooks V2 GET /api/v1/workspacekinds Backend API supports filtering


andyatmiami:
## Description
We want all our Backend APIs to eventually support the myriad of filtering options required by the frontend.  However, we want to start small and focus specifically on `GET /api/v1/workspacekinds` in order to agree upon a framework/implementation the upstream community is comfortable with.

Best I can tell - the [Kubeflow Model Registry code base]([https://github.com/kubeflow/model-registry](https://github.com/kubeflow/model-registry/pull/918)) supports limited filtering today.  The implementation seems to be based on specific query parameters named after the attribute to be filtered on.  For now - we should plan on following this pattern.

Example:
- `GET /api/v1/workspacekinds?name=jupyter&other=foo`

**Please be mindful that acceptance criteria may change and any opened PR may require rework based on community feedback. I (@andyatmiami) am trying to outline this work but have not been able to get feedback on specific implementation guidance we should target - so I'm doing a bit of guesswork!**

 

## Acceptance Criteria
- `GET /api/v1/workspacekinds` supports filtering of results based on:
    - `name`
    - ⚠ _TODO: explicitly list any other fields we want to initially target_
- The filtering operation in the `backend` is performed with a `startsWith()` comparison
- Attempting to provide a filtering query parameter for an attribute that does not support filtering should result in a `400` response with a helpful error message adhering to the error response schema
- Unit test(s) exist confirming happy path behavior
- Auto-generated Swagger documentation includes details about the query parameter

PR #366, opened 2025-05-21
feat(ws): Notebooks V2 GET /api/v1/workspacekinds Backend API support…

abdallahsamabd:
fix for #335 
adding filter option for the returned workspacekinds
an example
 curl -i http://localhost:4000/api/v1/workspacekinds?filter=name::abc,status::active
 will return workinds that have name prefix abc and status active (deprecated false)

### DIFF
--- a/workspaces/backend/api/workspacekinds_handler.go
+++ b/workspaces/backend/api/workspacekinds_handler.go
@@ -19,6 +19,7 @@ package api
 import (
 	"errors"
 	"net/http"
+	"strings"
 
 	"github.com/julienschmidt/httprouter"
 	kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"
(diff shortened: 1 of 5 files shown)


workspaces/backend/api/workspacekinds_handler.go
```diff
@@ -121,7 +123,55 @@ func (a *App) GetWorkspaceKindsHandler(w http.ResponseWriter, r *http.Request, _
 		a.serverErrorResponse(w, r, err)
 		return
 	}
-
+	// ========================== Filtering =======================
+	filterParams := r.URL.Query().Get("filter")
```
andyatmiami, line 127: @abdallahsamabd 

My apologies here for confusion... initial (internal) guidance I gave you was _imprecise_ - and while I was more specific in the #335 issue itself - I should have followed up with you to call out the discrepancy clearly...

for our filtering syntax - #335 states (highlighted):

image


So instead of a singular `filter` query parameter - we will want parameters for each attribute (following the paradigm established in model registry)

workspaces/backend/api/workspacekinds_handler.go
```diff
@@ -121,7 +123,55 @@ func (a *App) GetWorkspaceKindsHandler(w http.ResponseWriter, r *http.Request, _
...
+			var filtered []models.WorkspaceKind
+			parts := strings.Split(filterParam, "::")
+			if len(parts) != 2 {
+				a.failedValidationResponse(w, r, "Invalid filter syntax: "+filterParam, nil, nil)
+				return
+			}
+
+			filterKey := parts[0]
+			filterVal := parts[1]
+			for _, wk := range workspaceKinds {
```
andyatmiami, line 140: Filtering should reduce load on server by aggressively filtering out results as early as possible....

We want to push the "filter" **all the way down** into the controller for a scalable design...

Having k8s return **EVERY RECORD** for us to then have to iterate in a tight loop to restrict the result set is too inefficient

We should be parsing the query parameters and placing some type of `struct` on `Context` prior to this call:
- https://github.com/kubeflow/notebooks/pull/366/files#diff-3ad5f66021fc51045196e93c26657b03fd06a7d004e2e09acb4c0898045055f7R121

This way, when the results are returned from the `controller` - no post-processing / wasted space is introduced.

andyatmiami: Additionally, the solution as presented is wholly bound and tightly coupled to the `GetWorkspaceKindsHandler` function.

Filtering is a cross-cutting concern that will apply to numerous APIs... would be good to consider how to implement some foundational structs/middleware to remove the need to copy/paste a lot of boilerplate logic as we introduce filtering in other areas.

Issue #336, opened 2025-05-13
Notebooks V2 GET /api/v1/workspacekinds Backend API supports sorting


andyatmiami:
## Description
We want all our Backend APIs to eventually support the myriad of sorting options required by the frontend.  However, we want to start small and focus specifically on `GET /api/v1/workspacekinds` in order to agree upon a framework/implementation the upstream community is comfortable with.

We can use the Model Registry work as an example/inspiration:
- https://github.com/kubeflow/model-registry/blob/main/internal/server/openapi/api_model_registry_service.go#L708-L722

Example:
- `GET /api/v1/workspacekinds?orderBy=name&sortOrder=ASC`
    - `sortOrder=DESC`also supported

**Please be mindful that acceptance criteria may change and any opened PR may require rework based on community feedback. I (@andyatmiami) am trying to outline this work but have not been able to get feedback on specific implementation guidance we should target - so I'm doing a bit of guesswork!**

 

## Acceptance Criteria
- `GET /api/v1/workspacekinds` supports sorting of results in `ASC` or `DESC` order based on:
    - `name`
    - ⚠ _TODO: explicitly list any other fields we want to initially target_
- If no sorting query parameters defined - order of results should default to `name` `ASC`
- Attempting to provide a sorting query parameter for an attribute that does not support sorting should result in a `400` response with a helpful error message adhering to the error response schema
- Providing a value that is not either `ASC` or `DESC` for the ordering query parameter should result in a `400` response with a helpful error message adhering to the error response schema
- Unit test(s) exist confirming happy path behavior
- Auto-generated Swagger documentation includes details about the query parameter

Noa-limoy:
Hey @Adembc 
I saw that you’ve taken the issue and assigned it to yourself. I’ve already worked on it and my PR is prepared.
Would you mind if I go ahead and continue with it?

Thanks and regards,
Noa

PR #367, opened 2025-05-21
feat(ws): Notebooks V2 GET /api/v1/workspacekinds Backend API support sorting #336

noalimoy:
implementation for #336

Adds support for sorting in GET /api/v1/workspacekinds endpoint. 
for example: 
curl -i "http://localhost:4000/api/v1/workspacekinds?orderBy=status&sortOrder=asc"
curl -i "http://localhost:4000/api/v1/workspacekinds?orderBy=name&sortOrder=desc"

Changes:
- Introduced support for 'orderBy' (name or status) and  'sortOrder' (asc or desc) query parameters.
- Added unit test to verify the sorting api call.
- Auto-generated swagger documantion now include details about the sorting cal with new parameters.

 closes: #336

### DIFF
--- a/workspaces/backend/api/app.go
+++ b/workspaces/backend/api/app.go
@@ -37,6 +37,7 @@ const (
 
 	NamespacePathParam    = "namespace"
 	ResourceNamePathParam = "name"
+	StatusPathParam       = "status"
 
 	// healthcheck
 	HealthCheckPath = PathPrefix + "/healthcheck"
--- a/workspaces/backend/api/workspacekinds_handler.go
+++ b/workspaces/backend/api/workspacekinds_handler.go
@@ -19,6 +19,8 @@ package api
 import (
 	"errors"
 	"net/http"
+	"net/url"
+	"sort"
 
 	"github.com/julienschmidt/httprouter"
 	kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"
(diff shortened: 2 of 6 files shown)


workspaces/backend/api/app.go
```diff
@@ -37,6 +37,7 @@ const (
 
 	NamespacePathParam    = "namespace"
 	ResourceNamePathParam = "name"
+	StatusPathParam       = "status"
```
andyatmiami, line 40:
```suggestion
	StatusQueryParam       = "status"
```

workspaces/backend/api/workspacekinds_handler.go
```diff
@@ -122,6 +126,37 @@ func (a *App) GetWorkspaceKindsHandler(w http.ResponseWriter, r *http.Request, _
...
+	// ========================== Sorting =========================
+
+	workspaceKinds = GetWorkspaceKindsSorted(workspaceKinds, query)
+
 	responseEnvelope := &WorkspaceKindListEnvelope{Data: workspaceKinds}
 	a.dataResponse(w, r, responseEnvelope)
 }
+
+// GetWorkspaceKindsSorted sorts the given list of WorkspaceKind objects based on the query parameters.
+func GetWorkspaceKindsSorted(workspaceKinds []models.WorkspaceKind, query url.Values) []models.WorkspaceKind {
```
andyatmiami, line 140: The solution as presented is wholly bound and tightly coupled to the GetWorkspaceKindsHandler function.

Sorting is a cross-cutting concern that will apply to numerous APIs... would be good to consider how to implement some foundational structs/middleware to remove the need to copy/paste a lot of boilerplate logic as we introduce filtering in other areas.

workspaces/backend/api/workspacekinds_handler.go
```diff
@@ -122,6 +126,37 @@ func (a *App) GetWorkspaceKindsHandler(w http.ResponseWriter, r *http.Request, _
 		return
 	}
 
+	query := r.URL.Query()
+
+	// ========================== Sorting =========================
+
+	workspaceKinds = GetWorkspaceKindsSorted(workspaceKinds, query)
```
andyatmiami, line 133: Having to iterate over objects returned by k8s as a means to order the result set is too inefficient

We should be parsing the query parameters and placing some type of struct on Context prior to this call:

https://github.com/kubeflow/notebooks/pull/367/files#diff-3ad5f66021fc51045196e93c26657b03fd06a7d004e2e09acb4c0898045055f7R123

This way, when the results are returned from the controller - no post-processing / wasted space is introduced.  Additionally, this logic would break once/when we introduce pagination - as this logic would only handle sorting "the current page" .

workspaces/backend/api/workspacekinds_handler.go
andyatmiami: Consider cases like sorting on `name` or sorting on timestamp fields inherent in all k8s object (note the timestamp use case out of scope to this work for now!)

defining an interface (and implementing said methods on our relevant models) would allow common sorting behavior to be defined centrally (in `helpers.go` or something) so each handler would not need to "re-invent" the logic - focusing only on providing custom logic when it is actually custom

Issue #352, opened 2025-05-16
[TASK] Notebooks V2 Backend WorkspaceKind response schema should include "counts" information


andyatmiami:
## Description

The Workspace Kind CRD has "metrics" related to the counts of workspaces that reference a given workspace kind.

- [CRD](https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/controller/config/crd/bases/kubeflow.org_workspacekinds.yaml#L4536-L4598)
- [Go Structs](https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/controller/api/v1beta1/workspacekind_types.go#L504-L537) 

Unfortunately, these concepts are not present in the Backend API schema:
- [Backend Structs](https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/internal/models/workspacekinds/types.go#L19-L29)

As the Frontend requires rendering this data to inform the user - its essential the counts defined in the CRD are present in the Backend API response payload.

## Acceptance Criteria

- Notebooks V2 Backend API for WorkspaceKind Responses include "counts"-related data defined in the CRD
    - Applies to:
        - GET /api/v1/workspacekinds
        - GET /api/v1/workspacekinds/{kind}
- Structure of the object in the API response should mimic that of the CRD
- Unit tests exist asserting presence of this information in the backend response
- Auto-generated Swagger documentation includes details about the new schema additions

roee1313:
Hi @Adembc 
Ive seen you assigned yourself to this issue, 
I already started working on it and have a PR ready.. so if that's ok ill take it from here ok?
Didnt got to assign myself to that yet.. sorry

PR #368, opened 2025-05-21
feat(ws): add ws counts to backend wsk model

roee1313:
This PR solves issue #352

### DIFF
--- a/workspaces/backend/internal/models/workspacekinds/funcs.go
+++ b/workspaces/backend/internal/models/workspacekinds/funcs.go
@@ -36,6 +36,8 @@ func NewWorkspaceKindModelFromWorkspaceKind(wsk *kubefloworgv1beta1.WorkspaceKin
 			podAnnotations[k] = v
 		}
 	}
+	statusImageConfigMap := buildOptionMetricsMap(wsk.Status.PodTemplateOptions.ImageConfig)
+	statusPodConfigMap := buildOptionMetricsMap(wsk.Status.PodTemplateOptions.PodConfig)
 
 	// TODO: icons can either be a remote URL or read from a ConfigMap.
 	//       in BOTH cases, we should cache and serve the image under a path on the backend API:
@@ -60,6 +62,10 @@ func NewWorkspaceKindModelFromWorkspaceKind(wsk *kubefloworgv1beta1.WorkspaceKin
 		Hidden:             ptr.Deref(wsk.Spec.Spawner.Hidden, false),
 		Icon:               iconRef,
 		Logo:               logoRef,
+		// TODO: in the future will need to support including exactly one of clusterMetrics or namespaceMetrics based on request context
+		ClusterMetrics: clusterMetrics{
+			Workspaces: wsk.Status.Workspaces,
+		},
 		PodTemplate: PodTemplate{
 			PodMetadata: PodMetadata{
 				Labels:      podLabels,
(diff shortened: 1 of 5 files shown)


andyatmiami (changes requested):
I managed to (briefly) talk with Mathew during the WG call today (Thursday, May 29th) and he stated he was comfortable embedding this "counts" information more naturally inline with the response schema (and not worrying about adhering to a 1:1 structural mapping w.r.t how the information is presented from k8s)

The below image represents my proposal on how this "embedding" could work.  A couple notes:
- The same schema should be applied to the "list" as well as the "get-by-id" endpoints
    - For sake of brevity - I am only showing the "get-by-id" payload below
- I opted to "co-locate" the counts information closest to relevant data within the response.  This is certainly _most clean_ for the `values` arrays - but then leads to a root `workspacesCount` attribute at the root of the `WorkspaceKind` element as there isn't a more suitable place for it.  
- For sake of consistency/readability - I chose to re-use the same attribute name throughout (`workspacesCount`) - with the "pattern" being `<child resource name>Counts`... in the event this paradigm crops up elsewhere in the API - we could reuse this design _philosophy_
- Certainly open to other's opinions on if this is the most appropriate design.  I had considered a `counts: {....}` object as well... but chose not to advocate for that option as its too redundant/nested (having to repeat context like `imageConfig` as well as the IDs in order for a client to then have to perform the mapping to relevant object to display inline as we do in our table view today)

image



FYI: @thesuperzapper @ederign @paulovmr

workspaces/backend/internal/models/workspacekinds/funcs.go
```diff
@@ -153,3 +156,12 @@ func buildOptionRedirect(redirect *kubefloworgv1beta1.OptionRedirect) *OptionRed
 		Message: message,
 	}
 }
+
+func FindConfigWorkspacesById(configs []kubefloworgv1beta1.OptionMetric, targetId string) int32 {
```
andyatmiami, line 160: For optimal efficiency - I worry about this pattern of iterating through an array while we are **already** iterating over an array that gives us O(n^2)

We can improve this by iterating over the respective `[]kubefloworgv1beta1.OptionMetric` once each (for imageConfig and podConfig) and building up a map object that we can then pass into the `build...ConfigValues`

This ensures while we are iterating over the PodConfig/ImageConfig `.Values` - we can get O(1) lookups for the respective counts.

Although being done in a slightly different context - you can see examples of "map initialization" in the`NewWorkspaceKindModelFromWorkspaceKind` function:
- https://github.com/kubeflow/notebooks/pull/368/files#diff-5a4650065c6e40d76e86fc1df9189efbbdc0d93250c74928b71601e13eac2890R30

So, feels like this new use case also "fits"...

andyatmiami:
/lgtm

Re-testing code after latest changes:
- verified `GET api/v1/workspacekinds` includes `workspacesCount` in response for all elements
- verified `GET api/v1/workspacekinds/{name}` includes `workspacesCount` in response

also confirmed Swagger UI properly includes these new attributes in schema:
image

andyatmiami:
@thesuperzapper - PR ready / looking for your review/approval now - thanks!

workspaces/backend/internal/models/workspacekinds/types.go
```diff
@@ -58,12 +63,13 @@ type ImageConfig struct {
...
-	Labels      []OptionLabel   `json:"labels"`
-	Hidden      bool            `json:"hidden"`
-	Redirect    *OptionRedirect `json:"redirect,omitempty"`
+	Id             string          `json:"id"`
+	DisplayName    string          `json:"displayName"`
+	Description    string          `json:"description"`
+	Labels         []OptionLabel   `json:"labels"`
+	Hidden         bool            `json:"hidden"`
+	Redirect       *OptionRedirect `json:"redirect,omitempty"`
+	ClusterMetrics clusterMetrics  `json:"clusterMetrics"`
```
andyatmiami, line 72:
```suggestion
	ClusterMetrics clusterMetrics  `json:"clusterMetrics,omitempty"`
```

workspaces/backend/internal/models/workspacekinds/funcs.go
```diff
@@ -93,12 +107,15 @@ func buildImageConfigValues(imageConfig kubefloworgv1beta1.ImageConfig) []ImageC
 			Labels:      buildOptionLabels(option.Spawner.Labels),
 			Hidden:      ptr.Deref(option.Spawner.Hidden, false),
 			Redirect:    buildOptionRedirect(option.Redirect),
+			ClusterMetrics: clusterMetrics{
```
andyatmiami, line 110: Maybe include the `// TODO` disclaimer here as outlined in https://github.com/kubeflow/notebooks/pull/368/files#r2178441024

workspaces/backend/internal/models/workspacekinds/funcs.go
```diff
@@ -109,6 +126,9 @@ func buildPodConfigValues(podConfig kubefloworgv1beta1.PodConfig) []PodConfigVal
...
```
andyatmiami, line 129: Maybe include the `// TODO` disclaimer here as outlined in https://github.com/kubeflow/notebooks/pull/368/files#r2178441024

workspaces/backend/internal/models/workspacekinds/funcs.go
```diff
@@ -71,18 +77,26 @@ func NewWorkspaceKindModelFromWorkspaceKind(wsk *kubefloworgv1beta1.WorkspaceKin
...
 }
 
-func buildImageConfigValues(imageConfig kubefloworgv1beta1.ImageConfig) []ImageConfigValue {
+func buildOptionMetricsMap(metrics []kubefloworgv1beta1.OptionMetric) map[string]int32 {
+	resultMap := make(map[string]int32)
+	for _, metric := range metrics {
+		resultMap[metric.Id] = metric.Workspaces
+	}
+	return resultMap
+}
```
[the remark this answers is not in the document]
andyatmiami: given these objects we are handling are inherently small (and bounded to be reasonably sure they will always be small) - the code here as written is sufficient.

andyatmiami (approved):

thesuperzapper:
@roee1313 thanks for this, great to have your first PR merged.

/lgtm
/approve

PR #397, opened 2025-06-02
ci(ws): run tests when the github actions workflow definition changes

jiridanek:


andyatmiami (approved):

thesuperzapper:
@jiridanek thanks for this, but I think we should just work under the assumption that test is never run when the test definition changes, and act accordingly.

Also, tying the name of the file to the content of the file makes refactoring more error-prone, so I will close this for now.

/close

Issue #379, opened 2025-05-22
[TASK] Persist coverage report for backend directory


andyatmiami:
## Description

_Eventually_ - we want safeguards in place on the repo to ensure our code coverage does not decrease over time.  These safeguards would prevent PRs from being merged in the event it was detected the decrease in coverage was "too severe" (TBD what any such threshold would be).

As an incremental step in realizing this goal, we want to store the present state of our coverage to easily compare against any new changes that could have an impact.  Relying on recomputing this value on demand is challenging for a developer who is contributing code - as they would need to switch branches/stash changes to get the "baseline" value.

Therefore, we want to store the output of the `backend/` coverage report generated by `go` within that directory.  This output should be computed upon a commit being introduced to the `notebooks-v2` branch.  A GitHub Action is the obvious solution to reliably trigger this behavior.

Future work will build on top of this to parse and act on this information.  This initial issue is simply about getting the implementation in place to effectively surface the "baseline" measurements in a source-controlled manner.

## Acceptance Criteria

- When code is merged to the `notebooks-v2` branch that modifies the contents of the `workspaces/backend` directory, a `coverage.baseline` file is computed and persisted to the `notebooks-v2` branch within the `workspaces/backend` directory

mkoushni:
i will take it

PR #402, opened 2025-06-03
feat(ws): persist coverage report for backend directory #379

mkoushni:


andyatmiami (changes requested):
@mkoushni 

Feel free to reach out to me to sync up as I worry the acceptance criteria in the GH issue maybe wasn't as clear as it should have been 😇 

There are numerous changes I think we need to make to achieve what I hope to get out of this work item... I did a very quick hackjob ( **UNTESTED** ) to highlight areas I think we need to rework:

image
- the `.yml` file needs to reside in `.github/workflows` in order for it to be honored
- we want the action to fire after code is **pushed** to the `notebooks-v2` branch
    - i.e. _after_ a PR is merged, this action should fire
- lets not hardcode the version of `go` in the `.yml` file - its too brittle... the example pictured above copies the `Setup Go` logic from the `ws-backend-test.yml` action - which seems a more flexible/maintainable approach
- for this initial work - we simply want to **commit** the `coverage.baseline` file to `workspaces/backend` ... this file can/will then always act as a "source of truth" w.r.t. what the current state of coverage is for the `notebooks-v2` branch.
    - in the future, we can build up additional github actions to use **this file** along with the coverage report generated during `-test` github actions to be able to compute a diff and establish safeguards to avoid regression in our coverage... but that _realization_ is out of scope to this initial work.
    
You can/should be able to simulate this working end-to-end on your own `fork` of `kubeflow/notebooks` as you can commit this GitHub Action file to your own repo without review - and then play with it by creating additional PRs on your repo with arbitrary changes to `workspaces/backend` files to "see" the GHA behaving appropriately.

.github/workflows/go-coverage.yml
```diff
@@ -0,0 +1,47 @@
...
+    paths:
+      - 'workspaces/backend/**'
+    branches:
+      - notebook-v2
+
+permissions:
+  contents: write
+
+jobs:
+  test:
```
andyatmiami, line 14:
```suggestion
  backend-baseline:
```

mkoushni: done

.github/workflows/go-coverage.yml
```diff
@@ -0,0 +1,47 @@
+name: Go Baseline Coverage
+
+on:
+  push:
+    paths:
+      - 'workspaces/backend/**'
+    branches:
+      - notebook-v2
```
andyatmiami, line 8:
```suggestion
      - notebooks-v2
```

.github/workflows/go-coverage.yml
```diff
@@ -0,0 +1,47 @@
...
+    name: Run Tests and Generate Coverage
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        working-directory: workspaces/backend
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
```
andyatmiami, line 24:
```suggestion
        uses: actions/checkout@v4
```

andyatmiami: Looks like v4 is the latest:
- https://github.com/actions/checkout/releases

andyatmiami (approved):

andyatmiami:
discussed with community and it was decided we will not merge this PR...

there were significant concerns about the number of commits this approach would generate - as well as potential security issues with having the github bot auto-merge changes.

Issue #380, opened 2025-05-22
[TASK] Backend can read data served by envtest


andyatmiami:
## Description

k8s (and `KinD`) are extremely powerful platforms, but with that power comes added complexity/dependencies.  A developer willing to contribute front end code to the project may be discouraged by the added overhead of requiring a k8s cluster (including `KinD`), particularly if they have issues setting up and/or maintaining this dependency - particularly if its not working for any variety of reasons.

We believe [`envtest`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest) can play a critical role in not only helping in testing, but also in enabling a more streamlined experience for frontend developers, tackling the issue above.

With the separation provided by the `backend` (proxy) layer, we have a nice architecture to drive this solution.  We want to be able to structure the code, such that when a feature flag is set, calls from the `backend` to read/interact with (today) a k8s cluster could instead by serviced by `envtest`.

This design pattern has already been implemented by Kubeflow Model Registry.
- [k8s client factory](https://github.com/kubeflow/model-registry/blob/main/clients/ui/bff/internal/integrations/kubernetes/factory.go)
- [envtest client factory](https://github.com/kubeflow/model-registry/blob/main/clients/ui/bff/internal/integrations/kubernetes/k8mocks/mock_factory.go)

Please note the initial requirements for this work are **much** simpler as compared to the current Model Registry implementation.  Model Registry should only be used as a reference to help understand the basic structure of the code.
    - Specifically, this issue is **NOT** focused on adding the support of various authentication schemes!

At a high level, the desire is to have:
- a "client factory" that encapsulates the current logic of the `notebooks-v2` branch
- a "client factory" that allows reading data from `envtest`
- a feature flag that gives developers the ability to toggle which "mode" to run the application

## Acceptance Criteria
- As a frontend developer, I should be able to run the `frontend` and `backend` process of `kubeflow/notebooks` locally without any k8s cluster available, and still be able to have meaningful interactions in the UI.
- At least a single namespace, with a single workspace kind, with a single workspace should be able to be retrieved in the UI

PR #403, opened 2025-06-03
feat(ws): Notebooks 2.0 // Backend // Envtest creation 

yehudit1987:
This PR solve issue #380 

- envtest setup. 
- Factory for clients creations - with that mocked/real manager as well as rest.Config.


workspaces/backend/localdev/envsetup.go
```diff
@@ -0,0 +1,123 @@
...
+		} else {
+			cleanupLog.Info("Envtest control plane stopped successfully.")
+		}
+	} else {
+		cleanupLog.Info("testEnv was nil, nothing to stop.")
+	}
+	ctrl.Log.Info("Local dev environment stopped.")
+}
+
+func getProjectRoot() (string, error) {
```
andyatmiami, line 108: I am still myself upskilling my knowledge in Golang - so take what I say with a grain of salt!

I worry the reliance on `os.Getwd()` could lead to strange failure patterns...

From (quick) research - I feel something like the following may be _more robust_ - but open to discussion!

```
   import (
       "go/build"
       "golang.org/x/mod/modfile"
   )
   
   func getProjectRoot() (string, error) {
       // First try to find the module root
       if modRoot := os.Getenv("GOMOD"); modRoot != "" {
           return filepath.Dir(modRoot), nil
       }
       
       // Fall back to build.Import
       pkg, err := build.Import("github.com/kubeflow/notebooks", "", build.FindOnly)
       if err != nil {
           return "", err
       }
       return pkg.Dir, nil
   }
   ```
   
   example:
   ```
   ➜ backend/ git:(feat/#380) $ go env GOMOD
/Users/astonebe/Development/Code/GitHub/kubeflow-notebooks/workspaces/backend/go.mod
```

workspaces/backend/localdev/envtest_helper.go
```diff
@@ -0,0 +1,329 @@
...
+		if !apierrors.IsAlreadyExists(err) {
+			logger.Error(err, "Failed to create namespace", "name", namespaceName)
+			return fmt.Errorf("failed to create namespace %s: %w", namespaceName, err)
+		}
+		logger.Info("Namespace already exists", "name", namespaceName)
+	}
+	return nil
+}
+
+func loadAndCreateWorkspaceKindsFromDir(ctx context.Context, cl client.Client,
```
andyatmiami, line 59: **FOR DISCUSSION - NOT ADVOCATING FOR CHANGES QUITE YET**

I wonder if instead of relying on YAML files that exist in source control.. coupled then with all this application logic.. if it would just be easier to adopt an approach similar to what the `controller` is doing in its tests
- example: https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/controller/internal/controller/suite_test.go#L183


If we defined **everything** programmatically... I imagine the code to construct the workspaces would be easier to understand/implement as we can rely on `const` values (or at minimum variables defined purely within the application logic) and not need to parse out values from querying the control plan..

Thoughts?

workspaces/backend/internal/k8sclientfactory/client_factory.go
```diff
@@ -0,0 +1,124 @@
...
+		f.logger.Info("Using envtest mode: setting up local Kubernetes environment...")
+		var testEnvInstance *envtest.Environment
+
+		cfg, mgr, testEnvInstance, err = localdev.StartLocalDevEnvironment(ctx, f.crdPaths, f.scheme)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("could not start local dev environment: %w", err)
+		}
+		f.logger.Info("Local dev K8s API (envtest) is ready.", "host", cfg.Host)
+
+		if testEnvInstance != nil {
```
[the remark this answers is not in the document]
yehudit1987: For now you are correct that `testEnvInstance`  is not suppose to be `nil `at that point. 
In case there will be any changes in the future with  `StartLocalDevEnvironment ` it's more robust (or so I believe) to still verify that option. But again you are right that `else ` is needed here. Fix that, let me know what you think.

workspaces/backend/localdev/envsetup.go
yehudit1987: I see what you are saying. The use of `os.Getwd()` sets the dependency on the user which is in general less safe. 
I did some changes there (hope you'll find them acceptable), but still running that code via the make file is either way possible just from backend path. 

If there is other aspect, please share with me so I'll learn.

workspaces/backend/localdev/envtest_helper.go
yehudit1987: In general I think that separating the YAML's (the source) from the code (the actions) is much cleaner but I'm open for discussion of course. 
**YAML way:** 
pros:

- Separate responsibilities
- Easy to apply changes
- Can be re-used (not sure where but still). 

**Structs way:**
pros: 

- Types safety (although we will not push PR's with YAML's breaking the go). 
- No file system dependency (although it's not time consuming). 
- Simple data flow. 

Any other aspects you think about?

Issue #381, opened 2025-05-22
[TASK] Ensure test files exist in backend for any executed code


andyatmiami:
## Description

The `backend/` directory in `kubeflow/notebooks` `notebooks-v2` branch is currently pretty light in tests.  As we plan on ramping up development in this area, it is essential we have tests in place to ensure no regressions.  Additionally, to make it easier for contributors to add to the code, we need to remove  unnecessary impediments (like needing to create a net new `_test.go` file when only adding or changing functionality of an existing file).

Today, we see this:

```
$ gmake test
go fmt ./...
go vet ./...
KUBEBUILDER_ASSETS="/Users/astonebe/Development/Code/GitHub/kubeflow-notebooks/workspaces/backend/bin/k8s/1.31.0-darwin-arm64" go test ./... -coverprofile cover.out
?   	github.com/kubeflow/notebooks/workspaces/backend/internal/models/health_check	[no test files]
?   	github.com/kubeflow/notebooks/workspaces/backend/internal/config	[no test files]
	github.com/kubeflow/notebooks/workspaces/backend/internal/models/workspaces		coverage: 0.0% of statements
	github.com/kubeflow/notebooks/workspaces/backend/internal/models/workspacekinds		coverage: 0.0% of statements
	github.com/kubeflow/notebooks/workspaces/backend/cmd		coverage: 0.0% of statements
	github.com/kubeflow/notebooks/workspaces/backend/internal/helper		coverage: 0.0% of statements
	github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/namespaces		coverage: 0.0% of statements
	github.com/kubeflow/notebooks/workspaces/backend/internal/repositories		coverage: 0.0% of statements
	github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/health_check		coverage: 0.0% of statements
	github.com/kubeflow/notebooks/workspaces/backend/internal/models/namespaces		coverage: 0.0% of statements
	github.com/kubeflow/notebooks/workspaces/backend/internal/auth		coverage: 0.0% of statements
	github.com/kubeflow/notebooks/workspaces/backend/openapi		coverage: 0.0% of statements
	github.com/kubeflow/notebooks/workspaces/backend/internal/server		coverage: 0.0% of statements
	github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/workspaces		coverage: 0.0% of statements
[...]

PR #410, opened 2025-06-09
test(ws): Notebooks 2.0 // Backend // Add tests

yehudit1987:
This PR solves partially issue #381

### DIFF
--- a/workspaces/backend/internal/auth/authentication_test.go
+++ b/workspaces/backend/internal/auth/authentication_test.go
@@ -1,0 +1,106 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package auth
+
+import (
+	"net/http"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("NewRequestAuthenticator", func() {
+	const (
+		userHeader   = "X-User"
+		groupsHeader = "X-Groups"
+		userPrefix   = "service-account:"
+	)
+
+	type authTestInput struct {
+		userHeaderValue     string
+		groupsHeaderValue   string
+		userPrefix          string
+		expectAuthenticated bool
+		expectedUserName    string
+		expectedGroups      []string
+	}
+
+	runAuthTest := func(input authTestInput) {
+		authn, err := NewRequestAuthenticator(userHeader, input.userPrefix, groupsHeader)
+		Expect(err).NotTo(HaveOccurred())
+
+		req, _ := http.NewRequest("GET", "/", http.NoBody)
+		if input.userHeaderValue != "" {
+			req.Header.Set(userHeader, input.userHeaderValue)
+		}
+		if input.groupsHeaderValue != "" {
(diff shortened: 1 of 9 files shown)


workspaces/backend/internal/auth/authentication_test.go
```diff
@@ -0,0 +1,91 @@
...
+package auth
+
+import (
+	"net/http"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("NewRequestAuthenticator", func() {
```
andyatmiami, line 26: admittedly could be viewed as "overkill" here - but seems like we could write a generic helper function to _run the authenticate tests_

By simply providing `user` + `groups` + `prefx` as input to said generic function - it could do the invoking and also dynamically make the right assertions based on input being defined/empty

workspaces/backend/internal/helper/validation_test.go
```diff
@@ -0,0 +1,125 @@
...
+			errs := ValidateFieldIsNotEmpty(path, "")
+			Expect(errs).To(HaveLen(1))
+			Expect(errs[0].Type).To(Equal(field.ErrorTypeRequired))
+		})
+	})
+
+	Describe("ValidateFieldIsDNS1123Subdomain", func() {
+		path := field.NewPath("metadata", "name")
+
+		DescribeTable("should validate DNS1123 subdomain",
```
andyatmiami, line 90: nice use of `DescribeTable` 💯

workspaces/backend/internal/repositories/workspaces/suite_test.go
```diff
@@ -0,0 +1,218 @@
...
+	})
+
+	Describe("GetWorkspaces", func() {
+		var (
+			ns1     = "ns1"
+			ns2     = "ns2"
+			wsKind1 = &kubefloworgv1beta1.WorkspaceKind{ObjectMeta: metav1.ObjectMeta{Name: "kind1"}}
+			ws1     = &kubefloworgv1beta1.Workspace{ObjectMeta: metav1.ObjectMeta{Name: "ws1", Namespace: ns1}, Spec: kubefloworgv1beta1.WorkspaceSpec{Kind: "kind1"}}
+			ws2     = &kubefloworgv1beta1.Workspace{ObjectMeta: metav1.ObjectMeta{Name: "ws2", Namespace: ns1}, Spec: kubefloworgv1beta1.WorkspaceSpec{Kind: "kind-nonexistent"}}
+			ws3     = &kubefloworgv1beta1.Workspace{ObjectMeta: metav1.ObjectMeta{Name: "ws3", Namespace: ns2}, Spec: kubefloworgv1beta1.WorkspaceSpec{Kind: "kind1"}}
```
andyatmiami, line 109: While I wholly appreciate what this test setup is trying to do - I feel a little _uneasy_ about constructing this scenario in a way that is literally impossible to encounter in the "real world"...

i.e. if we are declaring a `WorkspaceSpec` - it feels weird/wrong to me to set the `Kind` to some value other than `Workspace`

Don't feel strongly about the change - but I wonder if we could simply include some non-`Workspace` objects in the `initObjs` parameter that then more closely simulates how this works in reality...

workspaces/backend/internal/repositories/workspaces/suite_test.go
```diff
@@ -0,0 +1,218 @@
...
+		})
+
+		It("returns empty list when namespace has no workspaces", func() {
+			r := newTestWorkspaceRepository(ws1, ws2)
+			wss, err := r.GetWorkspaces(ctx, "nonexistent-ns")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(wss).To(BeEmpty())
+		})
+
+		It("returns all workspaces in the cluster", func() {
```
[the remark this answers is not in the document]
andyatmiami: i was getting confused on `kind` vs. `spec.kind` - which @yehudit1987 kinda clearly up for me...

This comment can be ignored.

andyatmiami (approved):

PR #412, opened 2025-06-10
test(ws): Ensure test files exist in backend for any executed code #381

henschwartz:
this are tests for backend directory, for ticket 381 (this is partial PR for all backend dir)

 related: #381

### DIFF
--- a/workspaces/backend/internal/models/workspaces/funcs_test.go
+++ b/workspaces/backend/internal/models/workspaces/funcs_test.go
@@ -1,0 +1,574 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package workspaces_test
+
+import (
+	"time"
+
+	kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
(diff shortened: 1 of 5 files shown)


workspaces/backend/internal/models/workspacekinds/funcs_test.go
```diff
@@ -0,0 +1,160 @@
...
+				Expect(result.DeprecationMessage).To(Equal(""))
+				Expect(result.Hidden).To(BeFalse())
+				Expect(result.Icon.URL).To(Equal("/workspaces/backend/api/v1/workspacekinds/test-workspacekind/assets/icon"))
+				Expect(result.Logo.URL).To(Equal("/workspaces/backend/api/v1/workspacekinds/test-workspacekind/assets/logo"))
+				Expect(result.PodTemplate.PodMetadata.Labels).To(HaveKeyWithValue("app", "test-workspacekind"))
+				Expect(result.PodTemplate.PodMetadata.Annotations).To(HaveKeyWithValue("annotation-key", "annotation-value"))
+			})
+		})
+
+		Context("with empty strings in required fields", func() {
```
andyatmiami, line 87: All 3 of the tests being added share some similarities in that we are providing various inputs to `Spawner` (amongst other things) and asserting on expected output.

This is a great use case to start messing around with table-driven design patterns!
- https://go.dev/wiki/TableDrivenTests

Please note I am not asking to create **more** than what you have here (as we want this initial PR to simply "set the foundation") - but I think we could adapt these as a blueprint for others to follow on how Table-Driven tests should look.

WDYT?

workspaces/backend/internal/models/workspaces/funcs_test.go
```diff
@@ -0,0 +1,420 @@
+/*
```
andyatmiami, line 1: admittedly, there is a lot of "static information" that needs to be set up/provided when doing these workspace tests...

i appreciate your usage of `DescribeTable` to help make adding future tests easier.. but I still feel we have a lot of hard-coded "magic strings" here that I would love to avoid...

I'm not sure there is an obvious/elegant answer - but can you _do some thinkin'_ and try to come up with an implementation that has less hard-coding between model instantiation and test assertions?  I don't wanna over-engineer these tests - but hopefully some well-defined helper functions and/or variable declarations can make these tests more maintainable (?)

PR #414, opened 2025-06-11
chore(ws): enforce named imports for react hooks

paulovmr:
Currently, some typescript files use named imports and others use namespace imports to import React Hooks. This PR enforces through ESLint that React Hooks must be imported using named imports, as a way to organize and standardize the frontend code.

With this PR, if a React Hook is used with the `React.` prefix, there will be a lint error, for example: 
```
error  Import React hook "useMemo" directly instead of using React.useMemo          no-react-hook-namespace
```


ederign:
@paulovmr PR lgtm, but needs a rebase.

PR #424, opened 2025-06-16
feat(ws): fix swagger warnings and only generate json output

liavweiss:
This fix has no GitHub issue and will fix this:
While you put our swagger yaml into https://editor.swagger.io/, you will get a few warnings popping up.
image

This pr will not fix the "example" warning for this reason:
In `Swagger 2.0` `example` should be with `x-` `example`, the issue is in `swaggo/swag` recognize only `example()` in the documentation above each function, and it will generate `example: ` in the yaml file.

Bottom line:
It works fine in Swagger UI, but causes errors in Swagger Editor, and I dont think it's a blocker in our project.


andyatmiami (commented):
This is certainly an _open discussion_ on whether or not we'd want to do this (and/or how important it is to adhere to Swagger 2.0 spec) - but if we really want to resolve the warnings about the `example` attribute - it seems we can do it as such:

image

By leveraging the `extensions(...)` construct - this does render appropriately in the client:
image

image

liavweiss:
@andyatmiami Regarding the `x-example`, you are right (Im sorry I don't know how I missed that) - will be fixed.

Regarding the swagg init output: I’ll drop the YAML output - since the tooling (including the auto type generation) works fine with JSON and we’re not relying on YAML for anything specific.
We can add it in the future if necessary.

andyatmiami:
/lgtm

Verified now that the generated `swagger.json` file reports no errors when pasted into Swagger Editor

image

andyatmiami (approved):

PR #428, opened 2025-06-16
fix(ws): Updates to Table Columns, Expandable Rows, and Theming

### LINKED ISSUE #372: [TASK] Workspace List: Column Header Visibility
Column headers in the workbench creation process are not fully visible due to the number of columns. It is unclear what all columns represent.

Image

UX Recommendations:

- [x] Review and potentially reorganize properties displayed in the row and expanded section.
- [ ] Consider allowing users to control visible columns or implement horizontal scrolling.

jenny-s51:
Closes #372

## Overview
- Improved workspaces table column organization and expandable content.
- Fixed several visual and theming issues.

## Table Updates
- **Primary columns**: Now show Name, Image (with redirect status icon), Kind, State, and Last Activity.
- **Expandable rows**: Now display Home Volume, Pod Config, CPU, and Memory. Uses PatternFly Description List for best practices.

## Theming & Visual Fixes (MUI-theme.scss)
- **Buttons**: Added `--mui-button--BorderWidth: 1px` and improved split button border color, style, and consistency.
- **Table**: Fixed cell spacing and improved column width handling for better layout stability.

---

Before:
Screenshot 2025-06-16 at 5 14 59 PM

After:
Screenshot 2025-06-18 at 1 11 48 PM


jenny-s51 (commented):
> @jenny-s51 This looks pretty good, but I have a couple questions/thoughts.
> 
> 1. Can you make the top level row names wrap so users can read the full name? I.e. Last active

Yes, great idea 
> 2. I should probably know this, but what does Home volume refer to?

I am wondering the same - can you provide any insight here @paulovmr ?
> 3. Not sure if the Package info title is missing because we don't have package info, or you missed it, but it would be good to include the title no matter what the state is


Will update with title and package info:
Screenshot 2025-06-17 at 5 38 09 PM


> 4. I think the titles for CPU and Memory could be more descriptive. Are these limits, requests, or type (I know they aren't type, but it could be confusing to someone else ;))?

Nice catch, will rename

> 5. The titles are a mix of Sentence case and Title Case. For PF we recommend Sentence case (so that's my vote), but I'm not sure if Kubeflow has a different recommendation. Either way we should be consistent.

👍 Will update this as well

> 
> Overall I think all these changes could be done in a follow up. This is significantly easier to understand than the last version! Awesome job!

jenny-s51:
See #432 for subsequent updates.

Issue #416, opened 2025-06-11
[TASK] API that allows frontend to update a WorkspaceKind by uploading a YAML file containing a full WorkspaceKind definition


andyatmiami:
## Description

Previously, we have been relying on `kubectl` to update `WorkspaceKind` resources.  We want our Notebooks 2.0 users to be able to effectively manage their cluster all from within the UI.  As such, we need an API that the `frontend` can call which will update a `WorkspaceKind` resource.

**HOWEVER**, given the complexity and nesting present in the `WorkspaceKind`, we do not yet want to commit to building a full blown user experience (wizard flow or otherwise) as its not clear at this time what an "ideal experience" could/would be.  In the near term, then, to enable this functionality while not overcommitting on a design that could prove to be wasted effort, we want to have a lightweight API that simply accepts the "raw YAML manifest' for an existing WorkspaceKind that the backend can use as input to a `client.Update` call.
- ⚠ Get confirmation if we want to do `Update` vs. `Patch` vs. _SSA_ 
    
## Acceptance Criteria
- Endpoint registered in `backend` under `PUT api/v1/workspacekinds/{name}`
- Only accepts a `Content-Type` of `application/vnd.kubeflow-notebooks.manifest+yaml`
    - ⚠ Note the value may change based on community feedback - but the important thing to note is that `application/json` will **NOT** be a supported `Content-Type`
    - Unsupported `Content-Type` should result in a `415` error and "fail fast" (prior to attempting to parse the request body) 
- API logic expects and parses a YAML file that adheres to the `WorkspaceKind` CRD
- Immutable fields should not be able to be modified
    - ⚠ Get confirmation if we want to return error vs. overwrite (I think we should fail)
    - List of immutable fields ( ⚠ needs confirmation ):
        - `.metadata.name`
            - value must match what is in URL path
        - `.spec.podTemplate.serviceAccount.name`
        - `.spec.volumeMounts.home`
        - `options.imageConfig.values` + `options.podConfig.values`
            - elements cannot be removed (this includes renaming `.id`)
- `client.Update` is used to apply the payload to the k8s cluster
    - ⚠ Get confirmation if we want to do `Update` vs. `Patch` vs. _SSA_ 
- The **response** of the API request is `application/json` and follows the current `backend` schema for `GET /workspacekinds`

PR #429, opened 2025-06-17
feat(ws): Notebooks 2.0 // Backend // Add API for wsk update

yehuditkerido:
This PR solve issue #416

### DIFF
--- a/workspaces/backend/api/helpers.go
+++ b/workspaces/backend/api/helpers.go
@@ -19,9 +19,13 @@ package api
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"mime"
 	"net/http"
 	"strings"
+
+	kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"
+	"sigs.k8s.io/yaml"
 )
 
 // Envelope is the body of all requests and responses that contain data.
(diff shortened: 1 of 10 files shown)


workspaces/backend/api/helpers.go
```diff
@@ -94,3 +98,107 @@ func (a *App) LocationGetWorkspace(namespace, name string) string {
...
+	Field    string
+	BadValue interface{}
+	Detail   string
+}
+
+func (e *ImmutableFieldError) Error() string {
+	return fmt.Sprintf("Field %s is immutable: %v", e.Field, e.Detail)
+}
+
+func validateImmutableFields(newObj, oldObj *kubefloworgv1beta1.WorkspaceKind) *ImmutableFieldError {
```
andyatmiami, line 131: This function seems **very** tightly coupled to workspace kinds.. feels _out of place_ in the more generic `api/helper.go` file... as I view `helpers` to be more generic/applicable to any/all `api` files...

workspaces/backend/internal/models/workspacekinds/funcs.go
```diff
@@ -53,6 +53,7 @@ func NewWorkspaceKindModelFromWorkspaceKind(wsk *kubefloworgv1beta1.WorkspaceKin
 
 	return WorkspaceKind{
 		Name:               wsk.Name,
+		ResourceVersion:    wsk.ResourceVersion,
```
andyatmiami, line 56: I'm not sure we wanna expose this very k8s-centric mechanisms in our `backend` model...

things like `ResourceVersion` and `UID` seem to be concepts that the `backend` should "shield"  from the `frontend`

workspaces/backend/internal/models/workspacekinds/types.go
```diff
@@ -18,6 +18,7 @@ package workspacekinds
 
 type WorkspaceKind struct {
 	Name               string      `json:"name"`
+	ResourceVersion    string      `json:"resourceVersion"`
```
andyatmiami, line 21: see: https://github.com/kubeflow/notebooks/pull/429/files#r2155284059

workspaces/backend/internal/repositories/workspacekinds/repo.go
```diff
@@ -68,3 +68,19 @@ func (r *WorkspaceKindRepository) GetWorkspaceKinds(ctx context.Context) ([]mode
 
 	return workspaceKindsModels, nil
 }
+
+func (r *WorkspaceKindRepository) GetWorkspaceKindCR(ctx context.Context, name string) (*kubefloworgv1beta1.WorkspaceKind, error) {
```
andyatmiami, line 72: Feels like we could/should also leverage this function here to avoid some code duplication:
- https://github.com/kubeflow/notebooks/pull/429/files#diff-9fb242426dc632e71cc75286bce68c506a3dd1dda17426cd46e26410b618033eR42

Admittedly handling of propagating the `apierrors.IsNotFound` case requires a little extra consideration...

andyatmiami:
Discussed amongst the community and realized there was some miscommunication that led to me defining incorrect acceptance criteria for this work.

We will **NOT** (ever) expose an update (`PUT`) API that accepts the "raw" YAML manifest as input.  Any "update API" will expect a JSON data structure aligned to a `backend` model.

@thesuperzapper is working on a comprehensive/best-practices example of how we will handle `Workspace` update - which we can then use as a baseline for how we will want to structure `WorkspaceKind` update.

PR #432, opened 2025-06-18
fix(ws): updates to table columns

### LINKED ISSUE #372: [TASK] Workspace List: Column Header Visibility

jenny-s51:
Closes #372

## Overview
- Improved workspaces table column organization and expandable content.
- Fixed several visual and theming issues.

## Table Updates
- **Primary columns**: Now show Name, Image (with redirect status icon), Kind, State, and Last Activity.
- **Expandable rows**: Now display Home Volume, Pod Config, CPU, and Memory. Uses PatternFly Description List for best practices.

## Theming & Visual Fixes (MUI-theme.scss)
- **Buttons**: Added `--mui-button--BorderWidth: 1px` and improved split button border color, style, and consistency.
- **Table**: Fixed cell spacing and improved column width handling for better layout stability.

---

Before:
Screenshot 2025-06-16 at 5 14 59 PM

After:
Screenshot 2025-06-18 at 1 11 48 PM


workspaces/frontend/src/app/types.ts
```diff
@@ -7,6 +7,20 @@ import {
   Workspace,
 } from '~/shared/api/backendApiTypes';
 
+export interface WorkspacesColumnNames {
```
caponetto, line 10: This interface is not being used anymore. Could you please remove it?

workspaces/frontend/src/app/pages/Workspaces/WorkspacePackageDetails.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface WorkspacePackageDetailsProps {
+  workspace: Workspace;
+}
+
+export const WorkspacePackageDetails: React.FC<WorkspacePackageDetailsProps> = ({ workspace }) => {
+  const imageConfig = workspace.podTemplate.options.imageConfig.current;
+  const jupyterLabVersion = imageConfig.labels.find(
+    (label) => label.key === 'jupyterlabVersion',
+  )?.value;
+  const pythonVersion = imageConfig.labels.find((label) => label.key === 'pythonVersion')?.value;
```
caponetto, line 21: I think we can define a map of label keys and their corresponding text that we want to extract, and then create a generic code to extract and render them down there.

Something like:

```ts

const PACKAGE_LABELS: Record<string, string> = {
  jupyterlabVersion: 'JupyterLab',
  pythonVersion: 'Python',
  // Add more mappings here as needed later
};

...

const { labels } = workspace.podTemplate.options.imageConfig.current;

const renderedItems = Object.entries(PACKAGE_LABELS).flatMap(([key, label]) => {
  const value = labels.find((l) => l.key === key)?.value;
  return value ? <ListItem key={key}>{`${label} v${value}`}</ListItem> : [];
});

...

<List isPlain>{renderedItems}</List>
```

We'd probably also want to show something else if no matching labels are found, i.e., `renderedItems.length === 0`

workspaces/frontend/src/app/components/WorkspaceTable.tsx
```diff
@@ -38,7 +40,6 @@ import {
   FilterableDataFieldKey,
   SortableDataFieldKey,
 } from '~/app/filterableDataHelper';
-import { ExpandedWorkspaceRow } from '~/app/pages/Workspaces/ExpandedWorkspaceRow';
```
caponetto, line 41: As `ExpandedWorkspaceRow` is not being used anymore, can we delete its file?

workspaces/frontend/src/app/components/WorkspaceTable.tsx
```diff
@@ -52,26 +53,25 @@ import {
...
   filterableKeyArray: filterableWsTableColumnKeyArray,
 } = defineDataFields({
-  redirectStatus: { label: 'Redirect Status', isFilterable: false, isSortable: false },
   name: { label: 'Name', isFilterable: true, isSortable: true },
   kind: { label: 'Kind', isFilterable: true, isSortable: true },
   namespace: { label: 'Namespace', isFilterable: true, isSortable: true },
   image: { label: 'Image', isFilterable: true, isSortable: true },
-  podConfig: { label: 'Pod Config', isFilterable: true, isSortable: true },
   state: { label: 'State', isFilterable: true, isSortable: true },
   homeVol: { label: 'Home Vol', isFilterable: true, isSortable: true },
```
caponetto, line 71: I think we should remove `homeVol` from here as well, right?

workspaces/frontend/src/app/components/WorkspaceTable.tsx
```diff
@@ -408,150 +502,121 @@ const WorkspaceTable = React.forwardRef<WorkspaceTableRef, WorkspaceTableProps>(
...
+                    <Td>
+                      <ExpandableRowContent>
+                        <WorkspaceConfigDetails workspace={workspace} />
+                      </ExpandableRowContent>
+                    </Td>
+                    <Td />
+                    <Td />
+                    <Td />
+                    <Td />
+                  </Tr>
```
caponetto, line 619: Can we extract this piece of code to a dedicated component like it was being done with `ExpandedWorkspaceRow`?
Also, we are assuming here 8 `<Td>` but we should account for the scenario where we could have less columns (dictated by `visibleColumnKeys`) due to the `hiddenColumns` prop. All this logic could be calculated inside its own component. Saying this because the layout would break if we keep this code and show less columns.

caponetto (commented):
Thanks for the PR @jenny-s51!

workspaces/frontend/src/app/types.ts
jenny-s51: Yes, nice catch here @caponetto

workspaces/frontend/src/app/pages/Workspaces/WorkspacePackageDetails.tsx
jenny-s51: Thank you @caponetto, great idea - this is much cleaner and more maintainable!

paulovmr (changes requested):
Hi @jenny-s51 , thanks for the changes, it looks much better!

One thing I believe we could consider is making the width used by the `Home volume` and `Cluster storage` information a little bigger. Perhaps we could force the "lock" and "copy" icons to stay at the same line as the volume name and mount path even when the screen is smaller, wdyt?

workspaces/frontend/src/app/components/WorkspaceTable.tsx
```diff
@@ -371,19 +369,109 @@ const WorkspaceTable = React.forwardRef<WorkspaceTableRef, WorkspaceTableProps>(
...
+          ouiaId="SortableTable"
+          style={{ tableLayout: 'fixed' }}
+        >
           <Thead>
             <Tr>
-              {canExpandRows && <Th screenReaderText="expand-action" />}
-              {visibleColumnKeys.map((columnKey) => (
+              {canExpandRows && <Th width={10} screenReaderText="expand-action" />}
+              {visibleColumnKeys.includes('name') && (
+                <Th
```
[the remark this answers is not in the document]
caponetto: Exactly! And also replace all the repeaded `<Th ...` definitions with
```ts
              {visibleColumnKeys.map((columnKey) => (
                <Th
                  width={wsTableColumns[columnKey].width}
                  key={`workspace-table-column-${columnKey}`}
                  sort={getSortParams(columnKey)}
                  aria-label={columnKey}
                  modifier="wrap"
                  screenReaderText={columnKey}
                >
                  {wsTableColumns[columnKey].label}
                </Th>
              ))}
```
WDYT?

jenny-s51 (commented):
Thank you for your review @paulovmr , and nice catch - I've updated the layout and column width to prevent the icons from wrapping. Glad to handle any more changes in a follow-up PR so we can handle it outside of this scope - let me know if this is what you had in mind.

paulovmr (changes requested):
@jenny-s51 Just one thing, I think we are only missing the change about the repeated `<Th ...` that @caponetto commented to be able to merge this PR.

/remove-lable lgtm

jenny-s51 (commented):
Thank you @paulovmr , updated.

paulovmr (changes requested):
.

PR #436, opened 2025-06-24
feat(ws): validate podMetadata for ws and wsk in webhook

thesuperzapper:
This PR updates the controller's validating webhook to validate the following fields:

- Workspace:
     - `spec.podTemplate.podMetadata.labels`
     - `spec.podTemplate.podMetadata.annotations`
- WorkspaceKind:
     - `spec.podTemplate.podMetadata.labels`
     - `spec.podTemplate.podMetadata.annotations`

We do this by calling the upstream [`apimachinery`](https://pkg.go.dev/k8s.io/apimachinery) implementation so we validate exactly the same as upstream:
- Labels: [`v1validation.ValidateLabels()`](https://github.com/kubernetes/apimachinery/blob/v0.33.2/pkg/apis/meta/v1/validation/validation.go#L62-L72)
- Annotations: [`apivalidation.ValidateAnnotations()`](https://github.com/kubernetes/apimachinery/blob/v0.33.2/pkg/api/validation/objectmeta.go#L44-L56)

This is important because it's quite complex to correctly validate a [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) or [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set) metadata map, because some seemingly invalid label keys are valid, e.g. `example.com/my-label-key`.

### DIFF
--- a/workspaces/controller/internal/webhook/workspace_webhook.go
+++ b/workspaces/controller/internal/webhook/workspace_webhook.go
@@ -20,7 +20,10 @@ import (
 	"context"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apivalidation "k8s.io/apimachinery/pkg/api/validation"
+	v1validation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -76,12 +79,12 @@ func (v *WorkspaceValidator) ValidateCreate(ctx context.Context, obj runtime.Obj
 	}
 
 	// validate the Workspace
-	if err := v.validateImageConfig(workspace, workspaceKind); err != nil {
-		allErrs = append(allErrs, err)
-	}
-	if err := v.validatePodConfig(workspace, workspaceKind); err != nil {
-		allErrs = append(allErrs, err)
-	}
+	// NOTE: we do this after fetching the WorkspaceKind as there will be multiple types in the future,
+	//       and we need to know which one the Workspace is using to validate it correctly.
+	allErrs = append(allErrs, v.validatePodTemplatePodMetadata(workspace)...)
+	allErrs = append(allErrs, v.validateImageConfig(workspace, workspaceKind)...)
+	allErrs = append(allErrs, v.validatePodConfig(workspace, workspaceKind)...)
+
 	if len(allErrs) == 0 {
 		return nil, nil
 	}
@@ -113,10 +116,14 @@ func (v *WorkspaceValidator) ValidateUpdate(ctx context.Context, oldObj, newObj 
 
 	// check if workspace kind related fields have changed
 	var workspaceKindChange = false
+	var podMetadataChange = false
 	var imageConfigChange = false
 	var podConfigChange = false
 	if newWorkspace.Spec.Kind != oldWorkspace.Spec.Kind {
 		workspaceKindChange = true
+	}
+	if !equality.Semantic.DeepEqual(newWorkspace.Spec.PodTemplate.PodMetadata, oldWorkspace.Spec.PodTemplate.PodMetadata) {
+		podMetadataChange = true
 	}
 	if newWorkspace.Spec.PodTemplate.Options.ImageConfig != oldWorkspace.Spec.PodTemplate.Options.ImageConfig {
 		imageConfigChange = true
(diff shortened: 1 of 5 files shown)


workspaces/controller/internal/webhook/workspace_webhook.go
```diff
@@ -197,36 +205,79 @@ func (v *WorkspaceValidator) validateWorkspaceKind(ctx context.Context, workspac
...
+	labelsPath := podMetadataPath.Child("labels")
+	errs = append(errs, v1validation.ValidateLabels(labels, labelsPath)...)
+
+	// validate annotations
+	annotations := podMetadata.Annotations
+	annotationsPath := podMetadataPath.Child("annotations")
+	errs = append(errs, apivalidation.ValidateAnnotations(annotations, annotationsPath)...)
+
+	return errs
+}
```
andyatmiami, line 231: The body of this function is identical its `WorkspaceKind` counterpart:
- https://github.com/kubeflow/notebooks/pull/436/files#diff-8a79276e0ce09b79596b68df827ed583a3e2f80b55fa6c23db648a6914f57c4eR496-R516

Should we abstract this into a common function to avoid duplication? Or do we anticipate this to potentially diverge in the future - such as having these implementations be mutually exclusive (albeit duplicitous now) is a _feature_ ?

workspaces/controller/internal/webhook/workspace_webhook.go
```diff
@@ -197,36 +205,79 @@ func (v *WorkspaceValidator) validateWorkspaceKind(ctx context.Context, workspac
...
+
+// validateImageConfig validates the imageConfig selected by a Workspace
+func (v *WorkspaceValidator) validateImageConfig(workspace *kubefloworgv1beta1.Workspace, workspaceKind *kubefloworgv1beta1.WorkspaceKind) []*field.Error {
+	var errs []*field.Error
```
andyatmiami, line 235: Do we instantiate this as a slice to simply make it "cleaner" for calling code to use the spread (...) operator?

Because If I am reading this correctly - this slice would only ever have a length of 0 or 1, right?

workspaces/controller/internal/webhook/workspacekind_webhook.go
```diff
@@ -140,8 +146,13 @@ func (v *WorkspaceKindValidator) ValidateUpdate(ctx context.Context, oldObj, new
...
 	getImageConfigUsageCount, getPodConfigUsageCount := v.getLazyOptionUsageCountFuncs(ctx, oldWorkspaceKind)
 
+	// validate the pod metadata
+	if !equality.Semantic.DeepEqual(newWorkspaceKind.Spec.PodTemplate.PodMetadata, oldWorkspaceKind.Spec.PodTemplate.PodMetadata) {
+		allErrs = append(allErrs, v.validatePodTemplatePodMetadata(newWorkspaceKind)...)
+	}
+
 	// validate the extra environment variables
-	if !reflect.DeepEqual(newWorkspaceKind.Spec.PodTemplate.ExtraEnv, oldWorkspaceKind.Spec.PodTemplate.ExtraEnv) {
+	if !equality.Semantic.DeepEqual(newWorkspaceKind.Spec.PodTemplate.ExtraEnv, oldWorkspaceKind.Spec.PodTemplate.ExtraEnv) {
```
andyatmiami, line 155: might be nice to call out in the PR description for greater awareness that we are changing from `reflect.DeepEqual` to `equality.Semantic` as its the more idiomatic Kubernetes approach...

as I would image we'd want to ensure others would _follow suit_ in the future.

workspaces/controller/internal/webhook/workspace_webhook.go
```diff
@@ -197,36 +205,79 @@ func (v *WorkspaceValidator) validateWorkspaceKind(ctx context.Context, workspac
...
+	}
+
+	return errs
 }
 
-// validatePodConfig checks if the podConfig selected by a Workspace exists a WorkspaceKind
-func (v *WorkspaceValidator) validatePodConfig(workspace *kubefloworgv1beta1.Workspace, workspaceKind *kubefloworgv1beta1.WorkspaceKind) *field.Error {
+// validatePodConfig validates the podConfig selected by a Workspace
+func (v *WorkspaceValidator) validatePodConfig(workspace *kubefloworgv1beta1.Workspace, workspaceKind *kubefloworgv1beta1.WorkspaceKind) []*field.Error {
+	var errs []*field.Error
```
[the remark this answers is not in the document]
thesuperzapper: This is correct, I chose to swap to a slice of errors so that we could:

1. be consistent with the other validation apis (which can sometimes return more than 1)
2. make the unpacking of the returns of all validation functions have similar signatures
3. account for the possibility that we want to apply more than one validation to this section

workspaces/controller/internal/webhook/workspacekind_webhook.go
thesuperzapper: Yep, this was identified last year, and even has a stale PR to apply it: https://github.com/kubeflow/notebooks/issues/69

The problem is that we need to be very careful not to assume the semantic equality check is actually what we are wanting to check (in the more complex parts of the codebase that use `reflect.DeepEqual`), which is why the existing PR was not merged yet: https://github.com/kubeflow/notebooks/pull/152

workspaces/controller/internal/webhook/workspace_webhook.go
thesuperzapper: See https://github.com/kubeflow/notebooks/pull/436#discussion_r2169708293

andyatmiami (approved):

thesuperzapper:
Thanks @andyatmiami!

I will approve for now, please raise any further comments in new issues/PRs.

/approve

Issue #477, opened 2025-07-10
[TASK] Refactor frontend client when Creating a Workspace Kind


thaorell:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

Right now, the request from `frontend` to create a new WorkspaceKind has `Content-Type:application/vnd.kubeflow-notebooks.manifest+yaml`, which should be `application/yaml`
In addition, `restYAML` utility function is redundant and duplicates `callRestJSON`

### Acceptance Criteria

- Request to create new Workspace Kind from `frontend` should have Content-Type: application/yaml in the header
- restFILE should be used when calling api.createWorkspaceKind

PR #478, opened 2025-07-10
feat(ws): Refactor restYAML to restFILE

thaorell:
closes: #477 
DESCRIPTION: 

- Remove restYAML utility function and use restFILE with directYAML flag when handling WorkspaceKind "POST" call
- Update header `Content-Type` to `application/yaml`


paulovmr (approved):
/lgtm

Issue #46, opened 2024-09-15
Notebooks 2.0 // Backend // Auto generate Open API spec


ederign:
Auto generates Open API spec for endpoints.

gaurangkudale:
Hi @ederign 
I want to work on this issue, Can you please assign it to me if it's still open?

gaurangkudale:
Hi @ederign 
I created this draft PR: https://github.com/kubeflow/notebooks/pull/175
Please take a look at it and let me know if any changes are needed

ederign:
This issue is still planned for Notebooks 2.0

PR #488, opened 2025-07-17
feat(ws): add @ID swag annotation for better codegen readability

andyatmiami:
related: #46 

- added @ID annotations for all API routes to populate operationId Swagger attribute
- split GetWorkspacesHandler into 2 separate handlers to account for @ID needing to be unique-per-route
    - GetAllWorkspacesHandler now services GET /workspaces
    - GetWorkspacesByNamespaceHandler now services GET /workspaces/{namespace}
    - non-exported getWorkspacesHandler function contains all business logic that existed in GetWorkspacesHandler
- Adjusted test cases to align with the new handler names.


thesuperzapper:
Thanks @andyatmiami!

/approve

Issue #430, opened 2025-06-17
[TASK] Add `required` fields to Swagger definitions for accurate frontend type auto-generation


caponetto:
## Description

Currently, the autogen Swagger files do not specify `required` fields in the object schemas under `definitions`. This omission causes all properties in the auto-generated TypeScript types to be marked as optional, regardless of whether they are actually required by the backend.

This impacts frontend form validation and strongly typed API integrations.

## Acceptance Criteria

- All object definitions in `swagger.{yaml|json}` explicitly include a `required` array when applicable.
- For each object, the `required` array lists all fields that are required by the backend.

PR #489, opened 2025-07-18
chore(ws): update swaggo dependency to v1.16.5 and add required fields in OpenAPI definitions

andyatmiami:
related: #430 

- Updated swaggo/swag from v1.16.4 to v1.16.5 in go.mod and Makefile.
- Added required fields to various OpenAPI definitions in docs.go and swagger.json for better validation via `--requiredByDefault` flag
    - `swag` `v1.16.5` contains a commit we authored to treat `json:omitempty` as `required: false`.  That, in conjunction with `--requiredByDefault` flag, allows us to generate models with proper _required-ness_


thesuperzapper:
@andyatmiami thanks for this, should make it easier for the frontend to generate a client.

/lgtm
/approve

PR #500, opened 2025-07-24
feat: Setup virtualservice to route traffic to relevant workspace

harshad16:
related: https://github.com/kubeflow/notebooks/issues/37
follow-up: https://github.com/kubeflow/notebooks/pull/42

Setup:
TBD


workspaces/backend/internal/models/workspaces/funcs.go
```diff
@@ -350,7 +350,7 @@ func buildServices(ws *kubefloworgv1beta1.Workspace, wskPodTemplatePorts []kubef
 		case kubefloworgv1beta1.ImagePortProtocolHTTP:
 			services[i].HttpService = &HttpService{
 				DisplayName: ptr.Deref(port.DisplayName, wskPodTemplatePorts[i].DefaultDisplayName),
-				HttpPath:    fmt.Sprintf("/workspace/%s/%s/%s/", ws.Namespace, ws.Name, port.Id),
+				HttpPath:    fmt.Sprintf("/workspace/connect/%s/%s/%s/", ws.Namespace, ws.Name, port.Id),
```
thesuperzapper, line 353: Note, we should update all other instances which look like `/workspace/...` path string formatters (do a search over the whole project)

In addition, we should update the doc string of the WorkspaceKind types (in controller) and the doc srtring of the WorkspaceKind same (which should usually match the go doc string anyway).

harshad16: Made the effort to update these changes

workspaces/controller/config/manager/manager.yaml
```diff
@@ -54,6 +54,22 @@ spec:
...
+          - name: ISTIO_GATEWAY
+            valueFrom:
+              configMapKeyRef:
+                name: config
+                key: ISTIO_GATEWAY
+          - name: ISTIO_HOST
+            valueFrom:
+              configMapKeyRef:
+                name: config
+                key: ISTIO_HOST
```
thesuperzapper: These variables should be added via the kustomize patch on this deployment.

workspaces/controller/config/components/istio/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
...
+configMapGenerator:
+- envs:
+  - params.env
+  name: kubeflow-workspaces-istio-config
+generatorOptions:
+  disableNameSuffixHash: true
+
+labels:
+- pairs:
+    app.kubernetes.io/component: controller-manager
```
thesuperzapper: This should not be necessary, because `default/kustomization.yaml` already references `../manager` which includes the correct labels directive.

workspaces/controller/internal/controller/suite_test.go
```diff
@@ -106,13 +112,16 @@ var _ = BeforeSuite(func() {
...
+	// Use test config with UseIstio disabled to avoid VirtualService CRD requirements in tests
+	testCfg := &config.EnvConfig{UseIstio: false}
+	err = helper.SetupManagerFieldIndexers(k8sManager, testCfg)
 	Expect(err).NotTo(HaveOccurred())
 
 	By("setting up the Workspace controller")
 	err = (&WorkspaceReconciler{
 		Client: k8sManager.GetClient(),
 		Scheme: k8sManager.GetScheme(),
+		Config: testCfg,
```
thesuperzapper: Small nit for future work to enable istio tests in envtest:

```go
	envConfig := &config.EnvConfig{
		// TODO: make true once we install Istio CRDs in EnvTest.
		//       also create unit tests to ensure VirtualService is created by controller.
		UseIstio: false,
	}

	By("setting up the field indexers for the controller manager")
	err = helper.SetupManagerFieldIndexers(k8sManager, envConfig)
	Expect(err).NotTo(HaveOccurred())

	By("setting up the Workspace controller")
	err = (&WorkspaceReconciler{
		Client: k8sManager.GetClient(),
		Scheme: k8sManager.GetScheme(),
		Config: envConfig,
	}).SetupWithManager(k8sManager, controller.Options{
		RateLimiter: helper.BuildRateLimiter(),
	})
	Expect(err).NotTo(HaveOccurred())
```

workspaces/controller/config/overlays/istio/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
thesuperzapper: Let's just update the `default/kustomization.yaml` to include istio component rather than making an overlay.

In the future we will refactor the controller manifests to be similar to the backend and frontend (using base + components+overlays) but this is a larger work to integrate with crd generate for example.

workspaces/controller/internal/controller/suite_test.go
```diff
@@ -42,6 +42,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"
+
+	istiov1 "istio.io/client-go/pkg/apis/networking/v1"
+
```
thesuperzapper: This import format is strange, let's move the `istiov1 "istio.io/client-go/pkg/apis/networking/v1"` up to the "external" imports section below `"sigs.k8s.io/controller-runtime/pkg/log/zap"`.

Also, `kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"` is an internal import so should be at the bottom with the other internal imports.

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -356,10 +365,71 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-	//
+	if r.Config.UseIstio {
+		log.V(2).Info("reconciling VirtualService for Workspace")
+		// generateVirtualService
+		virtualsvc, err := generateVirtualService(workspace, workspaceKind, serviceName, currentImageConfig.Spec, r.Config)
+		if err != nil {
+			log.V(0).Info("failed to generate VirtualService for Workspace", "error", err.Error())
+			return r.updateWorkspaceState(ctx, log, workspace,
+				kubefloworgv1beta1.WorkspaceStateError,
+				fmt.Sprintf("failed to generate VirtualService for Workspace: %s", err.Error()),
```
thesuperzapper: We should create a variable like we do for all other WS state messages, `stateMsgErrorGenFailureVirtualService` (note, the log message is fine to hard code).

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -913,6 +996,105 @@ func generateService(workspace *kubefloworgv1beta1.Workspace, imageConfigSpec ku
...
+	}
+
+	return httpRoute
+}
+
+// generateVirtualService generates a VirtualService for a Workspace
+func generateVirtualService(workspace *kubefloworgv1beta1.Workspace, workspaceKind *kubefloworgv1beta1.WorkspaceKind, serviceName string, imageConfigSpec kubefloworgv1beta1.ImageConfigSpec, cfg *config.EnvConfig) (*istiov1.VirtualService, error) {
+	if cfg.IstioGateway == "" {
+		return nil, fmt.Errorf("ISTIO_GATEWAY environment variable is not set")
+	}
```
thesuperzapper: Let's not do this, and just trust the user has set it correctly to avoid runtime errors that are hard to debug.

In the future we could consider validating in `main.go` but probably not worth it now.

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -913,6 +996,105 @@ func generateService(workspace *kubefloworgv1beta1.Workspace, imageConfigSpec ku
...
+						Number: uint32(port.Port), //nolint:gosec
+					},
+				},
+			},
+		},
+	}
+
+	if !ptr.Deref(podTemplatePort.HTTPProxy.RemovePathPrefix, false) {
+		httpRoute.Rewrite.Uri = matchUriPrefix
+	}
```
thesuperzapper: Let's reverse the order of this function, same as all other `generateXXX` functions.

That is, let's construct the "pieces" (e.g. in this case the `Rewrite` object) at the top of the function, explicitly setting empty where appropriate. 

Then define a full `&networkingv1.HTTPRoute{}` object just before we return, so it's easy to see what fields are defined.

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -913,6 +996,105 @@ func generateService(workspace *kubefloworgv1beta1.Workspace, imageConfigSpec ku
...
+	currentPodTemplatePortsMap := make(map[kubefloworgv1beta1.PortId]kubefloworgv1beta1.WorkspaceKindPort)
+	for _, port := range workspaceKind.Spec.PodTemplate.Ports {
+		currentPodTemplatePortsMap[port.Id] = port
+	}
+
+	serviceHost := fmt.Sprintf("%s.%s.svc.%s", serviceName, workspace.Namespace, cfg.ClusterDomain)
+
+	httpRoutes := []*networkingv1.HTTPRoute{}
+	for _, port := range imageConfigSpec.Ports {
+		// Guard: skip if port doesn't exist in current pod template
```
thesuperzapper: To make it clear why we don't error, we should say:

```suggestion
		// silently ignore port ids not defined in the workspace kind
		// NOTE: this should not be possible as the webhook blocks undefined ports
```

workspaces/controller/test/e2e/e2e_test.go
```diff
@@ -76,12 +76,20 @@ var _ = Describe("controller", Ordered, func() {
 		cmd = exec.Command("kubectl", "create", "ns", workspaceNamespace)
 		_, _ = utils.Run(cmd) // ignore errors because namespace may already exist
 
+		// TODO: enable Istio injection once we have logic to create VirtualServices during Workspace reconciliation
```
thesuperzapper: This is correct (we should be labeling the namespaces) but lets remove this TODO.

workspaces/controller/config/components/istio/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+
+configMapGenerator:
+- envs:
+  - params.env
+  name: kubeflow-workspaces-istio-config
+generatorOptions:
+  disableNameSuffixHash: true
```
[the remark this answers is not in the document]
harshad16: Removed the hash, kustomize does resolve it automatically 
fixed it in commit: https://github.com/kubeflow/notebooks/pull/500/commits/9f071ad2786b8b2789f76391f696ebb46bebc00f

workspaces/controller/internal/controller/suite_test.go
harshad16: Fixed it with: https://github.com/kubeflow/notebooks/pull/500/commits/daa43c9c4bb5c9dccd158e4bf42dfcaf5f6903ff

workspaces/controller/internal/controller/workspace_controller.go
harshad16: made the new changes, with : https://github.com/kubeflow/notebooks/pull/500/commits/daa43c9c4bb5c9dccd158e4bf42dfcaf5f6903ff

workspaces/controller/internal/webhook/suite_test.go
```diff
@@ -45,7 +45,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"
+	"github.com/kubeflow/notebooks/workspaces/controller/internal/config"
 	"github.com/kubeflow/notebooks/workspaces/controller/internal/helper"
+
+	istiov1 "istio.io/client-go/pkg/apis/networking/v1"
```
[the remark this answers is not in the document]
harshad16: Fixed it with: https://github.com/kubeflow/notebooks/pull/500/commits/daa43c9c4bb5c9dccd158e4bf42dfcaf5f6903ff

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -913,6 +990,112 @@ func generateService(workspace *kubefloworgv1beta1.Workspace, imageConfigSpec ku
 	return service, nil
 }
 
+// generateVirtualServiceHTTPRoute creates an HTTPRoute for a given port configuration
+func generateVirtualServiceHTTPRoute(
+	workspace *kubefloworgv1beta1.Workspace,
+	service *corev1.Service,
+	imageConfigPort kubefloworgv1beta1.ImagePort,
+	podTemplatePort kubefloworgv1beta1.WorkspaceKindPort,
+	cfg *config.EnvConfig,
```
thesuperzapper: Lets make this a class method so we dont need to pass `cfg`.

workspaces/controller/test/e2e/e2e_test.go
```diff
@@ -145,23 +166,32 @@ var _ = Describe("controller", Ordered, func() {
 		)
 		_, _ = utils.Run(cmd)
 
-		By("deleting sample WorkspaceKind")
-		cmd = exec.Command("kubectl", "delete",
-			"-f", filepath.Join(projectDir, "config/samples/jupyterlab_v1beta1_workspacekind.yaml"),
+		By("waiting for workspace pod to be deleted")
+		cmd = exec.Command("kubectl", "wait", "pods",
```
thesuperzapper: Rather than waiting for the pod to delete, lets just do a blocking delete with a timeout on the WS delete step above.

workspaces/controller/internal/controller/workspace_controller.go
harshad16: thanks for the suggestion, 
made the chaneg with : https://github.com/kubeflow/notebooks/pull/500/commits/01955dd152137ed216ea1441a31b18a671fb07d3

PR #421, opened 2025-06-13
chore: enforce planning label rules for issues

andyatmiami:
related: https://github.com/kubeflow/notebooks/issues/327

This new GitHub Actions workflow listens for the 'opened' and 'labeled' events on GitHub issues and ensures any issue opened (or labelled) with a `kind/plan-xxx` label meets the following criteria:
- only users listed in the `AUTHORIZED_USER` JSON array can label an issue with the `kind/plan-xxx` label
- at most one of the `kind/plan-xxx` labels can exist on a given issue

The set of supported/expected `kind/plan-xxx` labels are:
- `kind/plan-epic`
- `kind/plan-feature`
- `kind/plan-task`

If an issue (with a `kind/plan-xxx` label) is **opened** by a non-authorized user - the issue will be automatically closed when the GitHub action fires.

If an issue is labelled with the `kind/plan-xxx` label by an unauthorized user - the `kind/plan-xxx` label is removed.

In both cases, a GitHub comment is added to the issue explaining why the issue was updated by the bot.

Please note, when opening a GH issue - **BOTH** the `opened` and `labeled` events will fire (if the issue has a label on it).  As such, the GitHub action in this PR was deliberately structured to account for multiple instances running in parallel on the same underlying GitHub issue.

Also, this work complements https://github.com/kubeflow/notebooks/pull/361 - which defines GH issue templates that use the labels in a manner appropriate to satisfy the checks of this action.

PR #515, opened 2025-07-31
Update validate-planning-label.yml

jenny-s51:
Quick fix to provide @jenny-s51 with planning label permissions.

Follow up to #421


thesuperzapper:
@jenny-s51 @andyatmiami thanks for fixing this, will merge.

/lgtm
/approve

Issue #511, opened 2025-07-30
[TASK] Configure Frontend Application to Run from /workspaces Subpath


thaorell:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

To host the application within a larger platform, the entire frontend must be configured to run from the /workspaces/ URL prefix. This involves updating the build configuration to correctly bundle static assets (JS/CSS) and configuring the client-side router to handle in-app navigation relative to this new base path.

### Acceptance Criteria

- The Webpack publicPath (or homepage in package.json) is set to "/workspaces/" to ensure all static assets are loaded from the correct subpath.
- The React Router <BrowserRouter> is configured with the basename="/workspaces" prop to ensure all client-side navigation is correctly prefixed.
- Navigating to https://<domain>/workspaces/ successfully loads the application.
- All in-app links and navigations correctly maintain the /workspaces/ prefix in the browser's address bar.

PR #517, opened 2025-08-01
feat: Make Frontend Basepath Configurable via APP_PREFIX env variable

thaorell:
closes: #511 

Description:

- Add APP_PREFIX environment variable in Webpack
- Apply the prefix to index.html, Routers and Images


thaorell:
> @thaorell are these changes aligned with the Model Registry repository? I just want to double-check it before starting the review.

@caponetto  We're largely aligned. They're using URL_PREFIX in `const.ts` while the majority of Kubeflow uses APP_PREFIX  so I

ederign (approved):

Issue #499, opened 2025-07-24
[TASK] Introduce Auto-Generated API Types and HTTP Client for Frontend


caponetto:
### Objective

Automate the generation of TypeScript types and a typed HTTP client layer for consuming backend APIs in the frontend area, using the backend's `swagger.json` as the single source of truth.

### Motivation

- Ensure the frontend stays aligned with backend APIs.
- Eliminate manual maintenance of API types and client logic.
- Reduce boilerplate and potential bugs caused by outdated API contracts.
- Improve developer efficiency and confidence when integrating with backend APIs.

### Acceptance Criteria

- Frontend successfully uses the generated HTTP client and types.
- Running `npm run generate:api` produces no diffs after a clean generation.
- CI enforces the generation is current on all frontend PRs.
- Clear documentation exists in `README.md` for frontend developers.
- No coupling required between backend PRs and frontend PRs.
- A clean process is defined to handle API breaking changes in follow-up frontend PRs.

PR #519, opened 2025-08-05
chore(ws): enforce CI failure on uncommitted generated code changes

caponetto:
related: https://github.com/kubeflow/notebooks/issues/499

This PR updates the CI pipeline for the frontend workspace to fail if there are uncommitted changes in generated code. This helps ensure that all generated files are properly committed and that CI reflects the actual state of the codebase.

**Note**: The `fetch-depth: 0` argument is required to fetch the full Git history, since the code lives in `notebooks-v2` branch, allowing us to retrieve the specific Swagger version referenced by the commit hash in the [swagger.version](https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/frontend/scripts/swagger.version) file.

### DIFF
--- a/.github/workflows/ws-frontend-test.yml
+++ b/.github/workflows/ws-frontend-test.yml
@@ -22,6 +22,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: Set up Node.js
         uses: actions/setup-node@v3
@@ -35,6 +37,10 @@ jobs:
       - name: Clean
         working-directory: workspaces/frontend
         run: npm run build:clean
+
+      - name: Regenerate API layer code
+        working-directory: workspaces/frontend
+        run: npm run generate:api
 
       - name: Build
         working-directory: workspaces/frontend


paulovmr (approved):
/lgtm

paulovmr:
Hi @thesuperzapper , although this is related to the frontend, because of the directory of the changed file, we might need your approval as well, if you could take a look please. Thanks!

thesuperzapper:
@caponetto @paulovmr @harshad16 @andyatmiami we were discussing this in the meeting today, and came to the concusion that this is probably the best of the options (that is, allowing the frontend devs to explicitly choose which commit of the swagger json to generate from).

This was because:

1. it ensures that frontend devs don't change the generated files (ci now runs generate and checks for diff)
2. if we always ran the generator based on the HEAD of the branch, it would result in the next frontend PR having to run the generator (and bring in a lot of diff) to have its PR checks pass

The main question is when and who should be updating the generator commit:
- OPTION 1: which ever frontend PR needs the new API feature first (or encounters a bug/breaking change)
- OPTION 2: some regular job which tries to get HEAD to use the latest commit that changed the swagger JSON

Either way, this is an improvement from the current situation, so we can merge.

/lgtm
/approve

PR #568, opened 2025-09-05
fix(nb): fix default workspace volume name to match default configs

mathis-marcotte:
Small fix to improve consistency in the default workspace volume name when creating a new Notebook.

I noticed that on initial load of the crud-web-apps/jupyter form, the default workspace volume name was "{notebook-name}-workspace".
But if you deleted that workspace volume, and then re-added a workspace volume to the Notebook you are creating, then the workspace volume name was "{notebook-name}-volume".

This PR is to FIX the default name of the workspace volume to always follow the same standard of "{notebook-name}-workspace" to match with the default configs.

**Original PR here:** https://github.com/kubeflow/kubeflow/pull/7705

### DIFF
--- a/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-new/form-workspace-volume/form-workspace-volume.component.spec.ts
+++ b/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-new/form-workspace-volume/form-workspace-volume.component.spec.ts
@@ -60,4 +60,24 @@ describe('FormWorkspaceVolumeComponent', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  it('should add new volume', () => {
+    // initialize a blank volGroup
+    component.volGroup = new FormGroup({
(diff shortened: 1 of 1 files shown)


andyatmiami:
Hi @mathis-marcotte - do you think as part of this change you could add an assertion (or provide a new test entirely) here to validate this change as well?
- https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-new/form-workspace-volume/form-workspace-volume.component.spec.ts

Thanks for the contribution as well as re-opening on this "new" repo!
🙇

mathis-marcotte:
Hi @andyatmiami, 
Yes I can totally add that assertion! Once I have some extra time soon, I will get that done.

andyatmiami:
@mathis-marcotte 

thank you for making this addition.. i'm a little back-logged on my _PR reviewing_ - but would expect to be able to focus on this mid-next week.

will be in touch!

PR #577, opened 2025-09-09
chore: Organize .gitignore for frontend

thaorell:
Description:

- Organize `.gitignore` in `workspaces/frontend` to follow standards from controller and backend


workspaces/frontend/.gitignore
```diff
@@ -1,11 +1,18 @@
...
 yarn-error.log
 yarn.lock
+
+# Test files generated by Jest or Cypress
 stats.json
 coverage
+
+# Misc
 .idea
 .vscode/*
```
andyatmiami, line 17: we can likely omit these settings here as they are specified in the _root level_ `.gitignore`
- https://github.com/kubeflow/notebooks/blob/notebooks-v2/.gitignore

Issue #583, opened 2025-09-10
[TASK] Define first release artifacts and branches for notebooks-v1


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

Before creating the first release for the `notebooks-v1` branch, we need to ensure that we are aligned on a couple of details that can affect future releases as well.

Specifically, before the first release we'll need to make sure we've agreed on:

- What image names should we use for the migrated components?
- What branching strategy should we use for cutting releases?
- Which docker tag should we use for the first release?
The above will be quite important to agree on before the release, as it will dictate the future releases as well.

cc @juliusvonkohout @thesuperzapper @kimwnasptd 

### Acceptance Criteria

- [ ] Consensus reached (explicitly or via _lazy consensus_) on image names for `notebooks-v1`
- [ ] Consensus reached (explicitly or via _lazy consensus_) on branching strategy for `notebooks-v1`
- [ ] Consensus reached (explicitly or via _lazy consensus_) on docker tags for `notebooks-v1`

thesuperzapper:
Thanks @andyatmiami, lets go this route.

Note, there is a separate discussion about what to do with the example notebook server images:
- https://github.com/kubeflow/notebooks/issues/585

thesuperzapper:
I have created the following registries and made the `kubeflow/notebooks` CI/CD able to push to them:

```
ghcr.io/kubeflow/notebooks/notebook-controller
ghcr.io/kubeflow/notebooks/tensorboard-controller
ghcr.io/kubeflow/notebooks/pvcviewer-controller
ghcr.io/kubeflow/notebooks/jupyter-web-app
ghcr.io/kubeflow/notebooks/tensorboards-web-app
ghcr.io/kubeflow/notebooks/volumes-web-app
```

PR #604, opened 2025-09-19
feat: Render Loading if Namespaces are fetching

thaorell:
closes: #583

Currently, the UI sends a faulty API call to `/api/v1/workspaces`, indicating no namespace is selected, resulting in a `404` error. Instead, render loading spinner when namespace is being fetched while loading Workspaces.


paulovmr (approved):
/lgtm

PR #626, opened 2025-09-29
feat: Migrate kubeflow/kubeflow jwa_docker_publish.yaml

yehudit1987:
Migrate Jupyter Web App Docker publish workflow from kubeflow/kubeflow 
to kubeflow/notebooks with multi-platform build improvements.

### Migration
- Move workflow to notebooks repo structure
- Update image registry to ghcr.io/kubeflow/notebooks/jupyter-web-app

### Docker Buildx Integration
- Replace sequential make commands with docker/build-push-action@v5
- Fix ARM64 hanging issues 
- Enable parallel multi-platform building (amd64,ppc64le,arm64)

### Improvements
- Add GitHub Actions caching (type=gha) for faster builds
- Eliminate timeout issues with better resource management
- Apply to all build scenarios: SHA tags, latest, and version builds

Fixes ARM64 emulation reliability while maintaining compatibility
with existing deployment workflows. Address issue #616


.github/workflows/jwa_docker_publish.yaml
```diff
@@ -0,0 +1,62 @@
...
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Setup QEMU
+      uses: docker/setup-qemu-action@v3
+      with:
+        platforms: amd64,ppc64le,arm64
```
andyatmiami: this is _interesting_ but not included in the original https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/jwa_docker_publish.yaml

can you comment on the benefits of adding this setting?  and furthermore - could we use `{{env.ARCH}}` here to not have a "future sync issue" with this being separately hard-coded?

If we cannot use `{{env.ARCH}` outright - I'd probably suggest removing this.. but I'm still interested in understanding what prompted the inclusion!

yehudit1987: Ok, so to be honest, I think this is leftover from some previous issues we encountered with the ARM platform. The main purpose of this step is just to specify which platforms to install, whereas the default behavior would be to set up QEMU for all platforms. Unfortunately, we can’t directly use {{env.ARCH}} here because it includes the linux/ prefix. So overall, I think the cleanest approach would be to remove this step entirely.

kimwnasptd:
Thanks for the yet another great PR @yehudit1987!

/approve

PR #666, opened 2025-10-27
feat: add labels to notebook statefulset

monotek:
What:

* Add labels to statefulset created by notebook controller
* Same labels as for the pods are used

Why:

* It's best practice to have labels on all resources
* Be able to exclude statefulset from certain cluster policies (kyverno, opa) by label


kimwnasptd:
Thank you for the PR @monotek! For now we are in the process of having a first release for Notebooks V1.

Once this is out of the way we'll go to reviewing other PRs for V1. Although it's a bit under discussion on what further features/fixes we will be having on the V1 branch, and its support window, considering that there are significant efforts to finalise the V2 release.

Issue #671, opened 2025-10-27
[TASK] Generate Secrets API Client


thaorell:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

We need to refresh our auto-generated client, adding the new type-safe models (e.g., Secret, CreateSecretRequest) and API methods (e.g., listSecrets, createSecret) to include the new Secrets CRUD endpoints.

### Acceptance Criteria

- [ ] Execute generating script.
- [ ] The auto-generated API client file (e.g., src/api/generated/client.ts) includes the new secrets-related interfaces
- [ ] The generated client file also clearly shows the addition of new methods for the secrets endpoints (e.g. listSecrets, getSecret, createSecret, updateSecret).

PR #698, opened 2025-10-29
feat: Generate secrets client

thaorell:
Depends on: #667 

closes: #671 
Generate Secrets client as well as new mocked objects for Secrets Management


paulovmr (approved):
/lgtm

Issue #723, opened 2025-11-09
[TASK] Upgrade Go to 1.24 - pvcviewer-controller component


liavweiss:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

Update the Go toolchain for `/pvcviewer-controller` to version 1.24.

### Scope:
- Update `go.mod` → go 1.24
- Update `Dockerfile`:
```
ARG GOLANG_VERSION=1.24
FROM golang:${GOLANG_VERSION}
```

- Update CI workflow `pvcviewer_controller_unit_test.yaml` → go-version: "1.24"
- Run:
`cd notebooks/components/pvcviewer-controller && go mod tidy && go build ./...
`
- Re-run unit/integration workflows.

### Acceptance Criteria

- Module compiles and builds under Go 1.24.
- All tests pass and Docker image builds successfully.
- Review all files under `/pvcviewer-controller` to ensure no additional Go-related updates are required.

PR #734, opened 2025-11-10
chore: upgrade to go 1.24 for pvcviewer-controller

henschwartz:
[TASK] Upgrade Go to 1.24 - pvcviewer-controller component #723

- Update CI workflow pvcviewer_controller_unit_test.yaml to use go-version 1.24
- Update go.mod to go 1.24
- Update Dockerfile GOLANG_VERSION to 1.24

### DIFF
--- a/.github/workflows/pvcviewer_controller_unit_test.yaml
+++ b/.github/workflows/pvcviewer_controller_unit_test.yaml
@@ -19,7 +19,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v5
       with:
-        go-version: "1.22.2"
+        go-version-file: 'components/pvcviewer-controller/go.mod'
         check-latest: true
 
     - name: Run unit tests
--- a/components/pvcviewer-controller/Dockerfile
+++ b/components/pvcviewer-controller/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-ARG GOLANG_VERSION=1.22.2
+ARG GOLANG_VERSION=1.24
 FROM golang:${GOLANG_VERSION} as builder
 
 WORKDIR /workspace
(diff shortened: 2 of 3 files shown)


henschwartz:
PVCViewer Controller Integration Test:
image
PVCViewer Contoller Multi-Arch Build Test:
image
Run PVCViewer Controller unit tests:
image
PVCViewer Controller Docker Publish:
image

{4625F31E-C85A-44E7-BEB8-0B5228D42CA3}

liavweiss:
Hey @henschwartz, following my earlier review, I noticed one more thing that could be improved. It’s not mandatory, but it would make the code cleaner.
https://github.com/kubeflow/notebooks/blob/77c656d84d853698fb2a25aeb5583e0e5597aef9/components/pvcviewer-controller/Dockerfile#L19

I checked our environment and the module cache is writable, so `-mod=mod` doesn’t add anything in Go 1.20+.
`GO111MODULE=on` is also deprecated and has no effect since modules are always enabled now.

We can remove these flags, though it’s not required for the PR approval.

.github/workflows/pvcviewer_controller_unit_test.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v5
       with:
-        go-version: "1.22.2"
+        go-version: "1.24"
```
andyatmiami:
```suggestion
        go-version-file: 'components/pvcviewer-controller/go.mod'
```

Given we already need to adjust this code block - I think we would be better served to change it more _substantially_ to avoid having to hardcode (and couple) this `go-version` _input_ with the related `go.mod` file

You can see this pattern employed in `notebooks-v2` code - and I think we should have `notebooks-v1` follow suit!  Will make any future updates easier - as we only need to modify `go.mod` and the GH workflow will then automatically pick up the right version.
- https://github.com/kubeflow/notebooks/blob/notebooks-v2/.github/workflows/ws-backend-test.yml#L31
    - note: for now lets **NOT** include the `cache` related key here - even though it is present in `notebooks-v2`

andyatmiami (changes requested):
Hey @henschwartz - thanks for this contribution 💯 

While I have tested the code in my own `Kind` cluster and feel confident about the changes - I'm hoping we can use this PR (opportunity) to improve the flexibility of the unit test configuration to make it _less redundant_ .  

If you can code and verify the simple suggestion I provide below - I will feel good about `lgtm`ing the PR quickly afterwards.

Issue #692, opened 2025-10-29
[TASK] Stub Logo/Icon API endpoints on backend


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

We want a _stubbed implementation_ of the Logo/Icon API as outlined in #689  .

Definition of a "stubbed implementation":
- Changes are expected to be delivered under following folders:
    - `workspaces/backend/api`
    - `workspaces/backend/internal/models`
    - `workspaces/backend/openapi`
        - ℹ️ files under this directory are auto-generated from `swag` `Makefile` command
 - Data returned by the API should be hard-coded and provide a minimally viable experience for the `frontend` to be able to exercise the API!
 - Changes should **NOT** be provided in `workspaces/backend/internal/repositories`
     - These changes would be included when working on the **real** implementation of the endpoints
 - Basic unit testing should still be provided for the changes
     - These tests would obviously require asserting on the hard-coded "mock" responses and will require rework then when we implement these endpoints **for real**

### Acceptance Criteria

- [ ]`swagger.json` includes all the appropriate metadata to expose APIs as outlined in #689

andyatmiami:
/close

work has been delivered

PR #736, opened 2025-11-10
feat: serve workspacekind assets (logos + icons)

andyatmiami:
closes: #692 

This commit implements serving of workspacekind assets (logos and icons) from ConfigMaps through the backend API, with SHA256 hash support for cache-busting.

Key changes:
- Add endpoint to serve workspacekind assets from ConfigMaps at `/api/v1/workspacekinds/{name}/assets/{icon|logo}.svg`
- Compute and include SHA256 hashes for ConfigMap-based assets as query parameters for cache-busting
- Add ConfigMap client support to repositories for accessing asset content
    - partial cache for performance/efficiency that keys off label:
        - `notebooks.kubeflow.org/image-source: true`
- Add GetConfigMapContent method to `WorkspaceKindRepository`  and `WorkspaceRepository` for retrieving asset content from ConfigMaps
- Update model functions to support SHA256 hashes in asset URLs:
  - NewWorkspaceKindModelFromWorkspaceKindWithAssetHashes
  - NewWorkspaceModelFromWorkspaceWithAssetHashes
- Added workspacekind_imagesource_configmap.yaml for reference example
- Refactored `backend` `mgr` setup in `main.go` to leverage `helper/k8s.go`
    - easier to keep logic in sync in e2e tests - particularly with new partial cache

Remote URL specification of assets is still supported - but the browser queries
the source URL directly for content.

#### SCREENSHOTS
image
image
image


thaorell (changes requested):
While reviewing the implementation for `/api/v1/workspacekinds/{name}/assets/{logo|icon|.svg`, I did a quick scan of the existing data flow.
It looks like the frontend/client already receives the full source URL for these icons and logos from ImageRef as part of the parent `WorkspaceKind` object or a payload from `WorkspaceKindInfo` in loading workspace(s).

If that's the case, could we simply have the browser render those URLs directly? That would allow us to avoid maintaining a dedicated proxy endpoint and increase maintainability.

Let me know if I’m missing a specific use case where this proxy is required

christian-heusel (changes requested):
Tested with the changes from #906, it seems like we also need to account for potentially differently sized images in the frontend:
image

Image

thesuperzapper:
@andyatmiami after our review I have pushed https://github.com/kubeflow/notebooks/pull/736/commits/eed53427e6f83353e780e7eb06f3b08eb52f5621 with changes discussed. 

You should test and rebase so we can merge soon.

thesuperzapper:
@andyatmiami I rebased it and created https://github.com/kubeflow/notebooks/pull/736/commits/73863995c61f4e66a2816877597eb2fa0a125ea8 to bring the new code up to speed with the changes/refactor of constants.

The one strange thing is that at some point in the last few commits we have broken the swagger UI (at least in tilt), and we now get un-templated fields at the top, I think this happens at top of tree right now anyway (so probably should not block this PR), but we should figure out what happend and fix it.

Screenshot 2026-05-01 at 01 38 51

thesuperzapper:
Even just before my own rebase it was fine when at the commit https://github.com/kubeflow/notebooks/pull/736/changes/eed53427e6f83353e780e7eb06f3b08eb52f5621
Screenshot 2026-05-01 at 01 41 31

andyatmiami:
/lgtm

reviewed the commits introduced by @thesuperzapper and applied a couple small fixes necessary to get successful end to end behavior through our `tilt` file

#### Swagger UI renders appropriately

image


#### API calls return SVG content 

image

image

thesuperzapper:
This is great, after all this time we are ready to merge this!

Lets make a follow up action to use the pattern of using `DescribeTable` established in [`./workspaces/backend/api/workspacekind_assets_handler_test.go`](https://github.com/andyatmiami/kubeflow-notebooks/blob/30cea2f5dd28947df8374596b2a821eecd378ddb/workspaces/backend/api/workspacekind_assets_handler_test.go) on all the other backend handler tests.

/approve
/lgtm

PR #750, opened 2025-11-17
chore: workspace update api stub

andyatmiami:
Closes: #754

Implement PUT /workspaces/{namespace}/{name} endpoint for updating
existing workspaces with optimistic concurrency control using
revision tokens.

Key changes:
- Add UpdateWorkspaceHandler with path/body validation and auth
- Implement revision-based optimistic locking using SHA256 hash of
  `<uid>:<name>:<generation>` to detect concurrent modifications
- Introduce WorkspaceUpdate model for update requests/responses
- Add WorkspaceListItem model for list endpoints (separation of
  concerns from update model)
- Add secret volume mounts support to PodVolumesMutate
- Add ErrWorkspaceGenerationConflict for 409 responses
- Update OpenAPI documentation with new endpoint and schemas
- Refactor GetWorkspace to return WorkspaceUpdate for edit flow

The revision token ensures clients are updating the expected version
of a workspace. Clients must include the revision received from GET
in their PUT request body. A 409 Conflict is returned if the current
generation differs from what the revision represents.

Note: The actual Kubernetes update logic is stubbed for further
implementation (marked with TODO).


thesuperzapper:
@andyatmiami this should be good to merge now but you can do a quick check.

We were on a call together and did a bunch of fixes, including some follow-ups to the secrets api stub PR that will help people implement it faster:
- https://github.com/kubeflow/notebooks/pull/667

PR #755, opened 2025-11-18
chore: upgrade to go 1.24 for tensorboard-controller

mkoushni:
Goal

This PR updates the required Go version for the tensorboard-controller component to Go 1.24.x. This change is necessary to maintain compatibility with updated Kubernetes client libraries and ensure the component is built with current, supported security patches.

Changes Made

Updated the go.mod file to specify go 1.24.

Synchronized dependencies via go mod tidy and go mod vendor to lock in compatible library versions (including necessary updates to transitive dependencies like golang.org/x/text, etc.).


andyatmiami (changes requested):
Thanks @mkoushni - appreciate your changes.

We are very close - given the poor (initial) state of the `Dockerfile` prior to your changes - just using this PR as a means to _right those wrongs_.  

Apply those changes (and ask any questions as you see fit!) - but imagine we can get this merged fairly quickly after the `Dockerfile` changes are delivered.

liavweiss:
close: #722

PR #763, opened 2025-11-22
ci: scaffold e2e testing framework infrastructure

andyatmiami:
ℹ️  _NO GH ISSUE_

This commit establishes the foundational infrastructure for end-to-end testing of the workspaces components. This is an intermediate step that sets up the testing framework; no actual test execution is implemented yet and will be added in subsequent work.

Changes include:

- Add new `testing/` directory with Makefile and setup scripts:
  * `setup-kind.sh`: Automated Kind cluster creation and configuration
  * `setup-cert-manager.sh`: Cert-manager installation (v1.12.13 LTS)
  * `setup-istio.sh`: Istio service mesh installation
  * Makefile targets for cluster setup, component deployment, and e2e test execution (currently placeholder with TODO)

- Add GitHub Actions workflow (`.github/workflows/ws-e2e-test.yml`):
  * Triggers on pushes to main branches and PRs affecting workspaces
  * Sets up Go environment and Kind cluster
  * Executes local-e2e target (placeholder for future test implementation)

The framework enables automated deployment of all three components (controller, backend, frontend) to a Kind cluster with cert-manager and Istio pre-configured, providing the foundation for comprehensive e2e test scenarios in future commits.

**VERIFICATION**
You can see what this looks like when I ran it on my fork:
- https://github.com/andyatmiami/kubeflow-notebooks/actions/runs/24545624945/job/71760430340?pr=84


testing/scripts/setup-kind.sh
christian-heusel: Can we maybe instead reuse some code from [`developing/scripts/setup-kind.sh`](https://github.com/kubeflow/notebooks/blob/notebooks-v2/developing/scripts/setup-kind.sh) to keept it more DRY? 🤔

testing/scripts/kind.yml
```diff
@@ -0,0 +1,18 @@
...
+      name: config
+    apiServer:
+      extraArgs:
+        "service-account-issuer": "kubernetes.default.svc"
+        "service-account-signing-key-file": "/etc/kubernetes/pki/sa.key"
+nodes:
+- role: control-plane
+  image: kindest/node:v1.33.1@sha256:050072256b9a903bd914c0b2866828150cb229cea0efe5892e2b644d5dd3b34f
+- role: worker
+  image: kindest/node:v1.33.1@sha256:050072256b9a903bd914c0b2866828150cb229cea0efe5892e2b644d5dd3b34f
```
christian-heusel: `kubeflow/manifests` tests against KinD 1.34 right now, maybe we should do the same? 🤔 
https://github.com/kubeflow/manifests/blob/1ad5eee77fdecdf2d3fc43e0b69a8784c312317d/tests/install_KinD_create_KinD_cluster_install_kustomize.sh#L58-L65

christian-heusel (changes requested):
Nice work @andyatmiami 🔥 

I think there are a few things that can be tweaked before we can get this over the finish line, I have left a few review comments with points that I found during a quick review below 😊

A few of those remarks (regarding the `.gitignore` for example) are a bit nitpicky, feel free to ignore them at your liking 🤗

workspaces/controller/.gitignore
```diff
@@ -13,6 +13,7 @@ Dockerfile.cross
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
+
```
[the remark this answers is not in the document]
andyatmiami: thanks. good catch.  will drop this file from PR.

workspaces/backend/.gitignore
```diff
@@ -13,5 +13,6 @@ Dockerfile.cross
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
+
```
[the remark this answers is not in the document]
andyatmiami: thanks. good catch. will drop this file from PR.

testing/Makefile
```diff
@@ -0,0 +1,150 @@
...
+	$(call go-install-tool,$(KIND),sigs.k8s.io/kind,$(KIND_VERSION))
+
+.PHONY: check-kubectl
+check-kubectl: ## Verify that kubectl is available in PATH.
+	@if ! command -v $(KUBECTL) >/dev/null 2>&1; then \
+		echo "✗ ERROR: kubectl is not installed or not found in PATH"; \
+		echo "  Please install kubectl: https://kubernetes.io/docs/tasks/tools/#kubectl"; \
+		exit 1; \
+	fi
+	@echo "✓ kubectl found: $$($(KUBECTL) version --client 2>/dev/null || echo 'version check failed')"
```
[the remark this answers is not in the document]
andyatmiami: `kubectl` is available by default in the GHA runner - but to date all our scripts/Makefiles have just "assumed" `kubectl` is there...

admittedly it _could be fixed_ but then should be done across all our `Makefile` - something I won't do _in this PR_

testing/scripts/setup-kind.sh
andyatmiami: general direction I had gotten from @thesuperzapper was to (for the time being) keep `developing/` and `testing/` folders mutually exclusive.

.github/workflows/ws-e2e-test.yml
```diff
@@ -0,0 +1,59 @@
...
+        working-directory: testing
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache: false
```
christian-heusel: I think similar to the other GHA definitions we should set this up to derive the go version from `go.mod`:
https://github.com/kubeflow/notebooks/blob/4db85a2333193b6a0d2cbee3c4a1ff3449bf3070/.github/workflows/ws-backend-test.yml#L29-L34

Maybe this is material for a follow-up PR (which I could raise if desired), but we could create a go workspace file: https://go.dev/doc/tutorial/workspaces

testing/Makefile
```diff
@@ -0,0 +1,152 @@
...
+
+.PHONY: istioctl
+istioctl: $(ISTIOCTL) ## Download istioctl locally if necessary.
+$(ISTIOCTL): $(LOCALBIN)
+	$(call go-install-tool,$(ISTIOCTL),istio.io/istio/istioctl/cmd/istioctl,$(ISTIOCTL_VERSION),"-X istio.io/istio/pkg/version.buildVersion=$(ISTIOCTL_VERSION) -X istio.io/istio/pkg/version.buildTag=$(ISTIOCTL_VERSION) -X istio.io/istio/pkg/version.buildHub=docker.io/istio")
+
+.PHONY: dependency-hash
+dependency-hash: ## Calculate hash of dependency versions for caching.
+	@echo -e "KIND_VERSION=$(KIND_VERSION)\nISTIOCTL_VERSION=$(ISTIOCTL_VERSION)" | sha256sum | cut -d' ' -f1 | head -c 16
```
christian-heusel: Any reason for the `head -c 16` here? 🤔 From how I understand https://github.com/actions/cache?tab=readme-ov-file#creating-a-cache-key I didn't see any restrictions on specific cache key size so maybe using the full shasum is the better way here?

.github/workflows/ws-e2e-test.yml
```diff
@@ -0,0 +1,59 @@
...
+      GO_VERSION: '1.24.13'
+    defaults:
+      run:
+        working-directory: testing
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
```
christian-heusel:
```suggestion
        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
```

.github/workflows/ws-e2e-test.yml
```diff
@@ -0,0 +1,59 @@
...
+  e2e-test:
+    runs-on: ubuntu-latest
+    env:
+      GO_VERSION: '1.24.13'
+    defaults:
+      run:
+        working-directory: testing
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
christian-heusel:
```suggestion       
        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
```

.github/workflows/ws-e2e-test.yml
```diff
@@ -0,0 +1,59 @@
...
+          cache: false
+
+      - name: Generate cache key from Makefile
+        id: cache-key
+        run: |
+          VERSION_HASH=$(make dependency-hash)
+          echo "cache_key=testing-bin-${{ runner.os }}-go${{ env.GO_VERSION }}-$VERSION_HASH" >> $GITHUB_OUTPUT
+
+      - name: Cache testing/bin directory
+        uses: actions/cache@v4
```
christian-heusel: This action should also be pinned for consistency with the other actions 😊 
```suggestion
        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
```

developing/scripts/kind.yaml
```diff
@@ -13,6 +13,6 @@ kubeadmConfigPatches:
         "service-account-signing-key-file": "/etc/kubernetes/pki/sa.key"
 nodes:
 - role: control-plane
-  image: kindest/node:v1.33.1@sha256:050072256b9a903bd914c0b2866828150cb229cea0efe5892e2b644d5dd3b34f
+  image: kindest/node:v1.34.3@sha256:08497ee19eace7b4b5348db5c6a1591d7752b164530a36f855cb0f2bdcbadd48
 - role: worker
-  image: kindest/node:v1.33.1@sha256:050072256b9a903bd914c0b2866828150cb229cea0efe5892e2b644d5dd3b34f
\ No newline at end of file
+  image: kindest/node:v1.34.3@sha256:08497ee19eace7b4b5348db5c6a1591d7752b164530a36f855cb0f2bdcbadd48
```
christian-heusel: Any reason to go for `1.34.3` here? Latest would be `1.35.1`: https://hub.docker.com/layers/kindest/node/v1.35.1/images/sha256-b08d1f9b5244fb2aae7bea9861c1d372a9535f259efed62a9e92e6b5ab500529

testing/scripts/sanity-check.sh
```diff
@@ -0,0 +1,194 @@
...
+  kubectl port-forward "${pf_target}" "${local_port}:${target_port}" -n "${NAMESPACE}" &
+  pf_pid=$!
+
+  # Ensure we clean up the port-forward on exit from this function
+  trap "kill ${pf_pid} 2>/dev/null || true; trap - RETURN" RETURN
+
+  # Wait for port-forward to be ready
+  local retries=0
+  local max_retries=10
+  while ! curl -s -o /dev/null "http://localhost:${local_port}${probe_path}" 2>/dev/null; do
```
christian-heusel: I prefer long versions of these arguments, also we should probably add `--fail`:
```suggestion
  while ! curl --silent --output /dev/null "http://localhost:${local_port}${probe_path}" 2>/dev/null; do
```

testing/scripts/sanity-check.sh
```diff
@@ -0,0 +1,194 @@
...
+check_http_endpoint() {
+  local label="$1"
+  local pf_target="$2"
+  local target_port="$3"
+  local probe_path="$4"
+  local local_port
+  local pf_pid
+
+  # Pick a random local port to avoid collisions
+  local_port=$((RANDOM % 50000 + 10000))
```
christian-heusel, line 48: This nerd-sniped me a bit and I went to go check how we could reliably choose a free port without the random assignment above (I think the above is totally sufficient for the setup we're going for, I thought I'd share nonetheless):

`kubectl port-forward` can choose a free port itself (or rather the kernel gives it one) if we use no argument or 0 as port to bind:
```
# equivalent to kubectl port-forward svc/istio-ingressgateway -n istio-system 0:80
$ kubectl port-forward svc/istio-ingressgateway -n istio-system :80
Forwarding from 127.0.0.1:46515 -> 8080
Forwarding from [::1]:46515 -> 8080
```
However it's a bit of a pain to actually extract this port from the command output, so the easier solution is to do the equivalent thing in python:
```bash
$ python -c "import socket; s = socket.socket(); s.bind(('', 0));print(s.getsockname()[1]);s.close()"
42017
```

As said, for the given proposal this makes things a bit more complicated than they need to be, but maybe it's useful for something else at some point 😊

testing/scripts/sanity-check.sh
```diff
@@ -0,0 +1,194 @@
...
+  local retries=0
+  local max_retries=10
+  while ! curl -s -o /dev/null "http://localhost:${local_port}${probe_path}" 2>/dev/null; do
+    retries=$((retries + 1))
+    if [ ${retries} -ge ${max_retries} ]; then
+      echo "✗ ERROR: ${label} endpoint not reachable after ${max_retries} attempts"
+      return 1
+    fi
+    sleep 2
+  done
```
christian-heusel: We can do all of this purely with `curl`:
```bash
if ! curl --silent --output /dev/null --fail \                                                          
     --retry 10 --retry-delay 2 --retry-all-errors \                                                    
     "http://localhost:${local_port}${probe_path}"; then                                                
  echo "✗ ERROR: ${label} endpoint not reachable after 10 attempts"                                     
  return 1                                                                                              
fi
```

However this is more of a nit/hint, the current code seems to do the right thing 🤗

testing/scripts/sanity-check.sh
```diff
@@ -0,0 +1,194 @@
...
+  # Verify we get a successful response
+  local http_code
+  http_code=$(curl -s -o /dev/null -w "%{http_code}" "http://localhost:${local_port}${probe_path}")
+
+  if [ "${http_code}" -ge 200 ] && [ "${http_code}" -lt 400 ]; then
+    echo "✓ ${label} HTTP endpoint returned ${http_code}"
+  else
+    echo "✗ ERROR: ${label} HTTP endpoint returned ${http_code}"
+    return 1
+  fi
```
christian-heusel, line 70: We could also use the `--fail` arg from `curl` instead of the `if` construct here:
> (HTTP) Fail with error code 22 and with no response body output at all for HTTP transfers returning HTTP response codes at 400 or greater.

christian-heusel (commented):
Nice work, I took a second look at this PR after the rebase and had a few minor things to note, but most of it are optional improvements / remarks, feel free to implement or ignore as desired 😊

testing/scripts/setup-istio.sh
```diff
@@ -0,0 +1,51 @@
...
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+TESTING_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+LOCALBIN="${TESTING_DIR}/bin"
+
+# Determine istioctl path - prefer LOCALBIN, fallback to PATH
+if [ -f "${LOCALBIN}/istioctl" ]; then
+  ISTIOCTL="${LOCALBIN}/istioctl"
+elif command -v istioctl >/dev/null 2>&1; then
+  ISTIOCTL="istioctl"
```
thesuperzapper: Lets remove the fallback to global `istioctl` so we know for a fact (if the user is calling this from make) that we have a specific version of `istioctl`.

Lets also make this change on the [`./developing/scripts/setup-istio.sh`](https://github.com/kubeflow/notebooks/blob/notebooks-v2/developing/scripts/setup-istio.sh) one which is subject to the same flaw.

thesuperzapper (changes requested):

testing/Makefile
```diff
@@ -0,0 +1,152 @@
...
+
+deploy-backend: kind check-kind-context ## Build and deploy the backend.
+	cd ../workspaces/backend && $(MAKE) docker-build IMG=$(BACKEND_IMG)
+	$(KIND) load docker-image $(BACKEND_IMG) --name $(KIND_CLUSTER_NAME)
+	cd ../workspaces/backend && $(MAKE) deploy IMG=$(BACKEND_IMG)
+
+deploy-frontend: kind check-kind-context ## Build and deploy the frontend.
+	cd ../workspaces/frontend && $(MAKE) docker-build IMG=$(FRONTEND_IMG)
+	$(KIND) load docker-image $(FRONTEND_IMG) --name $(KIND_CLUSTER_NAME)
+	cd ../workspaces/frontend && $(MAKE) deploy IMG=$(FRONTEND_IMG)
```
christian-heusel: It's also more of a style question but you could also use `make -C` for these targets here:

```
deploy-frontend: kind check-kind-context ## Build and deploy the frontend.
	$(MAKE) -C ../workspaces/frontend docker-build IMG=$(FRONTEND_IMG)
	$(KIND) load docker-image $(FRONTEND_IMG) --name $(KIND_CLUSTER_NAME)
	$(MAKE) -C ../workspaces/frontend deploy IMG=$(FRONTEND_IMG)
```

https://www.gnu.org/software/make/manual/make.html#Recursion

I'm not quite sure what style I like more (especially regarding readability), but I wanted to point out the option just in case 🤗

.github/workflows/ws-e2e-test.yml
andyatmiami: i will open something with `go.work` as i think that does warrant a separate change but I like the suggestion 💯

testing/Makefile
andyatmiami: removed!

testing/scripts/sanity-check.sh
andyatmiami: i like long names too!

testing/scripts/setup-istio.sh
andyatmiami: fallback removed

testing/Makefile
andyatmiami: i decided i think I like `-C` - so lets go!

.github/workflows/ws-e2e-test.yml
andyatmiami: Opened this PR to address this point (amongst other improvements/consolidations 😇 )
- https://github.com/kubeflow/notebooks/pull/1099

testing/scripts/sanity-check.sh
andyatmiami: The current code gives better diagnostic output (prints the actual HTTP code on success/failure). --fail would silently exit with code 22, losing that information. The current approach is more debuggable for CI.

testing/scripts/sanity-check.sh
andyatmiami: i like it!

developing/scripts/kind.yaml
andyatmiami: I used v1.35.0 (not v1.35.1 ) because v1.35.0 is the version pinned with a SHA256 digest in the official kind v0.31.0 release.

.github/workflows/ws-e2e-test.yml
christian-heusel: Nice work, we an consider this remark resolved in the context of this PR then 💯

testing/scripts/sanity-check.sh
```diff
@@ -0,0 +1,175 @@
...
+  local retries=0
+  local max_retries=10
+  while ! echo | openssl s_client -connect "localhost:${local_port}" 2>/dev/null | grep -q "CONNECTED"; do
+    retries=$((retries + 1))
+    if [ ${retries} -ge ${max_retries} ]; then
+      echo "✗ ERROR: ${label} TLS handshake failed after ${max_retries} attempts"
+      return 1
+    fi
+    sleep 2
+  done
```
christian-heusel, line 107: Any specific reason we're using `openssl` here over `curl` which we use in more places? 🤔 
If it's due to the self-signed cert we can just use `--insecure`:
```
     -k, --insecure
            (TLS SFTP SCP) By default, every secure connection curl makes is verified to be secure before the transfer  takes  place.
            This option makes curl skip the verification step and proceed without checking.
```

With this added depending on the HTTP tool we could also deduplicate the functionality with `check_http_endpoint()` and save a bit of complexity 😊

testing/scripts/setup-istio.sh
```diff
@@ -0,0 +1,48 @@
...
+  --timeout=120s
+
+echo "Waiting for Istio ingress gateway to be ready..."
+kubectl wait --for=condition=ready pod \
+  -l app=istio-ingressgateway \
+  -n istio-system \
+  --timeout=120s
+
+# Apply gateway resources directly (in developing/ these are managed by Tilt)
+kubectl create namespace kubeflow --dry-run=client -o yaml | kubectl apply -f -
```
christian-heusel: Creating the namespace this way will not apply needed annotations like PSS labels that the manifests use in this case, see:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: kubeflow
  labels:
    control-plane: kubeflow
    istio-injection: enabled
    pod-security.kubernetes.io/enforce: restricted
```

https://github.com/kubeflow/manifests/blob/master/common/kubeflow-namespace/base/kubeflow/namespace.yaml

christian-heusel (approved):
Only had two minor comments, left a LGTM 👍🏻

developing/scripts/kind.yaml
christian-heusel: https://github.com/kubeflow/notebooks/pull/1080#discussion_r3268939950

andyatmiami:
/lgtm

reviewed live with @thesuperzapper on the commits he pushed to this PR.  i acknowledge and accept these changes.

testing/scripts/setup-istio.sh
thesuperzapper: added the `pod-security.kubernetes.io/enforce: restricted` in https://github.com/kubeflow/notebooks/pull/763/commits/0a49ddcdec273edf44f66f34146f3cd8cdf7a3d6

the others are not needed in these tests

thesuperzapper:
So happy to have this merged, we finally have the base of our E2E tests running on every PR/commit!

/lgtm
/approve

PR #766, opened 2025-11-28
refactor: improve and expand Cypress test coverage

caponetto:
**This PR includes:**

### Cypress infra
- Refactored existing mocked tests for consistency and readability
- Added new mocked tests to cover the existing app flow (going from 9 to 163 tests)
- Removed files that were no longer needed
- Introduced page object models to group common asserts, queries and actions
- Added an `interceptApi` command so API intercepts are typed
- Enabled test retries in run mode to improve CI stability
- Added `data-testid` props to components to make them easier to target in tests

### Bug fixes
- Fixed an issue where the workspace table disappears after applying a filter when the expected row wasn't on the current page  
- Fixed an error that happened when editing a workspace image during “edit workspace kind” while no port was defined


ederign (approved):
/approve
/lgtm

PR #771, opened 2025-12-04
chore: fix envtest file permissions + backend k8s version

thesuperzapper:
This PR does two things:

1. Fixes the fact that we forgot to set `ENVTEST_K8S_VERSION` in the backend makefile, meaning we were always testing with the latest version of Kubernetes in unit tests, rather than a specific one.
2. Introduces a workaround for the fact that `setup-envtest` [sets the file permissions of the binaries it downloads to `0555`](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.22.4/tools/setup-envtest/store/store.go#L191) (not writable by owner), this was making it painful to delete the binaries and confusing contributors.

### DIFF
--- a/workspaces/backend/Makefile
+++ b/workspaces/backend/Makefile
@@ -9,6 +9,10 @@ IMG ?= $(REGISTRY)/$(NAME):$(TAG)
 ARCH ?= linux/arm64/v8,linux/amd64,linux/ppc64le
 
 
+# The version of Kubernetes to use for envtest.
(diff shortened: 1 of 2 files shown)


andyatmiami:
/ok-to-test
/lgtm
/approve

thanks for the rabbit hole on this one @thesuperzapper .

I did a live coding session with @thesuperzapper while he wrote up this commit and observed appropriate testing performed to ensure code achieves desired outcome.

PR #773, opened 2025-12-05
chore: update v1 readme + disallow semantic scopes

thesuperzapper:
This PR updates the `notebooks-v1` branch readme with info about v1/v2.

It also updates the semantic commit GHA to prevent using scopes in commit messages, as we now use prow OWNERS to add labels based on files changed.

It also removes the files which only need to be in the `main` branch, like `CONTRIBUTING.md` and `.github/ISSUE_TEMPLATE/*`.

It also moves the `area/v1` label config to the root OWNERS, so it always affects PRs into the `notebooks-v1` branch.

### DIFF
--- a/.github/workflows/OWNERS
+++ b/.github/workflows/OWNERS
@@ -1,3 +1,2 @@
 labels:
   - area/ci
-  - area/v1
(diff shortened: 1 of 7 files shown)


thesuperzapper:
@andyatmiami I will approve this as I am the only one who can, please LGTM if you agree.

/approve
/ok-to-test

PR #787, opened 2025-12-08
fix: update jsonpath-plus to 10.3.0 across CRUD web apps (Security v1.11) 

utruong309:
This PR updates jsonpath-plus to version 10.3.0 across all CRUD web app frontends:

- common/frontend
- jupyter/frontend
- tensorboards/frontend
- volumes/frontend

This addresses:
- CVE-2024-21534
- CVE-2025-1302

Steps performed:
- Updated package.json and package-lock.json
- Ran npm install in each frontend directory
- Verified lint and build succeed

Signed-off-by: Uyen Truong <uyenthutruong09@gmail.com>

### DIFF
--- a/components/crud-web-apps/common/frontend/kubeflow-common-lib/package-lock.json
+++ b/components/crud-web-apps/common/frontend/kubeflow-common-lib/package-lock.json
@@ -26,6 +26,7 @@
         "@fortawesome/free-brands-svg-icons": "^5.12.0",
         "@fortawesome/free-solid-svg-icons": "^5.12.0",
         "date-fns": "1.29.0",
+        "jsonpath-plus": "10.3.0",
         "lodash": "^4.17.21",
         "lodash-es": "^4.17.21",
         "material-icons": "^0.7.7",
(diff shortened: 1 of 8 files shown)


andyatmiami (changes requested):
@utruong309 

Appreciating you getting engaged in the community and helping with the critical work of keeping our codebases vulnerability-free! 💯 

When it comes to this particular work around `jsonpath-plus` - I would like to propose an alternative approach that I think will be cleaner in the long run...

`(1)` `jsonpath-plus` is actually only a transitive dependency - being pulled in through `@kubernetes/client-node`
`(2)` the usage of `@kubernetes/client-node` in the `crud-web-apps` directory is **very limited**.  we only use this library to pull in TS types of various k8s objects.
`(3)` TS type definitions for the k8s objects we are referencing are **very** stable

Given points 1,2,3 above - I'd appreciate if you could explore upgrading `@kubernetes/client-node` to version `^1.4.0`.  This actually includes the `10.3.0` version of `jsonpath-plus` - but leads to a simpler `package.json` since we don't need to add/manage non-direct dependencies.

From my cursory testing - I think this `@kubernetes/client-node` upgrade will be safe and trivial.. and with your additional eyes/involvement - I'm confident we can make it work.

Thoughts/questions? Let me know!

Issue #779, opened 2025-12-05
[TASK] Remove Automatic Redirect Application from Controller


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

This issue tracks the implementation work to remove automatic redirect application from the Workspace controller and complete the Workspace Update API to support user-initiated configuration changes.

#### Background

The controller currently contains logic that automatically applies redirects when a Workspace is paused and `spec.deferUpdates` is false. This block (~lines 218-237 in `workspace_controller.go`) mutates `Workspace.Spec.PodTemplate.Options.ImageConfig` and `.PodConfig` based on redirect chains defined in `WorkspaceKind`.

This automatic behavior is being removed in favor of explicit user-initiated updates via the PUT API.

#### What to KEEP

The controller must continue to:

- Compute redirect chains using `getImageConfig()` and `getPodConfig()` functions
- Populate status fields for dashboard consumption:
  - `status.podTemplateOptions.imageConfig.desired`
  - `status.podTemplateOptions.imageConfig.redirectChain`
  - `status.podTemplateOptions.podConfig.desired`
  - `status.podTemplateOptions.podConfig.redirectChain`
  - `status.pendingRestart`
- Validate redirect configurations via WorkspaceKind webhook (no cycles, targets exist)

#### What to REMOVE/CHANGE

| Component | Change |
|-----------|--------|
| Controller auto-update block | Remove the block (~lines 218-237 in `workspace_controller.go`) that checks `PendingRestart && Paused && !DeferUpdates` and mutates `Workspace.Spec` |
| `Workspace.Spec.DeferUpdates` field | Remove from CRD (`api/v1beta1/workspace_types.go` ~lines 39-43) |
| `DeferUpdates` in backend models | Remove from `WorkspaceCreate` and `WorkspaceUpdate` structs (`internal/models/workspaces/types_write.go`) |

#### Documentation Updates

Update comments in `workspaces/controller/config/samples/jupyterlab_v1beta1_workspacekind.yaml` (~lines 242-246) to reflect that redirects are no longer automatically applied by the controller.

#### Controller

- [ ] Auto-update block removed from `workspace_controller.go`
- [ ] `DeferUpdates` field removed from `WorkspaceSpec` in `workspace_types.go`
- [ ] Controller still populates redirect-related status fields:
  - `status.podTemplateOptions.imageConfig.desired`
[...]

PR #804, opened 2025-12-14
feat: remove automatic redirect application from Workspace controller

alokdangre:
This PR removes the automatic application of redirects by the Workspace controller and transitions to a user-initiated update model via the API.

 closes: #779 

**Changes:**
- Removed `DeferUpdates` field from `WorkspaceSpec` in CRD and backend models.
- Removed logic in `workspace_controller.go` that automatically mutated `Workspace.Spec` based on redirects.
- Updated backend API validation and models to exclude `DeferUpdates`.
- Updated frontend to remove `DeferUpdates` checkbox and logic.
- Regenerated CRD manifests and deepcopy code.
- Regenerated frontend API types.
- Updated tests in controller, backend, and frontend.

This ensures that workspace updates (redirects) are only applied when explicitly requested by the user, providing a safer and more predictable update process.


alokdangre:
so as decieded i have made a separate backend only pr https://github.com/kubeflow/notebooks/pull/812

alokdangre:
and i have also made a frontend only changes pr: https://github.com/kubeflow/notebooks/pull/813
which needs a swagger version to be updated, after the backend one merges

PR #815, opened 2025-12-22
refactor: consolidate filters into a single reusable component 

caponetto:
This PR consolidates all filters into a single reusable component. The shared filter is now used by the Workspaces table, Workspace Kinds table, and the workspace kind, image, and pod configuration selectors in the create/edit workspace wizard.

No visual changes are expected overall. The only behavior change is in the Workspaces table, where the "filter by state" now matches the Workspace Kinds "filter by status" by presenting available values in a dropdown instead of requiring free-text input.

Additionally, mock files have been excluded from the test coverage configuration.

https://github.com/user-attachments/assets/f7dfae8f-b57b-436f-ae5e-83f3f7171cc7


paulovmr (approved):
/lgtm

Issue #765, opened 2025-11-25
[TASK] Form components within modals are missing borders


jenny-s51:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

The following UI components are missing borders in the MUI theme:

- Delete workspace modal text field
- Key-value pairs in "Edit Workspace Kind" modals
- Image pull policy dropdown in Workspace Images -> Add image
- "Redirect message level" menu in Workspace Images -> Add image
- "Resource name" field in Pod configurations -> Add config -> Resources -> Add custom resource

<img width="573" height="356" alt="Image" src="https://github.com/user-attachments/assets/dbb0461c-0064-4034-b11a-e393eed5e6c2" />
<img width="735" height="238" alt="Image" src="https://github.com/user-attachments/assets/08a0fcaf-d933-4626-946c-ad4730af932c" />
<img width="726" height="303" alt="Image" src="https://github.com/user-attachments/assets/b901808c-6856-461b-995c-897a43a9d47f" />
<img width="720" height="319" alt="Image" src="https://github.com/user-attachments/assets/725f9c5c-c8a1-4e06-bd97-edb9485d24b4" />
<img width="732" height="553" alt="Image" src="https://github.com/user-attachments/assets/410e4f41-afb5-4c4b-863e-c47bcf3b1808" />

### Acceptance Criteria

- [ ] Wrap components in ThemeAwareFormGroupWrapper where needed
- [ ] Add CSS rules for modal-scoped table inputs if necessary

PR #824, opened 2026-01-09
fix: restore missing borders in MUI theme components

alokdangre:
Wrap several UI components in ThemeAwareFormGroupWrapper to ensure proper styling and borders are applied when using the MUI theme.

Fixes:
- Delete workspace modal text field
Screenshot 2026-01-09 135121

- Key-value pairs in 'Edit Workspace Kind' modals
Screenshot 2026-01-09 135225

- Image pull policy dropdown
Screenshot 2026-01-09 134910

- Redirect message level menu
Screenshot 2026-01-09 134931

- Custom resource name field in Pod configurations
Screenshot 2026-01-09 135034

 closes: #765


alokdangre:
PTAL @jenny-s51

jenny-s51 (commented):
Thank you @alokdangre for your contribution and great work wrapping these components with `ThemeAwareFormGroupWrapper` - this is the correct pattern and we should make sure we are using it consistently. 

Your changes here exposed a bug for a component that was not supported with the current styles (much appreciated!) and https://github.com/opendatahub-io/mod-arch-library/issues/84 needs a fix before we can merge your work here. 

cc @paulovmr, just a quick note that we should hold this one for now and then approve once the above is merged and the new versions are bumped here. WDYT?

workspaces/frontend/src/shared/components/DeleteModal.tsx
```diff
@@ -92,25 +93,32 @@ const DeleteModal: React.FC<DeleteModalProps> = ({
...
-            />
-            {showWarning && (
-              <HelperText data-testid="delete-modal-helper-text">
-                <HelperTextItem icon={<ExclamationCircleIcon />} variant="error">
-                  The name does not match. Please enter exactly: {resourceName}
-                </HelperTextItem>
-              </HelperText>
-            )}
+          </StackItem>
+          <StackItem>
```
jenny-s51:
```suggestion
          <StackItem>
            <Form>
```

jenny-s51 (changes requested):
Ah looks like you can fix the input in the `DeleteModal.tsx` by wrapping with a Form - can you apply this change @alokdangre ?

workspaces/frontend/src/shared/components/DeleteModal.tsx
jenny-s51:
```suggestion
        <Stack>
```

Issue #832, opened 2026-01-13
[TASK] Delete a secret created using Secret Create Modal should trigger API call


thaorell:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

In Workspace Create Wizard, if a user create a secret then remove the secret in WorkspaceFormPropertiesSecrets, it only removes the Secret from the Table, without calling the API to remove it from the Backend.

Ideally, removing a secret should also call the API to prevent idle Secrets building up in the Backend

### Acceptance Criteria

- Deleting a created secret should trigger an API call to `/secrets/${namespace}/${name}`
- Deleting a created secret should update Secrets Table
- Errors should show up in the Modal if the API returns an error
- Write unit tests to handle the scenario

PR #834, opened 2026-01-14
feat: API should be called when removing a Secret created by user

thaorell:
closes: #832 

Description: 
- Use DeleteModal component for deleting Secrets
- Add wrapper types with `isAttached` attribute to distinguish Attached mounts from mounts that are created during Workspace Form
- Deleting secrets that were created by the user on the Workspace Form will trigger API call and update table if call is successful
- Add mocked error handling 

https://github.com/user-attachments/assets/25aff5f4-89cf-4e5f-986e-f47fe53fb01e

### DIFF
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
@@ -185,7 +185,16 @@ const WorkspaceForm: React.FC = () => {
     setError(null);
 
     try {
-      await submitFormData({ mode, data, api, namespace });
+      // Strip the `isAttached` field from secrets before submitting to the API
+      const preparedData: WorkspaceFormData = {
+        ...data,
+        properties: {
+          ...data.properties,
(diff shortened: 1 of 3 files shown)


caponetto:
@thaorell could you please add a couple of tests to cover this?

workspaces/frontend/src/__tests__/cypress/cypress/pages/workspaces/createWorkspace.ts
```diff
@@ -108,6 +108,14 @@ class CreateWorkspace extends WorkspaceForm {
   clickClearAllFilters() {
     return this.findClearAllFiltersButton().click();
   }
+
+  clickAttachExistingSecrets(): Cypress.Chainable<JQuery<HTMLElement>> {
+    return cy.get('[data-testid="attach-existing-secrets-button"]').click();
```
caponetto:
```suggestion
    return cy.findByTestId('attach-existing-secrets-button').click();
```
You can use this utility function instead.

thaorell:
@caponetto fixed the Cypress methods :)

PR #826, opened 2026-01-10
feat: support namespaceFilter on GET /workspacekinds

utruong309:
### Summary
Adds optional `namespaceFilter` query parameter to `GET /workspacekinds` to support Bella-centric listing while preserving existing admin behavior.

### Behavior
- When `namespaceFilter` is provided:
  - Validates the value as a Kubernetes namespace
  - Applies `CREATE Workspace` authorization in the given namespace
- When not provided:
  - Uses existing `LIST WorkspaceKind` authorization

### Motivation
This supports the dual-persona nature of `/workspacekinds`:
- Bella: selecting a WorkspaceKind to create a Workspace
- Joel: administrative listing of WorkspaceKinds

### Acceptance Criteria
- [x] namespaceFilter query parameter supported
- [x] namespaceFilter validated (422 on invalid)
- [x] Authorization policy depends on presence of namespaceFilter
- [x] Existing behavior preserved when parameter is absent

PR #845, opened 2026-01-22
fix: authorization checks

andyatmiami:
⚠️ _NO GH ISSUE_

Discovered during review of #826, the authorization checks were failing for non-cluster-admin users because:

1. Inline-created objects (e.g., &Workspace{}) do not have their TypeMeta
   populated, so GetObjectKind().GroupVersionKind() returns empty values.

2. The Resource field in SubjectAccessReview was being set to the Kind
   (e.g., "Workspace") instead of the plural resource name (e.g., "workspaces").

3. The backend service account was missing permission to create
   SubjectAccessReview resources, causing 500 errors during auth checks.

This fix:
- Uses simple hardcoded map to resolve GV from the new ResourcePolicyResource enum
- Renames the ResourcePolicy.Kind field to ResourcePolicy.Resource for clarity
- Adds subjectaccessreviews create permission to the backend ClusterRole

Refactor:
- Simplified the API signature for NewResourcePolicy to only pass ObjectMeta vs.
a more strongly typed object - since we only care to extract name and/or namespace
if defined

### DIFF
--- a/workspaces/backend/api/app.go
+++ b/workspaces/backend/api/app.go
@@ -22,7 +22,6 @@ import (
 	"net/http"
 
 	"github.com/julienschmidt/httprouter"
-	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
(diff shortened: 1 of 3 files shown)


andyatmiami:
ℹ️ vibe-coded some simple scripts that work against a running `Tilt` cluster to enable auth and then verify various endpoints return proper http statuses

- [run-auth-endpoint-tests.sh](https://github.com/user-attachments/files/24946158/run-auth-endpoint-tests.sh)
- [setup-auth-test.sh](https://github.com/user-attachments/files/24946159/setup-auth-test.sh)

Issue #634, opened 2025-10-02
[TASK] Agree on architecture/implementation details for volume management APIs


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

As a community, we need to agree on how the Volume Management APIs will be designed and implemented.  An initial proposal has been worked on by @andyatmiami and written up as a Markdown file:
- [nbv2-volume-management.md](https://github.com/user-attachments/files/22661048/nbv2-volume-management.md)

We need to review and revise this document so that the expectations on the Volumes API is well-defined and able to be used as a source of truth for implementation.

### Acceptance Criteria

- [ ] API Proposal for Volumes Management API is agreed upon by community

PR #905, opened 2026-02-14
feat: add PVC and StorageClass backend API endpoints

andyatmiami:
closes: #634 
 closes: #693 
 closes: #849 

Implements List, Create, and Delete endpoints for PersistentVolumeClaims and a List endpoint for StorageClasses in the backend API.

PVC List endpoint includes cross-references to:
- Bound PersistentVolume (with StorageClass metadata)
- Pods that mount the PVC
- Workspaces that reference the PVC (via home or data volumes)

Cross-reference collections (pods, workspaces) always serialize as empty arrays `[]` rather than being omitted, matching the pattern used by `services` and `data` on WorkspaceListItem. The `pv` field uses `omitempty` and is omitted when the PVC is unbound.

Pod and Workspace cross-references use a list-once-and-map approach (single List call per resource type, then in-memory lookup) rather than per-PVC queries, since Kubernetes does not support field selectors for filtering pods by PVC claim name.

Notable:
- RBAC includes `persistentvolumes` (get/list/watch) in addition to `persistentvolumeclaims`, since bound PV metadata is resolved for each PVC in the list response.
- PVC Create returns an empty Location header because there is no GET-by-name endpoint for PVCs (only list by namespace).
- PVC sentinel errors use `errors.New` instead of `fmt.Errorf` (without format verbs), aligning with the workspacekinds repository pattern rather than the workspaces/secrets pattern.
- Sample manifests (PVCs, Secret) now include `can-mount` and `can-update` labels so resources created outside the API render correctly in list responses.

### DIFF
--- a/developing/scripts/setup-kind.sh
+++ b/developing/scripts/setup-kind.sh
@@ -28,4 +28,32 @@ fi
 # Ensure kubectl context is set to the Kind cluster
 kubectl config use-context "kind-${CLUSTER_NAME}" || true
 
+# Configure StorageClasses with Notebooks labels and annotations
+echo "Configuring StorageClasses for the Notebooks UI..."
+
+# Label and annotate the default 'standard' StorageClass
+kubectl label storageclass standard \
+  "notebooks.kubeflow.org/can-use=true" \
+  --overwrite
+kubectl annotate storageclass standard \
+  "notebooks.kubeflow.org/display-name=Standard (Local Path)" \
+  "notebooks.kubeflow.org/description=Local path provisioner for development. Data is stored on the node and not replicated." \
+  --overwrite
+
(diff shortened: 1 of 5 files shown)


thesuperzapper:
@andyatmiami as discussed I pushed some more changes in https://github.com/kubeflow/notebooks/pull/905/commits/5450955c21dbcbc7d6a8f05fcc09669efddab00f

Still need to fix up some tests which now fail because we are being more strict about PVC/StorageClasses existing when creating a Workspace/PVC.

andyatmiami:
@thesuperzapper 

pulled your changes (thank you!) and fixed up tests and a few other usability issues...

your changes /lgtm and i think we can finally move forward with merging this 💯

thesuperzapper:
Thanks, this is going to make us much closer to release of notebooks V2.

/lgtm
/approve

PR #913, opened 2026-02-16
feat: implement Workspace Update API method

andyatmiami:
Closes: #848 

This commit delivers the following:
- Complete the UpdateWorkspace repository method which was previously a stub (GH #848).
    - Apply all mutable spec fields from the update model to the K8s Workspace object and persist via client.Update().
- Add handler tests for successful update, 409 stale revision, and 404 not found.

Not included in this PR and to be addressed in a follow up:
- Retrying on 409 response from k8s server to account for stale cache
    - TODO comment still in code calling out this need

### DIFF
--- a/workspaces/backend/internal/models/workspaces/funcs_write.go
+++ b/workspaces/backend/internal/models/workspaces/funcs_write.go
@@ -17,13 +17,41 @@ limitations under the License.
 package workspaces
 
 import (
+	"context"
 	"crypto/sha256"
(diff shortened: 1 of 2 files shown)


andyatmiami:
/hold

need to refactor this a bit to leverage the additional validation abilities exposed in #905

andyatmiami:
/unhold

changes from #905 have been incorporated and an `Update` test has been added for resources that don't have `can-use: true`

Issue #910, opened 2026-02-16
WorkspaceKind validation: deprecationMessage is required when deprecated is false


akrem-chabchoub:
### Checks

- [x] I have searched the [existing issues](https://github.com/kubeflow/notebooks/issues).
- [x] My issue is related to one of the components in the [`kubeflow/notebooks`](https://github.com/kubeflow/notebooks) repository.

### Kubeflow Notebooks Version

v2 (alpha)

### Kubeflow Platform

Tilt

### Kubernetes Distribution

KIND

### Kubernetes Version

```shell
Client Version: v1.32.2
Kustomize Version: v5.5.0
Server Version: v1.33.1
```

### Description

I was trying to create a WorkspaceKind with `deprecated: false` and an empty deprecation message (`deprecationMessage: ""`), but the API returned 422 with:

`Invalid value: \"\": spec.spawner.deprecationMessage in body should be at least 2 chars long`

So non-deprecated kinds cannot have an empty deprecation message, validation forces at least 2 characters even when the kind is not deprecated. 

Expected: when `deprecated` is false, `deprecationMessage` may be empty and the min-length rule should only apply when `deprecated` is true.

### Relevant Logs

```shell

```

PR #923, opened 2026-02-20
fix: allow empty deprecationMessage in WorkspaceKind spawner

akrem-chabchoub:
closes: #910 

## Changes

- **CEL validation** on `WorkspaceKindSpawner`: allow missing, empty, or non-empty with length ≥ 2; otherwise validation fails with a clear message.
- **OpenAPI**: removed `MinLength` from `deprecationMessage` so the schema accepts `""`; non-empty length is enforced only by the CEL rule above.


thesuperzapper:
see https://github.com/kubeflow/notebooks/issues/910#issuecomment-3968436792
/close

PR #925, opened 2026-02-23
chore: use istio in tilt dev environment

andyatmiami:
ℹ️  _NO GH ISSUE_

Add Istio ingress gateway to the local development cluster so that
routing matches the production deployment topology. All traffic now
flows through the gateway at https://localhost:8443, using the same
VirtualService definitions used in production.

Istio setup:
- Add setup-istio.sh script to install Istio with the default profile;
  uses `istioctl verify-install` to skip if already installed
- Add gateway kustomization (developing/manifests/istio-gateway/) with
  self-signed TLS certificate, Kubeflow Gateway, and namespace; these
  are applied by Tilt rather than the setup script
- Add istioctl installation to Makefile (go-install-tool)
- Wire setup-istio as a dependency of `make tilt-up`
- Restructure setup-cert-manager.sh to use if/else instead of exit 0

Tiltfile changes:
- Use istio kustomize overlays for controller, backend, and frontend
  (removes TODO workarounds that pinned components to base manifests)
- Remove controller manifest mutation that force-disabled USE_ISTIO
- Remove direct port-forwards for backend and frontend; traffic now
  routes through the Istio gateway (controller port-forwards are kept
  since it opts out of the Istio sidecar)
- Remove VirtualService URI rewrite for frontend so webpack dev server
  receives requests with the /workspaces/ prefix intact
- Add catch-all /api/ route to backend VirtualService for dev to handle
  swagger redirects that lose the /workspaces/ prefix after URI rewrite
- Apply gateway manifests via k8s_yaml with pod_readiness="ignore" to
  avoid permanent "Runtime Pending" status for CRD objects
- Add istio-ingressgateway local_resource with clickable Tilt UI links;
  conditionally includes frontend link based on ENABLE_FRONTEND
- Update frontend startup probe path to /workspaces/

Frontend dev server:
- Switch to DEPLOYMENT_MODE=standalone with PUBLIC_PATH=/workspaces so
  the UI renders its own NavBar and sidebar (kubeflow mode expects the
  Central Dashboard wrapper which isn't present in dev)
- Remove PROXY_* vars and URL_PREFIX override; Istio handles API routing
- Disable HMR WebSocket client in Tilt mode (hot, liveReload, client)
  since the Istio gateway doesn't route WebSocket traffic; webpack still
  watches and rebuilds, but requires manual browser refresh

Docs:
- Update DEVELOPMENT_GUIDE.md with Istio gateway access URLs, self-
  signed certificate note, and manual refresh requirement

### DIFF
--- a/developing/Tiltfile
+++ b/developing/Tiltfile
@@ -50,10 +50,8 @@ manifests = kustomize(controller_kustomize_path, kustomize_bin=kustomize_bin)
 k8s_yaml(manifests, allow_duplicates=True)
 
 # Manage k8s_yaml resources
-#   - Apply Tilt-specific grouping/labels
-#   - Port-forward controller for local access
-# Note: controller opts out of Istio sidecar (sidecar.istio.io/inject: "false"),
-# so its ports are directly accessible without going through the gateway.
+# - Apply Tilt-specific grouping/labels
+# - Port-forward controller for local access
 k8s_resource(
     "workspaces-controller",
     port_forwards=[
@@ -112,8 +110,8 @@ k8s_yaml(
 )
 
 # Manage k8s_yaml resources
-#   - Apply Tilt-specific grouping/labels
-#   - Resource dependencies on controller
+# - Apply Tilt-specific grouping/labels
+# - Resource dependencies on controller
 k8s_resource(
     "workspaces-backend",
     resource_deps=["workspaces-controller"],
@@ -222,8 +220,8 @@ if enable_frontend:
     )
 
     # Manage k8s_yaml resources
-    #   - Apply Tilt-specific grouping/labels
-    #   - Resource dependencies on backend
+    # - Apply Tilt-specific grouping/labels
+    # - Resource dependencies on backend
     k8s_resource(
         "workspaces-frontend",
         # wait for backend to be ready to avoid 504 errors on initial page load
@@ -245,8 +243,10 @@ k8s_yaml(
     allow_duplicates=True,
 )
 
+# TODO: determine if we can capture these objects
+#       without hardcoding names here
 # Manage k8s_yaml resources
-#   - Apply Tilt-specific grouping/labels
+# - Apply Tilt-specific grouping/labels
 k8s_resource(
     objects=[
         "selfsigned-issuer:clusterissuer",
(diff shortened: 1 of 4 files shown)


developing/scripts/setup-istio.sh
```diff
@@ -0,0 +1,65 @@
...
+  echo "  or visit: https://istio.io/latest/docs/setup/getting-started/#download"
+  exit 1
+fi
+
+# Check if Istio is already installed by looking for its CRDs
+if kubectl get crd virtualservices.networking.istio.io >/dev/null 2>&1; then
+  echo "Istio is already installed"
+else
+  echo "Installing Istio with default profile..."
+  "${ISTIOCTL}" install --set profile=default -y
```
christian-heusel: It seems like this version of istio is already EOL again:
```
Setting up Istio...
Installing Istio with default profile..
[...]
WARNING: Istio 1.26.0 may be out of support (EOL) already: see https://istio.io/latest/docs/releases/supported-releases/ for supported releases
```

Currently supported would be 1.27-1.29, manifests is using 1.28 (https://github.com/kubeflow/manifests/pull/3276)

andyatmiami: will want to talk with @thesuperzapper to confirm desires here - but aligning to `kubeflow/manifests` seems sensible/reasonable.

maybe we address this as a +1 PR that updates the Istio version consistently across the various places it is referenced in `kubeflow/notebooks` 🤔

developing/Tiltfile
```diff
@@ -45,37 +45,13 @@ docker_build(
...
-                        env[i] = {"name": "USE_ISTIO", "value": "false"}
-                        use_istio_found = True
-                        break
-                if not use_istio_found:
-                    env.append({"name": "USE_ISTIO", "value": "false"})
-
-# Apply modified manifests
-k8s_yaml(encode_yaml_stream(controller_objects))
-
-# Port-Forward controller service for local access
```
thesuperzapper: Lets leave the comment.

developing/Tiltfile
```diff
@@ -214,13 +189,28 @@ if enable_frontend:
...
+# Istio Ingress Gateway
+# ============================================================================
+# Port-forward the Istio ingress gateway for local access to production-like routing
+# This provides access through the gateway VirtualServices:
+#   - https://localhost:8443/workspaces/     -> frontend
+#   - https://localhost:8443/workspaces/api/ -> backend API
+
+local_resource(
+    "istio-ingressgateway",
+    serve_cmd="kubectl port-forward -n istio-system svc/istio-ingressgateway 8443:443",
```
thesuperzapper: I wonder if we can use the standalone https://docs.tilt.dev/api.html#api.port_forward diretive here?

christian-heusel (approved):

Issue #836, opened 2026-01-15
[TASK] Add persistent pre-GA warning banner with version info


jenny-s51:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

Following the community discussion, we should add a persistent banner throughout application to inform users that Kubeflow Notebooks v2 is pre-GA. It should display a clear alert message with a link to feature maturity status / version and build information (see https://start.kubesmarts.org/dev/)

### Acceptance Criteria

- [ ] A persistent warning banner is displayed at the top of all pages (PF's [banner](https://www.patternfly.org/components/banner) component is a good fit for this)

PR #926, opened 2026-02-23
docs: Document current feature status of notebooks v2

christian-heusel:
> [!NOTE]
> This change is a WIP and the feature status displayed in the table is not yet accurate. 

The idea of this change is to have a source of truth where users can be pointed to in order to understand the current state of the project. We also want to link this state in the pre-GA banner (see #836 for more details on the banner).

The structure of the table is roughly inspired by https://github.com/kubeflow/sdk?tab=readme-ov-file#supported-kubeflow-projects

---

Feedback would be appreciated in general but especially on the following points:
- Is the current granularity too fine? Do we need to focus on higher-level / user-visible features?
- What is the current status on some of those features?
- Should we also add a link to documentation or details of each feature?
- Is the structure in a table good for the information that we want to convey or should we break it up into a list?

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -1,0 +1,28 @@
+# Current status of Kubeflow Notebooks v2
+
+> [!WARNING]
+>
+> Kubeflow Notebooks v2 is __not yet released__ and currently in an alpha stage.
+> We are actively developing the first stable release and will share updates soon.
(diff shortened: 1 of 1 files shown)


christian-heusel:
Superseeded by https://github.com/kubeflow/notebooks/wiki/Kubeflow-Notebooks-v2

PR #940, opened 2026-02-26
feat: add display name support for workspaces

andyatmiami:
ℹ️ _NO GH ISSUE_

Add `displayName` as an optional field on `WorkspaceSpec`, following
the pattern established by WorkspaceKind. This allows users to set a
human-readable name for their workspaces in the UI.

Changes:
- Add `DisplayName` field to `WorkspaceSpec` CRD type (max 128 chars)
- Update backend models (`WorkspaceListItem`, `WorkspaceCreate`,
  `WorkspaceUpdate`) to read/write from the new spec field
- Mark `displayName` as optional in the API contract (`omitempty`)
- Add handler test coverage for display name round-trip
- Regenerate CRD manifests and Swagger docs

### DIFF
--- a/workspaces/controller/internal/controller/suite_test.go
+++ b/workspaces/controller/internal/controller/suite_test.go
@@ -169,8 +169,9 @@ func NewExampleWorkspace1(name string, namespace string, workspaceKind string) *
 			Namespace: namespace,
 		},
 		Spec: kubefloworgv1beta1.WorkspaceSpec{
-			Paused: ptr.To(false),
-			Kind:   workspaceKind,
+			Paused:      ptr.To(false),
+			DisplayName: ptr.To("Example Workspace"),
+			Kind:        workspaceKind,
 			PodTemplate: kubefloworgv1beta1.WorkspacePodTemplate{
 				PodMetadata: &kubefloworgv1beta1.WorkspacePodMetadata{
 					Labels:      nil,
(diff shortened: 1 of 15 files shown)


andyatmiami:
/unhold

`DisplayName` now defined on `Workspace` CRD in line with `WorkspaceKind`

workspaces/controller/internal/controller/suite_test.go
```diff
@@ -195,6 +196,13 @@ func NewExampleWorkspace1(name string, namespace string, workspaceKind string) *
 	}
 }
 
+// NewExampleWorkspace1WithoutDisplayName returns a Workspace with no DisplayName set (nil).
+func NewExampleWorkspace1WithoutDisplayName(name string, namespace string, workspaceKind string) *kubefloworgv1beta1.Workspace {
+	ws := NewExampleWorkspace1(name, namespace, workspaceKind)
+	ws.Spec.DisplayName = nil
+	return ws
+}
+
```
christian-heusel, line 205: Why is this function duplicated with `NewExampleWorkspaceWithoutDisplayName()` in `workspaces/controller/internal/webhook/suite_test.go`? 🤔 IMO we should only have one such helper to keep things DRY

christian-heusel (approved):
Only have a minor comment, other than that LGTM ✔️

thesuperzapper:
Thanks for this!

We now need to work on having the frontend read this field on the Workspace list, and also do the slugging to generate the Workspace CRD name from the users arbitrary text of display name.

/lgtm
/approve

Issue #943, opened 2026-02-27
[TASK] Show redirect information in Workspaces table


paulovmr:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

Improve workspace list to show redirects information related to images and pod configs used by the Workspaces.

### Acceptance Criteria

- [ ] Workspace list table shows BOTH image and pod config as columns.
- [ ] Each column should have a warning when it has a redirect
- [ ] When hovering over a redirect warning in the table this message should display:
```
Your admin has redirected `<CURRENT DISPLAY NAME>` to `<NEXT DISPLAY NAME IN CHAIN>`
Level: <REDIRECT LEVEL>
Message: <REDIRECT MESSAGE>
---
Your admin has redirected `<NEXT DISPLAY NAME>` to `<NEXT + 1 DISPLAY NAME IN CHAIN>`
Level: <REDIRECT LEVEL>
Message: <REDIRECT MESSAGE>
```

PR #946, opened 2026-03-03
feat: add pod config column and redirect information to image and pod config in Workspaces table

paulovmr:
closes: https://github.com/kubeflow/notebooks/issues/943

image

image

image


christian-heusel (commented):
Nice work @paulovmr, this really give the new feature a nice highlight in the upcoming alpha! 🎉

What I noticed though is that the redirect status is not shown when creating a new workspace but only after the first refresh:

Image

Also maybe we could highlight that this is information is supplied by the cluster admin? Right now I fear that as a an ordinary user purely having the heading with "Redirect information" leaves some open questions on what that actually means 🤔 

Maybe we could have something like "**Redirect information** (supplied by your cluster admin)" and another line that indicates that the user should update to the new image? Or idk, you're probably more versed in what a potenttially good UI/UX would be to communicate this 😅

paulovmr:
Hi @christian-heusel , thanks for the feedback! For now we've thought this is enough for a first iteration, as the cluster admin can define any message they want for the redirect. 
I did have other changes suggested to this Popover as well, that we can discuss during the community meeting. In the meantime I think these changes can be worked on in a follow up task due to the time limit and amount of things to do for the alpha release.

PR #953, opened 2026-03-05
chore: align cert-manager ready check with e2e tests

christian-heusel:
Link: https://github.com/kubeflow/notebooks/blob/d76af839376385ee893ac666447a0ca1ed00e85d/workspaces/controller/test/utils/certmanager.go#L67-L73

### DIFF
--- a/developing/scripts/setup-cert-manager.sh
+++ b/developing/scripts/setup-cert-manager.sh
@@ -29,8 +29,8 @@ kubectl wait deployment.apps --for condition=Available \
   --all-namespaces \
   --timeout=120s
 
-kubectl wait endpoints \
-  --for "jsonpath=subsets[0].addresses[0].targetRef.kind=Pod" \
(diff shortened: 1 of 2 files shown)


andyatmiami (changes requested):
whoops, I think i was a poor communicator here...

we **DO** want to wait on `deployment` ..

but then to ensure we don't hit (reasonably speaking) **any** other race conditions..

- we still want to wait on `pods` as this check was originally

- we then also want to wait on `endpoints` as well
    - https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/controller/test/utils/certmanager.go#L81-L86


this is absolutely overkill - but since you were hitting some edge cases/race conditions - lets go hog wild!

thesuperzapper:
@christian-heusel thanks, this looks ok to me, I will merge now, if we missed something feel free to raise another PR.

/lgtm
/approve

Issue #125, opened 2024-11-12
Notebooks 2.0 // Controller // Name & Key Validation


thesuperzapper:
## Problem 1

We used the wrong regex for validating  some of the fields that reference Kubernetes objects, we often used `^[a-z0-9][-a-z0-9]*[a-z0-9]$`

The correct one would be either:

   - Secrets, PVCs, WorkspaceKind names, etc:
       - Regex: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
       - Reference: https://github.com/kubernetes/apimachinery/blob/kubernetes-1.31.2/pkg/util/validation/validation.go#L210

   - Namespaces:
      - Regex: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
      - Reference: https://github.com/kubernetes/apimachinery/blob/kubernetes-1.31.2/pkg/util/validation/validation.go#L183

So we need to update annotations like:

https://github.com/kubeflow/notebooks/blob/0cacff77eecd73980aa0bb4f3ca6d8233ee1d9c0/workspaces/controller/api/v1beta1/workspace_types.go#L107

Search for instances of `+kubebuilder:validation:Pattern` that match Kubernetes resources. 

## Problem 2

We forgot to specify the `namespace` (needed because WorkspaceKind is a cluster resource, and its not clear which namespace to read the ConfigMap from).

We also need to add some validation to the configmap name/key/namespace:

 - Name: 
      - Min Characters: `1`
      - Max Characters `253`
      - Regex: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
 - Namespace:
      - Min Characters: `1`
      - Max Character `63`
      - Regex: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
 - Key:
      - Min Characters: `1`
      - Max Characters: `253`
      - Regex: `^[-._a-zA-Z0-9]+$`

Here is the part of the CRD go type that we need to update:

https://github.com/kubeflow/notebooks/blob/9d02acfdd707282385221590bc953f892099a94c/workspaces/controller/api/v1beta1/workspacekind_types.go#L91-L97

Yael-F:
@thesuperzapper  Can I assign this issue to me?

christian-heusel:
[...]

PR #954, opened 2026-03-05
fix: add missing kubebuilder validation

christian-heusel:
Closes https://github.com/kubeflow/notebooks/issues/125

### DIFF
--- a/workspaces/controller/api/v1beta1/workspacekind_types.go
+++ b/workspaces/controller/api/v1beta1/workspacekind_types.go
@@ -33,7 +33,7 @@ import (
 //
 // +kubebuilder:validation:MinLength:=1
 // +kubebuilder:validation:MaxLength:=32
-// +kubebuilder:validation:Pattern:=^[a-z0-9][a-z0-9_-]*[a-z0-9]$
+// +kubebuilder:validation:Pattern:=^[a-z0-9]([a-z0-9_-]*[a-z0-9])?$
 type PortId string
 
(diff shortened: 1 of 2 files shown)


workspaces/controller/api/v1beta1/workspace_types.go
```diff
@@ -39,7 +39,7 @@ type WorkspaceSpec struct {
 	// the WorkspaceKind to use
 	// +kubebuilder:validation:MinLength:=2
 	// +kubebuilder:validation:MaxLength:=63
```
thesuperzapper: Min should be `1` max should be `253`, probably check the others too.

christian-heusel: Is there an underlying rule about the sizes? I'm currently failing to see the pattern here 😅

andyatmiami (approved):
/lgtm

thanks for these fixes @christian-heusel 

- confirmed all the resources whose validation is changing make sense to treat as DNS1123 subdomain values:
    - PVC names, Secret names, ConfigMap names, ServiceAccount names, and
   WorkspaceKind names
- confirmed regexes and minLength are aligned


the `ConfigMap` `Key` validation could probably benefit from the following:
- ` // +kubebuilder:validation:XValidation:rule="self != '.' && self != '..'",message="key must not be '.' or '..'"  `
    - i realize `XValidation` is a bit "gross" - but no better way (imho) to handle these 2 edge cases

but we can address that in a follow up PR.

Issue #844, opened 2026-01-21
[TASK] Implement Secrets Management API endpoints


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

Following the delivery of the stubbed implementation in #655, we now need to evolve the Secret API to a full, production-ready implementation. 

At a high level involves:
- implementing the actual business logic in within [`backend/internal/repositories`](https://github.com/kubeflow/notebooks/tree/8f0495d29296db9fe0eb690d2eca9172c8deb360/workspaces/backend/internal/repositories) to interact with k8s API server as defined in the architecture document referenced in #633
- moving away from any hard-coded mock data that may have been initially included
- updating (if necessary) any handler logic that may not have been included in stub to properly:
  - validate input data 
  - handle response conditions 
- improving unit tests to:
  - remove assertions on mocked data that may have been encoded 
  - add additional test cases if additional validation and or error handling is introduced
  
:warning: There should be **no changes** to the files contained within `backend/openapi` folder as a result of these changes.  Any changes introduced you feel are warranted should be raised in a separate issue and delivered in a separate PR.

### Acceptance Criteria

- [ ] Full implementation of the Logo/Icon API logic across all endpoints as defined in #633.
- [ ] API endpoints return real data rather than hard-coded values.
- [ ] All unit tests pass, covering common successful scenarios across all endpoints
- [ ] swagger.json remains consistent with the implementation

andyatmiami:
/close

addressed in https://github.com/kubeflow/notebooks/pull/955

PR #955, opened 2026-03-08
feat: implement secret mgmt apis

andyatmiami:
closes: #844

Replace stub/mock secret repository with real Kubernetes API interactions.

Business logic:
- CRUD operations using controller-runtime client against real K8s Secrets
- Label-based access control: can-update=true required for PUT/DELETE, returns 403 Forbidden when missing
- 3-way update semantics: set (base64 provided), preserve (empty {}), or delete (key omitted) individual secret keys
- Cross-reference resolution: list response includes workspaces that mount each secret via buildSecretMountMap()
- Created secrets automatically labeled with can-mount=true and can-update=true
- Audit annotations (created-by, updated-by) set from authenticated user
- Secret values are never returned in any API response

Secret listing:
- List endpoint uses PartialObjectMetadataList via the normal cached client so only ObjectMeta is held in memory, not secret data values
- Get/Create/Update/Delete use direct API calls for full Secret objects

Validation:
- Base64 values validated at the Validate() layer, returning 422 before reaching apply functions
- ValidateFieldIsConfigMapKey refactored to report individual failures with "invalid key name:" prefix
- New ValidateFieldIsSecretBase64Value helper validates base64 encoding
- SecretData.Validate() validates both keys and values with proper field paths

Handler fixes:
- ErrSecretNotCanUpdate / ErrPVCNotCanUpdate → 403 Forbidden (not 400)
- io.ReadAll error after max-bytes check → 500 (not 400)
- Removed dead ErrSecretBase64Invalid sentinel and handler checks
- Fix Swagger annotations for DELETE endpoint (@Accept/@Produce json, description)

Tests:
- Rewrite secrets_handler_test.go with test cases covering CRUD, label-based access control (403), cross-references, validation (422 for invalid base64), empty lists, and K8s
state verification
- Add PVC delete 403 test for ErrPVCNotCanUpdate
- Add ValidateFieldIsConfigMapKey and ValidateFieldIsSecretBase64Value unit tests

### DIFF
--- a/workspaces/backend/api/app.go
+++ b/workspaces/backend/api/app.go
@@ -44,8 +44,10 @@ type App struct {
 	RequestAuthZ         authorizer.Authorizer
 }
 
-// NewApp creates a new instance of the app
-func NewApp(cfg *config.EnvConfig, logger *slog.Logger, cl client.Client, configMapClient client.Client, scheme *runtime.Scheme, reqAuthN authenticator.Request, reqAuthZ authorizer.Authorizer) (*App, error) {
+// NewApp creates a new instance of the app.
+// configMapClient is a label-filtered cached client for image-source ConfigMaps.
+// secretMetadataClient is a metadata-only cached client for listing Secrets without caching sensitive data.
+func NewApp(cfg *config.EnvConfig, logger *slog.Logger, cl client.Client, configMapClient client.Client, secretMetadataClient client.Client, scheme *runtime.Scheme, reqAuthN authenticator.Request, reqAuthZ authorizer.Authorizer) (*App, error) {
 
 	// TODO: log the configuration on startup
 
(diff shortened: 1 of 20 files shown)


thesuperzapper (changes requested):
[mathew__1__Signed-off-by__Mathew_Wicks__5735406+thesuperzapper@users_noreply_github_com_.patch](https://github.com/user-attachments/files/29868162/mathew__1__Signed-off-by__Mathew_Wicks__5735406%2Bthesuperzapper%40users_noreply_github_com_.patch)

thesuperzapper:
@andyatmiami after a lot longer than I think anyone expected, this is great to finally have merged!

So many patterns for the backend were created between when this was originally proposed and others still were developed as part of this PR. Either way, the backend is now in a great state and has clear patterns to extend even for very complex validation.

/lgtm
/approve

PR #958, opened 2026-03-10
chore: enable auth in the production manifests

christian-heusel:
For the production manifests:

    $ kustomize build workspaces/backend/manifests/kustomize/overlays/istio/ | grep -iA1 DISABLE_AUTH
    - name: DISABLE_AUTH
      value: "false"

For the tilt setup:

    $ kubectl get deployments -n kubeflow-workspaces workspaces-backend -o yaml | grep -iA1 DISABLE_AUTH
    - name: DISABLE_AUTH
      value: "true"


thesuperzapper:
Thanks for this!

/approve
/lgtm

PR #969, opened 2026-03-17
chore: fix releasing script

thesuperzapper:
This PR fixes the releasing script so that it actually updates all the image tags, namely the example notebook-servers under `components/crud-web-apps/jupyter/manifests/base/configs/spawner_ui_config.yaml`.

It also fixes a critical typo which would have caused the images to still be pulled from `ghcr.io/kubeflow/kubeflow/xxxx` rather than `ghcr.io/kubeflow/notebooks/xxxx` (for the non notebook-servers images, as we left those on `ghcr.io/kubeflow/kubeflow/notebook-servers/xxxx` even though they now push from this repo.

Other changes:
- remove "run" prefix from unit test gha names (easier to find in the long list as it ensures grouping)
- fixes test failures by pinning envtest/controller-gen in all controller's makefiles
- uses retries for the flakey `go mod downloads` of `github.com/hashicorp` in Dockerfiles

### DIFF
--- a/components/notebook-controller/Dockerfile
+++ b/components/notebook-controller/Dockerfile
@@ -19,7 +19,19 @@ RUN cd /workspace/notebook-controller && go mod download
 # Force download and extract all HashiCorp modules so they exist under /go/pkg/mod
 RUN set -e; cd /workspace/notebook-controller; \
   for m in $(go list -m all | awk '/^github.com\/hashicorp\// {print $1 "@" $2}'); do \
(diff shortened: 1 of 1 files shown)


andyatmiami (approved):
/lgtm

- great alignment across `Makefile` now
- excellent catch on the `kubeflow/kubeflow` references... i'm surprised we let this glaring oversight _fester_ here - but thanks for update

Issue #691, opened 2025-10-29
[TASK] Stub Compatibility Selectors API endpoints on backend


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

We want a _stubbed implementation_ of the Compatibility Selectors API as outlined in #683 .

Definition of a "stubbed implementation":
- Changes are expected to be delivered under following folders:
    - `workspaces/backend/api`
    - `workspaces/backend/internal/models`
    - `workspaces/backend/openapi`
        - ℹ️ files under this directory are auto-generated from `swag` `Makefile` command
 - Data returned by the API should be hard-coded and provide a minimally viable experience for the `frontend` to be able to exercise the API!
 - Changes should **NOT** be provided in `workspaces/backend/internal/repositories`
     - These changes would be included when working on the **real** implementation of the endpoints
 - Basic unit testing should still be provided for the changes
     - These tests would obviously require asserting on the hard-coded "mock" responses and will require rework then when we implement these endpoints **for real**

💡 The API proposal includes the **existing** `GET /workspacekinds` (list) endpoint.  "Stubbing" in this case would be to simply ensure the additional metadata we want to include on the `data[]` elements returned are present (and can be hard-coded).

### Acceptance Criteria

- [ ]`swagger.json` includes all the appropriate metadata to expose APIs as outlined in #683

thesuperzapper:
Done in https://github.com/kubeflow/notebooks/pull/972

/close

PR #972, opened 2026-03-18
feat: add WSK listvalues api to backend

akrem-chabchoub:
Implements https://github.com/kubeflow/notebooks/issues/691

New **POST**  `.../workspacekinds/{name}/podtemplate/options/listvalues` that returns **imageConfig** and **podConfig** options for that workspace kind.

Each value includes **ruleEffects** — for now always `uiHide: false` until real selector logic exists.

Work on top of: https://github.com/kubeflow/notebooks/pull/875 (addressing review comments)


andyatmiami (changes requested):
@akrem-chabchoub 

Thanks for raising this PR - excited about the opportunities this will open up in the codebase in the future!

Got a few things I would like for you to take a look at to _harden_ this good foundation you have laid.  Let me know if you have any questions.

Once you get these changes in - I'll focus my subsequent review on testing within our `Tilt` environment to assert appropriate API behavior 💯

workspaces/backend/internal/models/workspacekinds/types.go
```diff
@@ -109,3 +115,89 @@ const (
...
+type ListValuesRequest struct {
+	Context *ListValuesContext `json:"context,omitempty"`
+}
+
+// Validate validates the listvalues request data by delegating to Context when present.
+func (d *ListValuesRequest) Validate(prefix *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	if d == nil {
+		return errs
+	}
```
andyatmiami: I don't think we need to have a `nil` guard here as this is the top level `Data` object which we check for nil-ness directly within the handler.

additionally, our other `Validate` functions don't do a nil-ness check on the receiver type.

workspaces/backend/internal/models/workspacekinds/types.go
```diff
@@ -109,3 +115,89 @@ const (
...
 	RedirectMessageLevelDanger  RedirectMessageLevel = "Danger"
 )
+
+type ListValuesRequest struct {
+	Context *ListValuesContext `json:"context,omitempty"`
+}
+
+// Validate validates the listvalues request data by delegating to Context when present.
+func (d *ListValuesRequest) Validate(prefix *field.Path) field.ErrorList {
+	var errs field.ErrorList
```
andyatmiami: To be consistent with other `Validate` functions - I think we want to return a type of `[]*field.Error` - which then also affects the implementation in the body..

workspaces/backend/internal/models/workspacekinds/types.go
```diff
@@ -109,3 +115,89 @@ const (
...
+
+type ListValuesContext struct {
+	// TODO: Namespace is reserved for future use, no behavior implemented yet.
+	Namespace   *ContextNamespace   `json:"namespace,omitempty"`
+	PodConfig   *ContextPodConfig   `json:"podConfig,omitempty"`
+	ImageConfig *ContextImageConfig `json:"imageConfig,omitempty"`
+}
+
+// Validate validates the context fields when present.
+func (c *ListValuesContext) Validate(prefix *field.Path) field.ErrorList {
```
andyatmiami: To be consistent with other `Validate` functions - I think we want to return a type of `[]*field.Error` - which then also affects the implementation in the body..

workspaces/backend/internal/models/workspacekinds/types.go
```diff
@@ -109,3 +115,89 @@ const (
...
+	var errs field.ErrorList
+	if c == nil {
+		return errs
+	}
```
andyatmiami: (just commenting "out loud" - no change requested here)

This `nil` check is OK if we are wishing to allow users to call this endpoint without the `context:{}` object... so it seems fine to leave to me

andyatmiami (changes requested):
PR is looking really good... just a few fairly trivial comments to keep alignment with existing code practices...

Exciting to see the API really start to take shape!

Image

workspaces/backend/internal/models/workspacekinds/types.go
akrem-chabchoub: Yes, my bad, I updated it

workspaces/backend/internal/models/workspacekinds/types.go
akrem-chabchoub: Thanks, I’ll keep it as-is.

akrem-chabchoub:
Thanks @andyatmiami for the review 💯 
I addressed all of your comments 🚀

thesuperzapper:
@akrem-chabchoub as discussed on call, I have rebased and pushed a commit https://github.com/kubeflow/notebooks/pull/972/commits/4ad5e3d6756cc31b2ec22f7e66104b2315f7fdce on top of the branch to fix the issues we discussed.

andyatmiami (approved):
/lgtm
/approve

thanks for this work @akrem-chabchoub and appreciate the thorough review to align/re-enforce conventions @thesuperzapper 

agree with the changes delivered and also tested this locally in kind using `Bruno` API client to verify positive and negative scenarios.

all tests check out - lets get this merged! 💯

PR #980, opened 2026-03-19
refactor: add new lint for omitempty on slice fields and apply it treewide 

christian-heusel:
- **feat: add jsonomitemptylint custom linter for Go slice fields**
- **fix: Add omitempty to slice fields in Go types**

The idea on this was brought up by @thesuperzapper regarding the discussion around https://github.com/kubeflow/notebooks/pull/976.

Alternatively we could aim to use the `nil` as empty array change from the upcoming from the jsonv2 proposal in golang:

> A nil Go slice is marshaled as an empty JSON array ([more discussion](https://github.com/golang/go/discussions/63397#discussioncomment-7201222)).

https://github.com/golang/go/discussions/63397 / https://pkg.go.dev/encoding/json/v2

---

This will most likely break the frontend and require adjustments in the tests, so this is not to be merged yet.

Would be nice to get some intput on this from @ederign for the frontend side of things as per @andyatmiami's tip.

### DIFF
--- a/tools/linters/cmd/jsonomitemptylint/main.go
+++ b/tools/linters/cmd/jsonomitemptylint/main.go
@@ -1,0 +1,133 @@
+// Copyright 2024.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"go/token"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/kubeflow/notebooks/tools/linters/jsonomitempty"
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/passes/inspect"
+	"golang.org/x/tools/go/packages"
+)
+
+func main() {
+	flag.Parse()
+	patterns := flag.Args()
+	if len(patterns) == 0 {
+		patterns = []string{"."}
+	}
+
+	cwd, err := os.Getwd()
+	if err != nil {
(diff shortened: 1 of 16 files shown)


thaorell:
Thanks for the work @christian-heusel. After discussing with @andyatmiami, we have estimated that the impacts would remain minimal, as long as swagger.json is consistent with the return type in the backend. The only work we need to do in the frontend is: 
1. Update the auto-generated client
2. Apply detection for "false"-y values and render the components

tools/linters/cmd/jsonomitemptylint/main.go
```diff
@@ -0,0 +1,133 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/kubeflow/notebooks/tools/linters/jsonomitempty"
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/passes/inspect"
+	"golang.org/x/tools/go/packages"
+)
+
+func main() {
```
thesuperzapper, line 31: This boilerplate code feels like something which might exist upstream in the `analysis` packages are depending on, or if there is a pre-made package we can depend on that allows us to only define our lint rules in this repo.

tools/linters/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/kubeflow/notebooks/tools/linters
+
+go 1.22.0
```
thesuperzapper, line 3: This should be moved under the `workspaces/dashboard` folder, and as a result will not need its own go mod?

tools/linters/cmd/jsonomitemptylint/main.go
```diff
@@ -0,0 +1,133 @@
...
+			exitCode = 1
+		}
+	}
+
+	os.Exit(exitCode)
+}
+
+// isUnderTestDir reports whether the given (relative) file path has a
+// path component named "test".
+func isUnderTestDir(path string) bool {
```
thesuperzapper, line 126: This is surprising to see, in a similar sense to the discussion above about if we can depend on a package which provides an interface to inject our own analysis code, and understand how to ignore specific file paths (similar to how we already configure [`.golangci.yml`](https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/.golangci.yml) to ignore test folders).

thesuperzapper (changes requested):
I think we can take your `analyzers.go` and just convert it into a new custom linter for `golangci-lint`, I am not 100% if we can add custom linters, but I assume its possible.

https://golangci-lint.run/docs/contributing/new-linters/

PR #986, opened 2026-03-24
chore: fix the line wrapping in ruamel.yaml

christian-heusel:
So far the library has caused line wraps when the version used is
(some version) greater than 18.11, which annoyed Mathew when doing
releases. Fix this by setting a really high value for maximum line
length.

This causes diffs like the following:

```diff
-    value: ghcr.io/kubeflow/kubeflow/notebook-servers/codeserver-python:latest
+    value:
+      ghcr.io/kubeflow/kubeflow/notebook-servers/codeserver-python:v1.11.0-rc.
```


thesuperzapper:
Thanks for making my life better.

/approve
/lgtm

PR #987, opened 2026-03-24
ci: enforce only publishing image version tags from release branches

christian-heusel:
This way the docker images are not accidentally updated from the `main` branch or the other development branches.

### DIFF
--- a/.github/workflows/example_notebook_servers_publish_TEMPLATE.yaml
+++ b/.github/workflows/example_notebook_servers_publish_TEMPLATE.yaml
@@ -49,7 +49,8 @@ jobs:
           ARCH: ${{ inputs.build_arch }}
           IMAGE_FOLDERS: ${{ inputs.image_folders }}
           SHOULD_TAG_LATEST: ${{ github.ref == 'refs/heads/notebooks-v1' }}
-          SHOULD_TAG_VERSION: ${{ steps.filter.outputs.version == 'true' }}
+          # We only publish version-tagged images from release branches
+          SHOULD_TAG_VERSION: ${{ steps.filter.outputs.version == 'true' && startsWith(github.ref, 'refs/heads/v') }}
         run: |
           if [[ "$SHOULD_TAG_LATEST" = "true" ]]; then
             export ALSO_TAG_LATEST=1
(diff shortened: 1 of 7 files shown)


thesuperzapper (changes requested):
There are a few GHA other files which push images, we should also apply the fix to them (I think there are 8 in total).

christian-heusel:
> There are a few GHA other files which push images, we should also apply the fix to them (I think there are 8 in total).

Any more missing? I only found 7 🤔 

```
$ rg -l "docker-build-push" .github
.github/workflows/vwa_docker_publish.yaml
.github/workflows/twa_docker_publish.yaml
.github/workflows/jwa_docker_publish.yaml
.github/workflows/tb_controller_docker_publish.yaml
.github/workflows/example_notebook_servers_publish_TEMPLATE.yaml
.github/workflows/nb_controller_docker_publish.yaml
.github/workflows/pvcviewer_controller_docker_publish.yaml

$ rg -l "docker-build-push" .github | wc -l
7
```

.github/workflows/example_notebook_servers_publish_TEMPLATE.yaml
```diff
@@ -49,7 +49,8 @@ jobs:
           ARCH: ${{ inputs.build_arch }}
           IMAGE_FOLDERS: ${{ inputs.image_folders }}
           SHOULD_TAG_LATEST: ${{ github.ref == 'refs/heads/notebooks-v1' }}
-          SHOULD_TAG_VERSION: ${{ steps.filter.outputs.version == 'true' }}
+          # We only publish version-tagged images from release branches or tags
+          SHOULD_TAG_VERSION: ${{ steps.filter.outputs.version == 'true' && (startsWith(github.ref, 'refs/heads/v') || startsWith(github.ref, 'refs/tags/v')) }}
```
thesuperzapper: Since we dont publish images from tags, only from commits on relase branches which change the `VERSION` file, we should probably only match on startswith `refs/heads/v`.

```suggestion
          # We only publish version-tagged images from release branches or tags
          SHOULD_TAG_VERSION: ${{ steps.filter.outputs.version == 'true' && startsWith(github.ref, 'refs/heads/v') }}
```

thesuperzapper (changes requested):

thesuperzapper:
We need to remember to chery pick this into all release branches (which there is only 1 right now with `v1.11-branch` so should be easy).

/lgtm
/approve

PR #1004, opened 2026-04-01
chore: disable gaudi images

christian-heusel:
- **chore: Drop the gaudi image due to build failures**


thesuperzapper:
superceeded by https://github.com/kubeflow/notebooks/pull/1101

which actually removes the image code rather than just stopping builds.

Issue #978, opened 2026-03-19
[TASK] WorkspaceForm Summary Improvements


thaorell:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

 The workspace create/edit form renders a Summary as a right-side Drawer panel (WorkspaceFormSummaryPanel inside DrawerPanelContent in WorkspaceForm.tsx). This panel shows a stacked set of cards summarizing the user's selections across all four steps (Workspace Kind, Image, Pod Config, Properties).

 As the number of selections grows — labels, descriptions, redirect popovers, volume/secret lists in Properties — the narrow drawer becomes cramped and requires scrolling to review everything before submitting. Moving the summary to its own full-width wizard step gives it more real estate and sets the stage for future enhancements (e.g., a YAML preview).

Additionally, in edit mode, SummaryDiffSection stacks "NEW" above "CURRENT" vertically with a Divider. A side-by-side Split layout would make it easier to compare original vs. changed values at a glance.

Proposal: 

  1. Add a "Summary" step to the WorkspaceForm wizard (at least for Create; optionally for Edit). The step appears after "Properties" and renders the summary content at full page width instead of in the drawer. The existing WorkspaceFormSummaryPanel component can be reused as the step body — the main work is wiring it into WorkspaceForm.tsx as a step.
  2. Keep summary cards clickable. The cards already have isClickable and onClick={() => onNavigateToStep(step)}. Verify this works correctly when Summary is its own step (clicking a card should navigate backward in the wizard).
  3. Use Split/SplitItem in SummaryDiffSection to place "CURRENT" and "NEW" side by side instead of stacked vertically, making comparisons easier to read.

### Acceptance Criteria

- [ ] Summary step in the wizard. The progress stepper shows a "Summary" step after "Properties". The summary content renders at full page width.
- [ ] Clickable cards navigate back. Clicking any summary card navigates the user back to that wizard step.
- [ ] Side-by-side diff in edit mode. SummaryDiffSection uses Split/SplitItem to render "CURRENT" on the left and "NEW" on the right, instead of stacking vertically.
- [ ] Drawer updated. The right-side drawer is hidden or removed on the Summary step. Decide whether to keep it as a live preview on non-summary steps or remove it entirely.
- [ ] Submit button on Summary step. "Create" / "Save" replaces "Next" on the Summary step. "Previous" goes back to "Properties".
[...]

PR #1012, opened 2026-04-03
feat: Improve WorkspaceForm Summary

thaorell:
closes: #978

- Added Summary Step
Screenshot 2026-04-03 at 11 25 21 AM
- For Pod Config and ImageConfig, moved the Diff section to Split for better comparison views

Before:
Screenshot 2026-04-03 at 11 40 09 AM
After:
Screenshot 2026-04-03 at 11 39 38 AM

- For Properties card: removed name on Edit mode, wrap volumes in labels

Before:

Screenshot 2026-04-03 at 11 31 47 AM

After:

Screenshot 2026-04-03 at 11 32 19 AM

- Use LabelGroupWithTooltip and remove SummaryLabels, because SummaryLabels records click event on a clickable Card component, resulting in confusing UX. 
Screenshot 2026-04-03 at 11 26 45 AM

- Add tests

### DIFF
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/SummaryPropertiesSection.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/SummaryPropertiesSection.tsx
@@ -18,20 +18,16 @@ export const SummaryPropertiesSection: FC<SummaryPropertiesSectionProps> = ({
   mode,
 }) => {
   const isEditMode = mode === 'update';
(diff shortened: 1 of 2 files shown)


caponetto (commented):
Left two minor comments inline and noticed the following while testing locally:

In the `Properties` step, the summary card does not show up while the workspace name is not filled, even when other properties are filled in. Is it expected?

Image

ederign (approved):

PR #1013, opened 2026-04-03
fix: forward incoming auth headers in webpack dev proxy

mturley:
## Description

The webpack dev server captures the kubeconfig token at startup and sets it as a static proxy header via `headers: getProxyHeaders()`. This overwrites any dynamically forwarded auth headers from a host backend proxy, breaking dev impersonation (e.g. `DEV_IMPERSONATE_USER`/`DEV_IMPERSONATE_PASSWORD`).

This PR replaces the static `headers` approach with an `onProxyReq` callback that dynamically forwards the `Authorization` header from each incoming request when present, falling back to the startup-captured token for standalone dev mode.

The same fix was applied to the model-registry UI in https://github.com/kubeflow/model-registry/pull/2544.

## Changes

- Split `getProxyHeaders()` into:
  - `getKubeconfigToken()` — captures kubeconfig token at startup (fallback)
  - `getProxyHeaders()` — returns only static headers (`kubeflow-userid` for `internal` auth)
  - `onProxyReq` — dynamically forwards `Authorization` and `x-forwarded-access-token` headers per request
- Proxy config now uses both `headers: getProxyHeaders()` (static) and `onProxyReq` (dynamic)

## How Has This Been Tested?

Tested manually in the model-registry module (same fix pattern) against an ODH cluster:
1. Started the host backend with `DEV_IMPERSONATE_USER` and `DEV_IMPERSONATE_PASSWORD` set
2. Verified the bug: with static headers, the impersonated user's token was overwritten by the startup token
3. Verified the fix: with `onProxyReq`, the incoming auth header from the host proxy is forwarded correctly

### DIFF
--- a/workspaces/frontend/config/webpack.dev.js
+++ b/workspaces/frontend/config/webpack.dev.js
@@ -28,36 +28,56 @@ const DEPLOYMENT_MODE = process.env._DEPLOYMENT_MODE;
 const AUTH_METHOD = process.env._AUTH_METHOD;
 const BASE_PATH = DEPLOYMENT_MODE === 'kubeflow' ? '/workspaces' : PUBLIC_PATH;
 
-// Function to generate headers based on deployment mode
+// Get the kubeconfig token at startup as a fallback for standalone dev mode.
+const getKubeconfigToken = () => {
+  try {
+    const token = execSync(
+      "kubectl config view --raw --minify --flatten -o jsonpath='{.users[].user.token}'",
+    )
+      .toString()
+      .trim();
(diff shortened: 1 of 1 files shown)


thaorell (approved):
/lgtm

I have tested and the changes work accordingly, @caponetto can you (optionally) offer a secondary review on this?

caponetto:
Thanks @mturley!

/lgtm

mturley:
@ederign the MR PR is merged now, should we unhold this one? Not sure if I missed discussions about the approach. cc @lucferbux

PR #1014, opened 2026-04-04
feat: setup Dependabot for `kubeflow/notebooks`

### LINKED ISSUE #818: [TASK] Add Dependabot for automated dependency updates
### Certification

### Description

See https://github.com/kubeflow/model-registry/blob/main/.github/dependabot.yml for reference. 

### Acceptance Criteria

- [ ] Create `.github/dependabot.yml`
- [ ] Group `mod-arch-*` packages together in a single PR to keep these packages up to date
- [ ] Should also monitor: Go modules (backend & controller), npm packages (frontend), Docker images, GitHub Actions

christian-heusel:
Main things to note about this setup are the following ones:

- We target three different branches (`main`, `notebooks-v1`,  `notebooks-v2`) via the `target-branch` directive
- The `cooldown` option delays updates a little to as a prevention mechanism for supply-chain attacks, similar to the `min-release-age` in NPM (see #1003)
- We try to group dependencies where sensible, however this will most likely need a bit of iteration
- Only allow non-breaking upgrades for things like the k8s packges and angular
- Part of the configuration for `notebooks-v1` is disabled to keep a good signal-to-noise ratio when starting out

Fixes https://github.com/kubeflow/notebooks/issues/818

cc @jenny-s51 who did the initial version of this in #819 
cc @andyatmiami who also gave a lot of input during development

---

I have already used this in a separate repository to evaluate it, checkout https://github.com/christian-heusel/notebooks-playground if you want to have a look!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,165 @@
+version: 2
+
+updates:
+
+  ########################
+  # --> notebooks-v2 <-- #
+  ########################
+
+  # - package-ecosystem: "gomod"
+  #   target-branch: "notebooks-v2"
+  #   open-pull-requests-limit: 5
+  #   cooldown:
+  #     default-days: 7
+  #   directories:
+  #     - "/workspaces/backend"
+  #     - "/workspaces/controller"
+  #   schedule:
+  #     interval: "daily"
+  #   commit-message:
+  #     prefix: "chore"
+  #   ignore:
+  #     # k8s upgrades are expected to be breaking changes and should be handled
+  #     # manually
+  #     - dependency-name: "k8s.io/*"
+  #       update-types:
+  #         - "version-update:semver-major"
+  #         - "version-update:semver-minor"
+  #     - dependency-name: "sigs.k8s.io/*"
+  #       update-types:
+  #         - "version-update:semver-major"
+  #         - "version-update:semver-minor"
+  #   groups:
+  #     k8s-packages:
+  #       patterns:
+  #         - "k8s.io/*"
+  #         - "sigs.k8s.io/*"
+  #     go-test-packages:
+  #       patterns:
+  #         - "github.com/onsi/*"
+
+  - package-ecosystem: "npm"
+    target-branch: "notebooks-v2"
+    open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
+    directory: "/workspaces/frontend"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "chore"
+    # Group mod-arch-* packages together for easier review
+    groups:
+      mod-arch-packages:
+        patterns:
+          - "mod-arch-*"
+
+  - package-ecosystem: "github-actions"
+    target-branch: "notebooks-v2"
+    open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
+    directory: "/"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "chore"
+    groups:
+      docker-actions:
+        patterns:
+          - "docker/*"
+
+  ########################
+  # --> notebooks-v1 <-- #
+  ########################
+
+  # - package-ecosystem: "gomod"
+  #   target-branch: "notebooks-v1"
+  #   open-pull-requests-limit: 5
+  #   cooldown:
+  #     default-days: 7
+  #   # find . -name "go.mod" | xargs -i dirname {}
+  #   directories:
+  #     - "/components/notebook-controller"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,211 @@
...
+      prefix: "chore"
+    ignore:
+      # major angular upgrades are expected to be breaking and should be
+      # handled manually
+      - dependency-name: "@angular/*"
+        update-types:
+          - "version-update:semver-major"
+      - dependency-name: "@angular-eslint/*"
+        update-types:
+          - "version-update:semver-major"
```
andyatmiami: @christian-heusel - I wonder if this is a copy/paste error from `notebooks-v1` and/or `dashboard`? I don't see evidence of `frontend` using `angular` 🤔 

also related here:
- https://github.com/kubeflow/notebooks/pull/1014/changes#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R65-R68

I could see a case to be made w.r.t. `react` and/or `patternfly` to be handled similarly to what we are doing with `angular` fwiw..

.github/dependabot.yml
```diff
@@ -0,0 +1,211 @@
...
+    groups:
+      mod-arch-packages:
+        patterns:
+          - "mod-arch-*"
+      angular-packages:
+        patterns:
+          - "@angular/*"
+          - "@angular-eslint/*"
+
+  - package-ecosystem: "docker"
```
andyatmiami: I don't see any `docker` PRs raised in your [playground](https://github.com/christian-heusel/notebooks-playground/pulls?q=is%3Apr+is%3Aopen+) - so its not clear if this has nothing to update or its not working...

Quickly scanning this link: https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#docker - its not obvious to me that we are meeting all pre-reqs to make this check useful....

here is some information according to Claude (Opus 4.6) that makes me feel like its not worth enabling this `docker` ecosystem behavior now - and we can talk about if we want to institute **other changes** that could potentially make this more meaningful...

Image

Image

Image

andyatmiami (changes requested):
thanks for driving forward on this @christian-heusel 

just a couple comments to address and/or respond to:

- `angular` in `frontend` for `notebooks-v2` seems like a copy/paste error
- i'm not convinced the `docker` section really provides much value right now

.github/dependabot.yml
christian-heusel: Yes this is a copy paste error, therefore I dropped it 😊

As soon as we merge this I guess the frontend people can decide for themselves what they would like to group or alternatively let me know so I can take care of it 🤗

.github/dependabot.yml
christian-heusel: Alright, dropped that one aswell 👍🏻 we can always revisit it later on

.github/dependabot.yml
```diff
@@ -0,0 +1,165 @@
...
+    directory: "/workspaces/frontend"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "chore"
+    # Group mod-arch-* packages together for easier review
+    groups:
+      mod-arch-packages:
+        patterns:
+          - "mod-arch-*"
```
andyatmiami, line 55: @caponetto | @paulovmr | @jenny-s51 

do y'all have any thoughts/opinions on how you might light the dependabot settings for `frontend` to be structured?

given the investment y'all have made in testing - seems more viable to expect to be able to have good coverage to enable validating dependabot PRs - but curious of your thoughts on the following:
- are you comfortable turning this on for `frontend` so we can start to "see what unfolds"? I'm sure iteration will be required.. but any reason we shouldn't start down this path now?
- any thoughts/preferences on other group `patterns` you might like to see defined?

thanks!

andyatmiami (commented):
Thanks for addressing feedback @christian-heusel - just wanna get some thoughts from `frontend` folks in case they have strong opinions on what this should look like _out the gate_ ...

.github/dependabot.yml
paulovmr: I like it, I was going to comment about the `open-pull-requests-limit`, but it was already used, so lgtm!

paulovmr (approved):

andyatmiami:
/lgtm

@thesuperzapper - need you to approve as this lives on `main` branch

thesuperzapper:
Thanks this is great to start automatically telling us about dependency updates, we can uncommnent the notebooks-v2 ones once we have E2E tests and are more confident in merging updates.

/approve

Issue #781, opened 2025-12-07
[TASK] [Security] Fix Vulnerabilities in Tensorboard Controller (v1.11 Release)


liavweiss:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

As part of the effort to release v1.11, we are addressing all high and critical security vulnerabilities identified in our recent scan. This ensures the security and stability of the upcoming release.

**Component:** `components/tensorboard-controller`

**Instruction:** Update Go dependencies to resolve critical and high CVEs. Please open a separate Pull Request for each item below to isolate changes.

**Action Items:**
- [ ] PR 1: Update `golang.org/x/net` to v0.38.0  
  Reason: Fixes CVE-2023-39325, CVE-2023-45288, etc.
- [ ] PR 2: Update `golang.org/x/crypto` to v0.45.0  
  Reason: Fixes CVE-2024-45337, CVE-2023-48795, etc.
- [ ] PR 3: Update `golang.org/x/oauth2` to v0.27.0  
  Reason: Fixes CVE-2025-22868.
- [ ] PR 4: Update `google.golang.org/protobuf` to v1.33.0  
  Reason: Fixes CVE-2024-24786.
- [ ] PR 5: Update `gopkg.in/yaml.v3` to v3.0.1  
  Reason: Fixes CVE-2022-28948.

### Acceptance Criteria

- Validate that the new dependencies keep the Tensorboard CRD and controller APIs backward-compatible with existing workloads.
- Execute `notebooks/.github/workflows/tb_controller_unit_test.yaml`, `tb_controller_integration_test.yaml`, and `tb_controller_multi_arch_test.yaml` to cover the unit, integration, and multi-arch scenarios.
- In `components/tensorboard-controller`, run `go mod tidy` and `make manager` (or another equivalent build target) to prove the component builds cleanly after the updates.
- Deploy `notebooks-v1`, then exercise the Tensorboard lifecycle (creating a TensorBoard resource, confirming pod readiness, and ensuring logs are accessible) to prove the controller still works end-to-end.

andyatmiami:
/close 

all planned work was delivered

Issue #98, opened 2022-12-09
KF Notebooks: Reconcile Helper overwrites or collides with image-field changes from image policy admission plugin


shalberd:
/kind bug

**What steps did you take and what happened:**
Applied a Notebook CR in conjunction with Notebook controller. Notebook controller creates a StatefulSet from the Notebook CR, containing all the specs needed for e.g. the container, including the image-field.
With DeploymentConfigs, Deployments, Pods, StatefulSets, Openshift has a mechanism called the Image Policy admission plugin,[ image.openshift.io/ImagePolicy](https://docs.openshift.com/container-platform/4.10/architecture/admission-plug-ins.html#admission-plug-ins-default_admission-plug-ins), that resolves [imagestreams](https://itnext.io/variations-on-imagestreams-in-openshift-4-f8ee5e8be633) to image-references in sha256 digest format, in that function being a mutating type of plugin.

Here is what I did:

- For same-namespace tests, I added an imagestream reference and tag to the image-field of a Notebook CR. The statefulset contains the image-field, but reconciliation of the StatefulSet keeps overwriting the image-field, thus never resolving the actual image-location. Reconcile [documentation on basics](https://cluster-api.sigs.k8s.io/developer/providers/implementers-guide/controllers_and_reconciliation.html)

https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/controllers/notebook_controller.go#L90

Related discussion at https://github.com/openshift/openshift-apiserver/issues/339#issuecomment-1343192025

**What did you expect to happen:**

I expected the image-field of the StatefulSet to contain the resolved image reference

```
spec:
      containers:
        - name: jupyter
          image: >-
            quay.io/thoth-station/s2i-generic-data-science-notebook@sha256:3f619c61501218f03d39d97541336dee024f446e64f3a47e2bc7e62cddeb2e58
```

instead the image-field stays at 
```
          image: 's2i-generic-data-science-notebook:v0.0.5'
```

for StatefulSets created from the Notebook CR. 

**Anything else you would like to add:**
Error occuring in context of [Open Data Hub Notebooks controller](https://github.com/opendatahub-io/odh-manifests/tree/master/odh-notebook-controller) context.
Reconcilehelper should exclude the image-field from reconciliation. Currently, the whole template.spec part of StatefulSets is synced as a whole.

https://github.com/kubeflow/kubeflow/blob/master/components/common/reconcilehelper/util.go#L131

[...]

PR #1025, opened 2026-04-16
fix: sync tensorboard container image on reconcile

jeffspahr:
## Summary
- `CopyDeploymentSetFields` only synced labels, replicas, and affinity, so any change to the `TENSORBOARD_IMAGE` env var only took effect on newly created Tensorboards. Existing Deployments drifted from the controller's desired spec indefinitely — including after security patches to the tensorboard image (see #781).
- Extend the helper to copy `Spec.Template.Spec.Containers[*].Image` when it differs. Containers are matched by name to avoid any cross-container assignment if ordering ever changes in `generateDeployment`.
- Other fields in the container spec (args, ports, volumes) are intentionally left alone: they derive from the immutable `Tensorboard` CR spec, so they cannot legitimately drift post-create. Narrowing the sync to image keeps the change minimal and avoids fighting with admission plugins that mutate container fields (context: #98).

## Why this matters
Today, operators rolling out a new `TENSORBOARD_IMAGE` (for CVE patches or feature upgrades) have no automated path to update running Tensorboards. The only workarounds are deleting and recreating each `Tensorboard` CR, or manually deleting the managed Deployments so the controller recreates them from the new template. With this change, a controller rollout is enough — the existing image-sync on reconcile will roll each Deployment to the new image.

## Test plan
- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `make fmt vet` — clean
- [x] `go test -run '^TestCopyDeploymentSetFields_' ./controllers/ -v` — 3 new unit tests pass:
  - `TestCopyDeploymentSetFields_UpdatesChangedContainerImage`
  - `TestCopyDeploymentSetFields_NoUpdateWhenImageMatches`
  - `TestCopyDeploymentSetFields_MatchesContainerByName`
- [ ] Maintainers: run the existing `tb_controller_unit_test.yaml`, `tb_controller_integration_test.yaml`, and `tb_controller_multi_arch_test.yaml` workflows.
- [ ] Manual verification against `notebooks-v1`: roll the controller with a bumped `TENSORBOARD_IMAGE` and confirm existing Tensorboards' Deployments are updated and their pods roll.

## Notes
- Target branch: `notebooks-v1`.
- Scope is intentionally narrow — I noticed separately that the Deployment has no `readinessProbe` and an implicit rollout strategy, which combined with single-replica RWO-PVC backed Tensorboards means any image change incurs a brief outage per instance. Happy to open a follow-up PR for a `Recreate` strategy + readiness probe if maintainers want it; keeping it out of this change.

### DIFF
--- a/components/tensorboard-controller/README.md
+++ b/components/tensorboard-controller/README.md
@@ -59,3 +59,7 @@ If you want to enable the scheduling functionality for Tensorboard servers that 
 2. Modify the `manager.yaml` file by navigating to the `deployment.spec.template.spec` field and manually setting the value of the `RWO_PVC_SCHEDULING` env var to `"true"` in the manager container.
 
 3. Run: `make deploy IMG=YOUR_IMAGE_NAME`
+
+## IMAGE UPGRADES
+
+The Tensorboard server image is configured via the `TENSORBOARD_IMAGE` env var on the controller (see `config/base/kustomization.yaml`). When this value changes and the controller rolls, the controller reconciles the image onto every existing `Tensorboard`'s managed Deployment, not just newly created ones. This means active users will see their Tensorboard pod briefly restart and their browser tab need a refresh; the underlying event logs (on the PVC or cloud bucket referenced by `spec.logspath`) are unaffected, so no data is lost.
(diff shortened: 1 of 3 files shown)


jeffspahr:
@thesuperzapper I'd love to see this merged.  Any questions or concerns I can address?

PR #1026, opened 2026-04-17
chore: upgrade to go 1.24

andyatmiami:
- Bump go directive to 1.24.13 in backend and controller go.mod files
- Update Dockerfile builder images from golang:1.22 to golang:1.24
- Upgrade golangci-lint from v1.61.0 to v1.64.8 for Go 1.24 compatibility
- Fix gocritic hugeParam lint: pass controller.Options by pointer in SetupWithManager signatures

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>


thesuperzapper:
Since this is needed to develop locally on macos 26 due to https://github.com/golang/go/issues/68678 we should merge asap.

/lgtm
/approve

PR #1000, opened 2026-03-31
fix: add back the `v` prefix to image tags

christian-heusel:
---

# Old Changes

The `ws-publish.yml` GHA pushes tags without "v" prefix if `tag_with_semver` is set, which is the case for all release workflows. Without this we're running into `ImagePullBackOff` issues in the release manifests.

---

`python releasing/update-manifests-images.py`
```diff
diff --git a/releasing/version/VERSION b/releasing/version/VERSION
index 9001211..ee6c00d 100644
--- a/releasing/version/VERSION
+++ b/releasing/version/VERSION
@@ -1 +1 @@
-dev
\ No newline at end of file
+v2.0.0-alpha.0
diff --git a/workspaces/backend/manifests/kustomize/base/kustomization.yaml b/workspaces/backend/manifests/kustomize/base/kustomization.yaml
index 793103a..9a8f6e0 100644
--- a/workspaces/backend/manifests/kustomize/base/kustomization.yaml
+++ b/workspaces/backend/manifests/kustomize/base/kustomization.yaml
@@ -18,4 +18,4 @@ labels:
 images:
 - name: workspaces-backend
   newName: ghcr.io/kubeflow/notebooks/workspaces-backend
-  newTag: latest
+  newTag: 2.0.0-alpha.0
diff --git a/workspaces/controller/manifests/kustomize/base/manager/kustomization.yaml b/workspaces/controller/manifests/kustomize/base/manager/kustomization.yaml
index 9bab808..e467be1 100644
--- a/workspaces/controller/manifests/kustomize/base/manager/kustomization.yaml
+++ b/workspaces/controller/manifests/kustomize/base/manager/kustomization.yaml
@@ -19,4 +19,4 @@ labels:
 images:
 - name: workspaces-controller
   newName: ghcr.io/kubeflow/notebooks/workspaces-controller
-  newTag: latest
+  newTag: 2.0.0-alpha.0
diff --git a/workspaces/frontend/manifests/kustomize/base/kustomization.yaml b/workspaces/frontend/manifests/kustomize/base/kustomization.yaml
index e3c2960..d41cd34 100644
--- a/workspaces/frontend/manifests/kustomize/base/kustomization.yaml
+++ b/workspaces/frontend/manifests/kustomize/base/kustomization.yaml
@@ -16,4 +16,4 @@ labels:
 images:
 - name: workspaces-frontend
   newName: ghcr.io/kubeflow/notebooks/workspaces-frontend
-  newTag: latest
[...]

PR #1032, opened 2026-04-19
fix: preserve v-prefixed prerelease workspaces image tags

danish9039:
## Summary

This fixes the release-tag mismatch for workspaces prerelease images.

At the moment, `releasing/version/VERSION` and the generated manifests use `v`-prefixed versions such as `v2.0.0-alpha.1`, but the published GHCR workspaces images end up without the leading `v`. That makes downstream manifests reference image tags that do not exist and can lead to `ImagePullBackOff` in consumers such as `kubeflow/manifests`.

PR #1000 was a pragmatic workaround that aligned manifests to the currently published no-`v` tags. This PR fixes the underlying publishing behavior so future prerelease releases are consistent at the source.

## Root cause

`ws-build-image.yml` uses `docker/metadata-action` for semver tagging.

For prerelease versions (`alpha`, `beta`, `rc`), `docker/metadata-action` normalizes semver output to `{{version}}` unless `{{raw}}` is used. That strips the leading `v`, even when the workflow is given a value like `v2.0.0-alpha.1`.

In practice this creates the following mismatch:

- `VERSION` = `v2.0.0-alpha.1`
- generated manifests expect `v2.0.0-alpha.1`
- published GHCR image becomes `2.0.0-alpha.1`

## Fix

Use the raw prerelease version string when generating workspaces image tags, so the published image tag matches `VERSION` exactly.

This restores one canonical version string across:
- git tags / releases
- `releasing/version/VERSION`
- generated manifests
- published GHCR image tags

## Why this is preferable

This keeps the release contract simple and predictable:
- if the release version is `vX.Y.Z-*`, the published workspaces image tag is also `vX.Y.Z-*`

### DIFF
--- a/.github/workflows/ws-build-image.yml
+++ b/.github/workflows/ws-build-image.yml
@@ -14,7 +14,7 @@ on:
         type: boolean
       tag_with_semver:
         default: false
-        description: "if true, image tags will include the version (from semver_raw), always adds 'v' prefix if not already provided in semver_raw"
+        description: "if true, image tags will include the raw version from semver_raw"
         type: boolean
       tag_with_sha:
         default: false
@@ -78,7 +78,7 @@ jobs:
           flavor: |
             latest=${{ inputs.tag_with_latest }}
           tags: |
-            type=semver,priority=1000,pattern=v{{version}},enable=${{ inputs.tag_with_semver }},value=${{ inputs.semver_raw }}
+            type=semver,priority=1000,pattern={{raw}},enable=${{ inputs.tag_with_semver }},value=${{ inputs.semver_raw }}
             type=sha,priority=200,prefix=sha-,format=short,enable=${{ inputs.tag_with_sha }}
             type=sha,priority=100,prefix=sha-,format=long,enable=${{ inputs.tag_with_sha }}
 


christian-heusel:
What problem do we see here? 🤔 This issue regarding the ImagePullBackOff (see https://github.com/kubeflow/manifests/pull/3430#discussion_r3012964803) has already been fixed in https://github.com/kubeflow/notebooks/pull/1000 🤔

danish9039:
> What problem do we see here? 🤔 This issue regarding the ImagePullBackOff (see [kubeflow/manifests#3430 (comment)](https://github.com/kubeflow/manifests/pull/3430#discussion_r3012964803)) has already been fixed in #1000 🤔

@christian-heusel Actually here for prerelease versions (alpha, beta, rc), docker/metadata-action normalizes semver output to {{version}} unless {{raw}} is used. That strips the leading v, even when the workflow is given a value like v2.0.0-alpha.1.   this is creating mismatch on the PR https://github.com/kubeflow/manifests/pull/3430  hence imagePullBackoff 

VERSION = v2.0.0-alpha.1  , generated manifests expect v2.0.0-alpha.1  and published GHCR image becomes 2.0.0-alpha.1   , we need to merge this PR and update the manifest https://github.com/kubeflow/manifests/pull/3430/changes#diff-2145495ad76e99d7bebb590b3ab3c4fb6dec9b87f0092b3ff4c7adf9cd36311dR20 , see    https://github.com/kubeflow/notebooks/pkgs/container/notebooks%2Fworkspaces-backend/783904632?tag=2.0.0-alpha.1 have validated this on my fork on top of your commits , just need to change the image tag we get after merging this PR and , see my fork https://github.com/danish9039/manifests/pull/4

christian-heusel:
@danish9039 thanks for your help and I gotta apologize, it indeed seems like the PR that I referenced does not fix the issue given that the last image tag is missing the `v` prefix: https://github.com/kubeflow/notebooks/pkgs/container/notebooks%2Fworkspaces-controller/783903402?tag=2.0.0-alpha.1

However I'm quite confused now I could swear I tested this in a local cluster 😅 I'll also check out your other comments as soon as I can, had a few quite busy days! 🤗

christian-heusel (approved):

andyatmiami (approved):
/lgtm
/approve

thanks @danish9039 for addressing this more comprehensively than our last attempt.

this statement from @danish9039 mostly describes what is happening
- https://github.com/kubeflow/notebooks/pull/1032#issuecomment-4276865633

But to reword it slightly - as I was confused and @christian-heusel helped clarify further... the problematic logic in `metadata-action` is here:
- https://github.com/docker/metadata-action/blob/deaa0dee567db6b95038c14152b037af85a8fe3b/src/meta.ts#L187-L196

When the tag is a prerelease, the code checks `isRawStatement()` on your pattern. If that returns `false`, your pattern is **thrown away** and `{{version}}` is **hardcoded** instead. So `v{{version}}` fails the `isRawStatement` check, gets ignored, and falls back to `{{version}}` - stripping the v.

Only patterns that satisfy `isRawStatement() `(i.e. `{{raw}}`) are honoured for prereleases. For stable tags the else branch runs normally, which is why `v{{version}}` worked fine for `v1.2.3`.

Issue #1019, opened 2026-04-08
[TASK] Add Tolerations to WorkspaceKind Pod Config Modal


thaorell:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

Right now, "resources" is the only spec that is available in the Pod Config Modal. When creating new Pod Config, we should allow Joel, the platform admin, to configure NodeSelector, Affinity, Tolerations

  Proposed Additions
A list of tolerations allowing pods to schedule onto nodes with matching taints. Each toleration has: key, operator (Exists | Equal), value, effect (NoSchedule |  PreferNoSchedule | NoExecute), and optional tolerationSeconds.   

### Acceptance Criteria

  - Platform admin can add tolerations with key, operator, value, and effect fields
  - Operator field supports Exists and Equal options
  - Effect field supports NoSchedule, PreferNoSchedule, and NoExecute options
  - When operator is Exists, the value field is disabled/hidden 
  - When effect is NoExecute, an optional tolerationSeconds field is available
  - Platform admin can edit existing tolerations
  - Platform admin can remove individual tolerations 
  - Tolerations are persisted when saving the Pod Config
  - Tolerations are displayed when editing an existing Pod Config (immutable)

Snehadas2005:
Hi @thaorell, I'm interested in working on this task to add Tolerations to the WorkspaceKind Pod Config modal. I am familiar with the v2 frontend environment and would love to take this up to help move the wizard forward. Would it be okay if I work on this?

PR #1044, opened 2026-04-22
feat: Add tolerations to WorkspaceKind Form

thaorell:
closes: #1019

Added:
- Custom types
- Refactored WorkspaceKind PodConfig Tables to have expandable rows
- Add Toleration Modal and Table to PodConfig expanded section: The modal will automatically hide `Value` input if the operator is `Exists` and only render Toleration seconds field if the effect is `NoExecute`
- Add tests

Before: 
Screenshot 2026-04-30 at 1 47 07 PM

After:
Screenshot 2026-04-30 at 1 47 33 PM
Screenshot 2026-04-30 at 1 51 07 PM

### DIFF
--- a/workspaces/frontend/src/app/pages/WorkspaceKinds/Form/podConfig/WorkspaceKindFormPodConfig.tsx
+++ b/workspaces/frontend/src/app/pages/WorkspaceKinds/Form/podConfig/WorkspaceKindFormPodConfig.tsx
@@ -197,8 +197,9 @@ export const WorkspaceKindFormPodConfig: React.FC<WorkspaceKindFormPodConfigProp
           <>
             <div className="pf-v6-u-pl-xl pf-v6-u-pt-sm pf-v6-u-pb-sm">
               <div>
-                Pod configurations allow specifying resource allocation for Workspace Kinds. Expand
-                each configuration to view and manage its tolerations.
+                Pod configurations define the hardware resource options available to a Workspace.
+                Expand each configuration to access nodeSelector and tolerations, allowing you to
+                constrain pods to specific node groups.
               </div>
             </div>
             <WorkspaceKindFormPaginatedTable
--- a/workspaces/frontend/src/app/pages/WorkspaceKinds/Form/podConfig/WorkspaceKindFormPodConfig.tsx
+++ b/workspaces/frontend/src/app/pages/WorkspaceKinds/Form/podConfig/WorkspaceKindFormPodConfig.tsx
@@ -43,7 +43,7 @@ export const WorkspaceKindFormPodConfig: React.FC<WorkspaceKindFormPodConfigProp
   const [editIndex, setEditIndex] = useState<number | null>(null);
   const [deleteIndex, setDeleteIndex] = useState<number | null>(null);
   const [currConfig, setCurrConfig] = useState<WorkspaceKindPodConfigValue>({ ...emptyPodConfig });
-  const [tolerationModalOpenFor, setTolerationModalOpenFor] = useState<number | null>(null);
+  const [tolerationModalOpenFor, setTolerationModalOpenFor] = useState<string | null>(null);
 
   const clearForm = useCallback(() => {
     setCurrConfig({ ...emptyPodConfig });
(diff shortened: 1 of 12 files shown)


workspaces/frontend/src/app/pages/WorkspaceKinds/Form/podConfig/WorkspaceKindFormPodConfig.tsx
```diff
@@ -87,6 +94,69 @@ export const WorkspaceKindFormPodConfig: React.FC<WorkspaceKindFormPodConfigProp
...
+  podConfigRef.current = podConfig;
+  const updatePodConfigRef = useRef(updatePodConfig);
+  updatePodConfigRef.current = updatePodConfig;
+
+  const tolerationSettersCache = useRef(
+    new Map<number, React.Dispatch<React.SetStateAction<TolerationEntry[]>>>(),
+  );
+  const modalSettersCache = useRef(
+    new Map<number, React.Dispatch<React.SetStateAction<boolean>>>(),
+  );
```
caponetto, line 112: `tolerationSettersCache` and `modalSettersCache` are `Map<number, ...>` refs that grow monotonically; entries are never removed when pod configs are deleted. While functionally correct (the `useMemo` re-maps fresh indices), the caches accumulate stale closures over time. For a form with frequent add/remove cycles, consider clearing the caches when `podConfig.values.length` changes, or switching to a `useMemo`-based approach without manual caching.

workspaces/frontend/src/app/types.ts
```diff
@@ -82,6 +82,28 @@ export interface WorkspaceKindImagePort {
   protocol: 'HTTP'; // ONLY HTTP is supported at the moment, per https://github.com/thesuperzapper/kubeflow-notebooks-v2-design/blob/main/crds/workspace-kind.yaml#L275
 }
 
+export enum TolerationOperator {
+  None = '',
```
caponetto: `TolerationOperator.None = ''` and `TolerationEffect.None = ''` use empty strings as enum values. When these are serialized and sent to the backend (or eventually to the Kubernetes API), empty strings may behave differently from an omitted field. Please confirm that the backend handles empty-string operator/effect correctly, or consider using `undefined`/omission to represent "no selection" and making the field optional in `TolerationEntry`.

workspaces/frontend/src/app/pages/WorkspaceKinds/Form/podConfig/WorkspaceKindFormPodConfig.tsx
```diff
@@ -123,6 +193,13 @@ export const WorkspaceKindFormPodConfig: React.FC<WorkspaceKindFormPodConfigProp
         )}
         {podConfig.values.length > 0 && (
           <>
+            <div className="pf-v6-u-pl-xl pf-v6-u-pt-sm pf-v6-u-pb-sm">
```
caponetto: The descriptive text block uses raw `<div>` elements with PatternFly utility classes. Since `Content` is already imported, prefer `<Content component="p">` for the description paragraph. This ensures consistent typography and spacing from PatternFly's design system rather than relying on utility classes alone.

caponetto (commented):
Left a few comments, please check if they make sense.

Only one UI observation (not blocker): The Toleration modal has a state where `Toleration Seconds` is not shown, for instance:

Image


When the item is added, the column is displayed, though:

Image


This could confuse users about where that comes from. I would rather disable `Toleration Seconds` with a helper reason than hide it.

workspaces/frontend/src/app/pages/WorkspaceKinds/Form/podConfig/WorkspaceKindFormPodConfig.tsx
thaorell: I'm applying a fix to delete stale cache entries in `handleDelete`, so when a config is deleted, the config's tolerations will be deleted as well

workspaces/frontend/src/app/types.ts
thaorell: After reading the [Kubernetes docs](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) in depth, I am setting the default Operator to `Equal`

thaorell: I also added a description for a None effect, which will be coerced to `undefined` upon submission

ederign (approved):

PR #1067, opened 2026-04-29
fix: toolbar actions should be outside filter ToolbarGroup

paulovmr:
Before:
image

After:
image


thaorell (approved):
/lgtm

andyatmiami:
@paulovmr - at risk of destroying any good will I have built up with you.. I **THINK** the `WorkspaceKind` screen suffers the same problem..

could we get a comparable fix in for that as well to be done with these "disappearing buttons" completely 😇

ederign (approved):

ederign:
/approve
/lgtm

@paulovmr merging this assuming you will do Andy's request in a follow up PR

Issue #1042, opened 2026-04-21
DeleteModal has no guard against concurrent execution


thaorell:
### Checks

- [x] I have searched the [existing issues](https://github.com/kubeflow/notebooks/issues).
- [x] My issue is related to one of the components in the [`kubeflow/notebooks`](https://github.com/kubeflow/notebooks) repository.

### Kubeflow Notebooks Version

2.0.0

### Kubeflow Platform

N/A

### Kubernetes Distribution

N/A

### Kubernetes Version

```shell
N/A
```

### Description

How to reproduce? Hitting the enter key twice on DeleteModal will submit two different API calls
  1. First Enter: onSubmit fires → handleDelete() runs → setIsDeleting(true) is called, but React batches this state update — it won't takeeffect until the next render.
  2. Second Enter (before React re-renders): onSubmit fires again → handleDelete() runs again → isDeleting is still false in the closure, so the only check is inputValue === resourceName, which is still true → onDelete(resourceName) is called a second time.  

### Relevant Logs

```shell

```

PR #1088, opened 2026-05-08
fix: add guard against concurrent execution in DeleteModal

Snehadas2005:
**Fixes** #1042

Rapidly hitting the "Enter" key or clicking the "Delete" button twice in the `DeleteModal` could trigger multiple duplicate API calls. This occurred because React's asynchronous state updates were too slow to block subsequent event triggers within the same render cycle.

This PR implements a synchronous guard to ensure only one deletion request is processed at a time.

**Changes:** 
- Synchronous Lock: Introduced a `useRef `guard (`isDeletingRef`) to block concurrent execution before the next render cycle immediately.
- HandleDelete Cleanup: Restructured the `handleDelete` function to verify the lock and reset it only upon failure, ensuring the action is "one-shot" per modal session.


workspaces/frontend/src/shared/components/DeleteModal.tsx
```diff
@@ -40,28 +40,33 @@ const DeleteModal: React.FC<DeleteModalProps> = ({
   const [inputValue, setInputValue] = useState('');
   const [isDeleting, setIsDeleting] = useState(false);
```
thaorell, line 41: Consider removing `isDeleting` state to avoid maintaining both ref and state and having the `Cancel` button always visible. Test to see if hitting the `Cancel` button while deleting causes any problem

thaorell (commented):
Hey @Snehadas2005, good work here. The comment I left is not a blocking issue, so let me know how it goes

Snehadas2005:
Thank you for the suggestion @thaorell! I have pushed an update that removes the `isDeleting` state entirely, as the `ActionButton` already manages its own internal loading state. I also tested hitting Cancel while a deletion is active; the modal closes cleanly, and the deletion completes successfully in the background without throwing any errors.

thaorell (commented):
/lgtm

workspaces/frontend/src/shared/components/DeleteModal.tsx
```diff
@@ -38,30 +38,31 @@ const DeleteModal: React.FC<DeleteModalProps> = ({
...
+
+    isDeletingRef.current = true;
+    setError(null);
+
+    try {
+      await onDelete(resourceName);
+      onClose();
+    } catch (err) {
+      setError(extractErrorMessage(err));
+      isDeletingRef.current = false;
```
paulovmr, line 69: Please consider having this line on the `finally` instead of the `catch` block.

paulovmr (changes requested):
Hi @Snehadas2005 , thanks for the pull request! I've reviewed and made one comment for a change, if you could take a look please. Also, could you please consider adding some test to cover this behavior?

Snehadas2005:
Hi @paulovmr, thank you for the review!
I refactored the delete handler to place `isDeletingRef.current = false;` in a finally block for cleaner execution. I have also added the requested unit tests for DeleteModal.tsx to cover the loading/deleting state lifecycle. 
Ready for your feedback! 🤗

workspaces/frontend/package.json
```diff
@@ -131,6 +131,7 @@
     "classnames": "^2.2.6",
     "date-fns": "^4.1.0",
     "dompurify": "^3.2.4",
+    "esbuild": "0.28.0",
```
thaorell: Hi @Snehadas2005, can you remove this package?

thaorell (changes requested):

thaorell (commented):
You still need to revert the changes in `package-lock.json`

Snehadas2005:
@thaorell, thank you for your guidance. I successfully removed the esbuild package from both `package.json` and `package-lock.json`.

thaorell (changes requested):
`workspaces/frontend/package-lock.json` still has 1,964 lines of changes here. Please revert the change

Snehadas2005:
@thaorell Thank you for the throughout guidance, I had successfully revered the changes from `package-lock.json`.

thaorell:
@Snehadas2005 the code looks ok to me but there is a failed check

Snehadas2005:
Hi @thaorell! Thank you so much for your patience and thorough guidance. 
I wanted to share a quick note on the Dockerfile changes I've made alongside the DeleteModal fix. During investigation of the failing `image-build` CI check, I noticed two engine mismatches that were silently causing issues:

1. **Node version**: The `mod-arch-core`, `mod-arch-kubeflow`, and `mod-arch-shared` packages (all at v1.16.1) declare a peer requirement of `node: >=22.0.0`, while our Dockerfile was using `node:20-slim`. This mismatch doesn't break the build outright, but it produces engine warnings on every `npm ci` run and could cause subtle runtime incompatibilities as those packages evolve.

2. **npm version**: Our `package.json` engines field requires `npm: ^11.10.0`, but `node:20-slim` ships with npm 10.x. Upgrading to `node:22-slim` and explicitly installing the latest npm brings both into alignment.

Updating the base image to `node:22-slim` (in both `Dockerfile` and `Dockerfile.dev`) and adding `RUN npm install -g npm@latest` resolves both mismatches cleanly and future-proofs us against stricter engine enforcement in upcoming dependency releases.

I've also updated the `engines` field in `package.json` to reflect `"node": "^22"` for consistency.

Happy to revert or adjust any of this if it falls outside the intended scope of this PR, please let me know!

thaorell:
@Snehadas2005 thanks for figuring it out. I'll fix it in a separate PR.

Snehadas2005:
Thank you for the update, @thaorell! 
Should I revert the Dockerfile and `package.json` changes on this branch to keep the PR focused on the DeleteModal fix? Let me know, and I’ll handle it.

thaorell:
@Snehadas2005 yes you should revert the change in those files

Snehadas2005:
Hi @thaorell, could you please review the changes and let me know if any additional changes are required?

thaorell (commented):
/lgtm

paulovmr (approved):
/lgtm
/approve

PR #1082, opened 2026-05-06
fix: align workspace path parameter name in OpenAPI annotations

christian-heusel:
`GetWorkspaceHandler` and `DeleteWorkspaceHandler` used `{workspace_name}` in their Swag `@Param`/`@Router` annotations while `UpdateWorkspaceHandler` used `{name}`, matching `ResourceNamePathParam`. This caused swagger to emit two separate path entries for the same route template, which OpenAPI linters flag as "Equivalent paths are not allowed".

Align all three handlers to `{name}` and regenerate the spec.

Link: https://editor.swagger.io?url=https://raw.githubusercontent.com/kubeflow/notebooks/e091ea3a398e008e988056c2d5ac6da07bc04719/workspaces/backend/openapi/swagger.json

I have noticed this issue while reviewing https://github.com/kubeflow/notebooks/pull/1081.

It can also be reproduced on the commandline with spectral or vacuum:
```
$ npx @stoplight/spectral-cli lint workspaces/backend/openapi/swagger.json
 1543:52    error  path-params            Paths "/workspaces/{namespace}/{name}" and "/workspaces/{namespace}/{workspace_name}" must not be equivalent.  paths./workspaces/{namespace}/{workspace_name}
```

### `Operations` violations
<details><summary>🔴 Errors: 1</summary>
🔴 no-ambiguous-paths : 1

| location | JSONPath                                            |
| -------- | --------------------------------------------------- |
| `1543:9` | $.paths['/workspaces/{namespace}/{workspace_name}'] |

</details>

<details><summary>⚠ Warnings: 1</summary>
⚠ no-http-verbs-in-path: 1

| location  | JSONPath                                                         |
| --------- | ---------------------------------------------------------------- |
| `1059:66` | $.paths['/workspacekinds/{name}/podtemplate/options/listvalues'] |

</details>

PR #1089, opened 2026-05-09
feat: add OpenAPI spec linting with vacuum

### LINKED ISSUE #1087: [TASK] Setup linting for the OpenAPI spec
### Certification

- [ ] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

In the context of https://github.com/kubeflow/notebooks/pull/1082 we realized that it would maybe be good to lint the OpenAPI spec that we produce with the `make swag` command. A quick check on that issue revealed that there are a few different options:
- [daveshanley/vacuum](github.com/daveshanley/vacuum): Tool is written in go and should be easy to integrate. Since the default output is a bit noisy something like  `vacuum lint -q --pipeline-output workspaces/backend/openapi/swagger.json` would be good and maybe we can even suppress the info level warnings from that output.
- [stoplightio/spectral](https://github.com/stoplightio/spectral): Tool is written with JS and requires manual setup for the various lint rules.

Given that both tools seem to be fairly popular and the go based one is easier to integrate into our current tooling ecosystem I would probably go with that, however maybe there are better tools that my quick google did not reveal 🤗

Also it is up for debate whether we want to include this only on the CLI (i.e. via the Makefiles `make lint` target) or else.

### Acceptance Criteria

- [ ] Some lint tool is being run whenever the OpenAPI spec is changed
- [ ] The tool is configured to produce good signal-to-noise output

madmecodes:
## Summary

- Integrates [daveshanley/vacuum](https://github.com/daveshanley/vacuum) as an OpenAPI linter into the backend Makefile
- Adds `make lint-openapi` target and wires it into the existing `make lint` target
- Uses the project's existing `go-install-tool` pattern for automatic binary management

## Details

vacuum is installed on demand via `go install` (same pattern as golangci-lint, swag, etc.), pinned to v0.26.4. The lint target runs with `--fail-severity error` so builds fail on actual spec errors (like the ambiguous paths caught in #1082) while still displaying warnings and info in the summary.

Fixes #1087

### DIFF
--- a/workspaces/backend/Makefile
+++ b/workspaces/backend/Makefile
@@ -68,12 +68,16 @@ test: swag fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS=${KUBEBUILDER_ASSETS} go test ./... -coverprofile cover.out
 
 .PHONY: lint
-lint: golangci-lint ## Run golangci-lint linter
+lint: golangci-lint lint-openapi ## Run golangci-lint and vacuum linters
 	$(GOLANGCI_LINT) run
 
 .PHONY: lint-fix
 lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 	$(GOLANGCI_LINT) run --fix
+
+.PHONY: lint-openapi
+lint-openapi: vacuum ## Run vacuum linter against the OpenAPI spec
+	$(VACUUM) lint -q --no-banner --details --errors --fail-severity error openapi/swagger.json
 
 ##@ Swagger
 ALL_GO_DIRS := $(shell find . -type f -name '*.go' -exec dirname {} \; | sed 's|^\./||' | sort -u)
@@ -127,12 +131,14 @@ KUSTOMIZE := $(LOCALBIN)/kustomize
 ENVTEST ?= $(LOCALBIN)/setup-envtest
 GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 SWAGGER = $(LOCALBIN)/swag
+VACUUM = $(LOCALBIN)/vacuum
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.5.0
 ENVTEST_VERSION ?= release-0.19
 GOLANGCI_LINT_VERSION ?= v1.64.8
 SWAGGER_VERSION ?= v1.16.6
+VACUUM_VERSION ?= v0.26.4
 
 .PHONY: SWAGGER
 SWAGGER: $(SWAGGER)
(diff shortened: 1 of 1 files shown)


workspaces/backend/Makefile
```diff
@@ -68,13 +68,17 @@ test: swag fmt vet envtest ## Run tests.
...
+lint: golangci-lint lint-openapi ## Run golangci-lint and vacuum linters
 	$(GOLANGCI_LINT) run
 
 .PHONY: lint-fix
 lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 	$(GOLANGCI_LINT) run --fix
 
+.PHONY: lint-openapi
+lint-openapi: vacuum ## Run vacuum linter against the OpenAPI spec
+	$(VACUUM) lint -q --no-banner --details --errors --fail-severity error openapi/swagger.json
```
christian-heusel, line 80: I don't really like the default vaccum output, could you try to integrate `--pipeline-output` and only make Warnings and Errors visible?

christian-heusel (commented):
Thanks for working on this! 🤗 Lets refine the exact vacuum output a bit, otherwise it looks fine! 👍🏻

christian-heusel:
Also did you use AI in any capacity while creating this contribution? I see that your commit does not carry a `Assisted-By: <...>` tag, so I wanted to make sure that you're aware of the [Kubeflow AI Policy](https://www.kubeflow.org/docs/about/ai_policy/). 🤗 🤖 
/ok-to-test

workspaces/backend/Makefile
Snehadas2005, line 80:
```suggestion
	$(VACUUM) lint -q --no-banner --pipeline-output --fail-severity error openapi/swagger.json
```

Snehadas2005 (commented):
Hey @madmecodes, I’ve updated the flags here to address Christian's feedback regarding the output noise:
- `--pipeline-output`: Switched to a cleaner, CI-friendly summary.
- `--no-banner:` Removes the ASCII header to reduce log noise.
- `--fail-severity error`: Fails the build only on spec errors, showing warnings.
This should make the signal-to-noise ratio much better for the maintainers!

christian-heusel:
@madmecodes could you please address the review feedback above? 🤗

workspaces/backend/Makefile
RajGautam2004: yes 

> I don't really like the default vaccum output, could you try to integrate `--pipeline-output` and only make Warnings and Errors visible?

yes i will do that
figuring out . sorry for above

PR #1090, opened 2026-05-10
fix: return HTTP 422 with field-level errors for JSON type mismatches

andyatmiami:
ℹ️ _NO GH ISSUE_

Every handler that calls `DecodeJSON` previously returned a generic HTTP `400` for JSON type mismatches (e.g., sending "yes" for a bool field), leaking Go-internal error text. Now, `json.UnmarshalTypeError` is detected and converted to structured validation errors with the field path, expected type, and actual type — consistent with the existing `failedValidationResponse` infrastructure.

### DIFF
--- a/workspaces/backend/api/helpers.go
+++ b/workspaces/backend/api/helpers.go
@@ -24,7 +24,10 @@ import (
 	"maps"
 	"mime"
 	"net/http"
+	"reflect"
 	"strings"
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/kubeflow/notebooks/workspaces/backend/api/constants"
 )
@@ -81,6 +84,11 @@ func (a *App) DecodeJSON(r *http.Request, v any) error {
 			return err
 		}
 
+		// NOTE: we don't wrap this error so we can unpack it in the caller
+		if a.IsUnmarshalTypeError(err) {
+			return err
+		}
+
 		// provide better error message for the case where the body is empty
 		// NOTE: io.EOF is only returned when the body is completely empty or contains only whitespace.
 		//       If there's any actual JSON content (even malformed), json.Decoder returns different errors.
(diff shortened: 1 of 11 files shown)


workspaces/backend/api/helpers.go
```diff
@@ -109,6 +117,72 @@ func (a *App) IsEOFError(err error) bool {
...
+	return errors.As(err, &unmarshalTypeError)
+}
+
+// FieldErrorsFromUnmarshalTypeError converts a json.UnmarshalTypeError into a field.ErrorList
+// with a single entry describing the type mismatch using user-friendly type names.
+func FieldErrorsFromUnmarshalTypeError(err error) field.ErrorList {
+	var unmarshalTypeError *json.UnmarshalTypeError
+	if !errors.As(err, &unmarshalTypeError) {
+		return nil
+	}
```
christian-heusel, line 129: We should most likely use `IsUnmarshalTypeError()` here to keep things more DRY.

workspaces/backend/api/helpers.go
```diff
@@ -109,6 +117,72 @@ func (a *App) IsEOFError(err error) bool {
...
+}
+
+// goTypeToJSONTypeName maps a Go reflect.Type to a user-friendly JSON type name.
+func goTypeToJSONTypeName(t reflect.Type) string {
+	if t == nil {
+		return "unknown"
+	}
+	for t.Kind() == reflect.Ptr {
+		t = t.Elem()
+	}
```
christian-heusel: Could you please add a comment explaining what this `for`-loop does? For me as a reader it wasn't immediately obvious (ofcourse after thinking about it for a moment it is), so this most likely warrants a small explanation like the following:
```
// We keep calling .Elem() to peel off pointer indirections until we reach the underlying concrete type, i.e. **int -> *int -> int.
```

christian-heusel: Also just maybe FYI because I found it an interesting edge case:
```
❯ Are there cases where this loop is not deterministic?                                                   

● Yes — if a self-referential pointer type is passed in, the loop is infinite.

  Go allows this:

  type A *A

  A is a pointer type whose element type is A itself. Under reflection:

  - t.Kind() → reflect.Ptr
  - t.Elem() → type A
  - t.Elem().Kind() → reflect.Ptr (still!)
  - ...forever

  In practice this can't happen here because json.UnmarshalTypeError.Type is populated by the standard
  library's JSON decoder, which itself rejects self-referential pointer types and would never produce such
   a reflect.Type. So the loop is safe in this specific context.

  But if goTypeToJSONTypeName were ever reused more broadly, it would be worth guarding against — either
  with a depth limit or by checking whether t.Elem() == t before the next iteration.
```

workspaces/backend/api/helpers.go
```diff
@@ -109,6 +117,72 @@ func (a *App) IsEOFError(err error) bool {
...
+// fieldPathFromJSONPath converts a dot-separated JSON field path (e.g., "data.podTemplate.paused")
+// into a field.Path.
+func fieldPathFromJSONPath(jsonPath string) *field.Path {
+	parts := strings.Split(jsonPath, ".")
+	path := field.NewPath(parts[0])
+	for _, part := range parts[1:] {
+		path = path.Child(part)
+	}
+	return path
+}
```
christian-heusel: I think this function is mostly unnecessary and should either become a simple wrapper or inline the logic at the current call-sites since `newPath()` can received variadic input:

```go
func NewPath(name string, moreNames ...string) *Path
```

Potentially we can replace the function with the following:
```go
parts := strings.Split(jsonPath, ".")
fieldPath := field.NewPath(parts[0], parts[1:]...)
```

This works due to the following:
```
❯ How does this work when there is only one entry in `parts`?                                             

● strings.Split("paused", ".") returns ["paused"], so parts[0] is "paused" and parts[1:] is an empty
  slice. Spreading an empty slice into a variadic parameter is valid Go — it's equivalent to calling
  field.NewPath("paused") with no extra arguments, which is exactly what the original code did.
```

christian-heusel (changes requested):
Thanks for tackling this category of leftover todos 😄🚀

I have left a bit of review below 🤗

workspaces/backend/api/helpers.go
andyatmiami: The errors.As on line 130 does double duty:                                                                                                                                                                                                                           
                                                                                                                                  
  1. Guard — returns false if it's not an UnmarshalTypeError                                                                                                                                                                                                                                                                                               
  2. Extraction — populates unmarshalTypeError for use on lines 134-143                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                           
If you replace the guard with !a.IsUnmarshalTypeError(err), the variable declared on line 129 stays nil. You'd then need a second errors.As call to populate it, which is strictly worse — two errors.As calls instead of one, for no benefit.

workspaces/backend/api/helpers.go
andyatmiami: great suggestion - will apply

workspaces/backend/api/helpers.go
andyatmiami: excellent thing to highlight.

i added the inline comment you suggested.. and even put a defensive check in the code (itself with its own inline description) around the self-referential pointer type.. as its relatively cheap (even if its impossible to hit with code as written/constructed **today**... who knows what tomorrow brings 😎 )

christian-heusel (approved):
Nice work @andyatmiami and thanks for addressing all the  feedback 🥳 💯

workspaces/backend/api/response_errors_test.go
thesuperzapper, line 1: We should make sure we test the "provide exta json fields" case, as we set `decoder.DisallowUnknownFields()` in `DecodeJSON`:

https://github.com/kubeflow/notebooks/blob/02c5bd25f81d95345334335dcdd87293761df931/workspaces/backend/api/helpers.go#L78-L96

workspaces/backend/api/helpers.go
```diff
@@ -109,6 +117,70 @@ func (a *App) IsEOFError(err error) bool {
...
+func FieldErrorsFromUnmarshalTypeError(err error) field.ErrorList {
+	var unmarshalTypeError *json.UnmarshalTypeError
+	if !errors.As(err, &unmarshalTypeError) {
+		return nil
+	}
+
+	expectedType := goTypeToJSONTypeName(unmarshalTypeError.Type)
+	detail := fmt.Sprintf("got JSON %s, but field requires %s", unmarshalTypeError.Value, expectedType)
+
+	if unmarshalTypeError.Field == "" {
```
thesuperzapper, line 137: Can we add a comment which explains what situation we end up with an empty field string.

workspaces/backend/api/helpers.go
```diff
@@ -109,6 +117,70 @@ func (a *App) IsEOFError(err error) bool {
...
+	for t.Kind() == reflect.Ptr {
+		next := t.Elem()
+		// guard against self-referential pointer types (e.g., `type A *A`) which would loop infinitely.
+		// this should never occur in practice because json.UnmarshalTypeError.Type is populated by the stdlib JSON decoder,
+		// which rejects self-referential pointer types.
+		if next == t {
+			break
+		}
+		t = next
+	}
```
thesuperzapper: Lets handle both these cases as part of the switch statement (nil, and pointer types).

To prevent against the strange situation where we get a self-referential pointer, we should simply do an if-condition before the switch, which panics in this case.

thesuperzapper (changes requested):

workspaces/backend/api/response_errors_test.go
andyatmiami: Unfortunately the error returned in this case is a plain error - so we couldn't easily trap on this case unless we wanted to mess around with (brittle) string matching...
- https://github.com/golang/go/blob/go1.24.13/src/encoding/json/decode.go#L735

For now - I will add a test to "document" this behavior - but I think we should live with the `400` (at least in the context of this PR)

christian-heusel (changes requested):
There are a few new lint warnings (see #1246) that need to be fixed to make CI green🤔

PR #1097, opened 2026-05-14
fix: add namespace to storageclasses api and check pvc create for auth

christian-heusel:
Regular Kubeflow users get 403 on /storageclasses because the backend issues a cluster-wide SAR (namespace="") that requires a ClusterRoleBinding, but users only have a namespace-scoped RoleBinding.

- Add namespace query param to GET /storageclasses: when provided the SAR
  is scoped to that namespace (matching the existing workspacekinds pattern) and checks if the users is allowed to create a PVC in the namespace. This is allowing regular users through; without it the endpoint remains admin-only
- Drop duplication linter for `"api/.*_handler.go` and remove leftover nolint directives

### DIFF
--- a/workspaces/backend/api/storageclasses_handler.go
+++ b/workspaces/backend/api/storageclasses_handler.go
@@ -44,7 +44,7 @@ type StorageClassListEnvelope Envelope[[]models.StorageClassListItem]
 //	@Failure		422			{object}	ErrorEnvelope				"Unprocessable Entity. Validation error."
 //	@Failure		500			{object}	ErrorEnvelope				"Internal server error"
 //	@Router			/storageclasses [get]
-func (a *App) GetStorageClassesHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) { //nolint:dupl
+func (a *App) GetStorageClassesHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 	namespace := r.URL.Query().Get(constants.NamespaceQueryParam)
 
 	// validate query parameters
--- a/workspaces/backend/api/storageclasses_handler.go
+++ b/workspaces/backend/api/storageclasses_handler.go
@@ -20,8 +20,11 @@ import (
 	"net/http"
 
 	"github.com/julienschmidt/httprouter"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"github.com/kubeflow/notebooks/workspaces/backend/api/constants"
 	"github.com/kubeflow/notebooks/workspaces/backend/internal/auth"
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/helper"
 	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/storageclasses"
 )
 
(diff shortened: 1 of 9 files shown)


workspaces/backend/api/storageclasses_handler.go
```diff
@@ -30,20 +33,40 @@ type StorageClassListEnvelope Envelope[[]models.StorageClassListItem]
...
-		auth.NewResourcePolicy(auth.VerbList, auth.StorageClasses, auth.ResourcePolicyResourceMeta{}),
+	var authPolicies []*auth.ResourcePolicy
+	if namespace != "" {
+		authPolicies = []*auth.ResourcePolicy{
+			auth.NewResourcePolicy(auth.VerbList, auth.StorageClasses, auth.ResourcePolicyResourceMeta{Namespace: namespace}),
+		}
+	} else {
+		authPolicies = []*auth.ResourcePolicy{
+			auth.NewResourcePolicy(auth.VerbList, auth.StorageClasses, auth.ResourcePolicyResourceMeta{}),
+		}
```
christian-heusel, line 71: When talking to Mathew he said that we should require permissions for PVC create for this namespace, however I'm not sure if that is semantically correct 🤔

workspaces/backend/api/storageclasses_handler.go
```diff
@@ -30,20 +33,40 @@ type StorageClassListEnvelope Envelope[[]models.StorageClassListItem]
...
-//	@Failure		500	{object}	ErrorEnvelope				"Internal server error"
+//	@Param			namespace	query		string						false	"Namespace to use for authorization scoping"	extensions(x-example=kubeflow-user-example-com)
+//	@Success		200			{object}	StorageClassListEnvelope	"Successful storage classes response"
+//	@Failure		401			{object}	ErrorEnvelope				"Unauthorized"
+//	@Failure		403			{object}	ErrorEnvelope				"Forbidden"
+//	@Failure		422			{object}	ErrorEnvelope				"Unprocessable Entity. Validation error."
+//	@Failure		500			{object}	ErrorEnvelope				"Internal server error"
 //	@Router			/storageclasses [get]
-func (a *App) GetStorageClassesHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+func (a *App) GetStorageClassesHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) { //nolint:dupl // TODO: Abstract common API patterns once implemented
```
christian-heusel: How would we work around the duplication lint here? What would be a sensible functionality to extract for both functions to share?

workspaces/controller/manifests/kustomize/base/manager/user_cluster_roles.yaml
```diff
@@ -71,3 +71,11 @@ rules:
...
   - list
   - watch
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - get
+  - list
+  - watch
```
christian-heusel: Is this the right place to add these permissions? Where else should we add them?

workspaces/backend/api/storageclasses_handler.go
thesuperzapper: While we should be creating helper functions to reduce duplication WITHIN the handler methods, the duplication of structure between similar resource handers should not indicate the need to make a common handler method across resource types.

As a result, lets just globally disable the `dupl` check in the `api/*_handler.go` files, similar to how we disable some checks in the test files: https://github.com/kubeflow/notebooks/blob/7afd2a6ef8bb28251bf17d1a48d738fef75c6114/workspaces/backend/.golangci.yml#L21-L28

workspaces/backend/api/storageclasses_handler.go
```diff
@@ -30,20 +33,40 @@ type StorageClassListEnvelope Envelope[[]models.StorageClassListItem]
...
-//	@Description	Returns a list of all storage classes in the cluster.
+//	@Description	Returns a list of all storage classes in the cluster. When namespace is provided, authorization checks whether the user can list storage classes in that namespace instead of requiring a cluster-wide permission.
 //	@Tags			storageclasses
 //	@ID				listStorageClasses
 //	@Produce		application/json
-//	@Success		200	{object}	StorageClassListEnvelope	"Successful storage classes response"
-//	@Failure		401	{object}	ErrorEnvelope				"Unauthorized"
-//	@Failure		403	{object}	ErrorEnvelope				"Forbidden"
-//	@Failure		500	{object}	ErrorEnvelope				"Internal server error"
+//	@Param			namespace	query		string						false	"Namespace to use for authorization scoping"	extensions(x-example=kubeflow-user-example-com)
```
thesuperzapper:
```suggestion
//	@Param			namespace	query		string						false	"Namespace to request storage classes for."	extensions(x-example=kubeflow-user-example-com)
```

workspaces/backend/api/storageclasses_handler.go
thesuperzapper, line 71: The check should be that the user can create a PVC, because that is the only context in which a non admin user will call this API.

We should document that in the code like we do on all other conditional ACL rules:

```suggestion
	if namespace != "" {
		// user intends to create a pvc in the namespace
		authPolicies = []*auth.ResourcePolicy{
			auth.NewResourcePolicy(auth.VerbCreate, auth.PersistentVolumeClaims, auth.ResourcePolicyResourceMeta{Namespace: namespace}),
		}
	} else {
		// administrative listing of storage classes
		authPolicies = []*auth.ResourcePolicy{
			auth.NewResourcePolicy(auth.VerbList, auth.StorageClasses, auth.ResourcePolicyResourceMeta{}),
		}
```

workspaces/backend/api/storageclasses_handler.go
```diff
@@ -30,20 +33,40 @@ type StorageClassListEnvelope Envelope[[]models.StorageClassListItem]
 // GetStorageClassesHandler returns a list of all storage classes in the cluster.
 //
 //	@Summary		List storage classes
-//	@Description	Returns a list of all storage classes in the cluster.
+//	@Description	Returns a list of all storage classes in the cluster. When namespace is provided, authorization checks whether the user can list storage classes in that namespace instead of requiring a cluster-wide permission.
```
thesuperzapper: Update this to align with the fact we are checking the ability to create a a pvc in the namespace now.

workspaces/controller/manifests/kustomize/base/manager/user_cluster_roles.yaml
christian-heusel: ➡️ drop it

thesuperzapper (changes requested):

thesuperzapper:
Thanks @christian-heusel , after we merge we need to update the frontend to always pass the `namespace` query param when listing StorageClasses (when creating PVCs in WS create/update flow).

/approve
/lgtm

cc @thaorell

Issue #861, opened 2026-01-29
[TASK] Integrate /listvalues API into Workspace Wizard for Image and Pod Config selection


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

The Workspace Wizard's Image and Pod Config selection steps must integrate with the `/listvalues` API endpoint to support compatibility-aware filtering of options. This enables the Compatibility Selectors feature (Epic #681) to dynamically filter options based on user selections and namespace context.

### Context

When users progress through the Workspace Wizard, they should only see Image and Pod Config options that are compatible with their prior selections. Platform administrators (Joel) can define rules on each WorkspaceKind that determine option visibility. For example:
- A GPU-requiring image should only show compatible pod configurations
- Pod configs with GPUs should be hidden when a CPU-only image is selected
- Options can be filtered based on namespace labels

The backend evaluates these rules and returns a `ruleEffects` attribute on each option indicating whether it should be hidden.

> **Reference**: See issue #683 for the complete API proposal and architectural details.

### API Endpoint

```
POST /api/v1/workspacekinds/{name}/podtemplate/options/listvalues
```

**Request**:
```json
{
  "data": {
    "context": {
      "namespace": {
        "name": "user-namespace"
      },
      "imageConfig": {
        "id": "selected-image-id"
      },
      "podConfig": {
        "id": "selected-pod-config-id"
      }
    }
  }
}
```

**Context behavior**:
- `context` is **optional**
- If neither `podConfig` nor `imageConfig` specified → both `values` collections returned with full filtering
- If `imageConfig.id` specified → `imageConfig.values` filtered to selected item, `podConfig.values` filtered by rules
- If `podConfig.id` specified → `podConfig.values` filtered to selected item, `imageConfig.values` filtered by rules

**Response**:
```json
{
  "data": {
    "imageConfig": {
      "default": "jupyterlab_scipy_190",
      "values": [
        {
          "id": "jupyterlab_scipy_190",
          "displayName": "jupyter-scipy:v1.9.0",
          "description": "JupyterLab, with SciPy Packages",
          "labels": [
            { "key": "python_version", "value": "3.11" }
          ],
          "hidden": false,
          "ruleEffects": {
            "uiHide": false,
            "aclDeny": false
          }
        }
      ]
    },
    "podConfig": {
      "default": "tiny_cpu",
      "values": [
        {
          "id": "tiny_cpu",
[...]

PR #1098, opened 2026-05-14
feat: use listValues to render Image and Pod Config steps in Workspac…

thaorell:
closes: #861 

Integrates the POST /workspacekinds/{name}/podtemplate/options/listvalues API endpoint into the Workspace Wizard form. Image and Pod Config options are now fetched via listValues instead of reading inline values from the WorkspaceKind object, enabling future compatibility selector filtering (e.g., hiding GPU pod configs when a CPU-only image is selected).

What changed

**New hook: usePodTemplateOptionsListValues**
- Wraps the podTemplateOptionsListValues API call with useFetchState
- Accepts kindName, namespace, and optional imageId context
- Uses initialPromisePurity: true to reset state on kind changes, preventing stale data from leaking across kind switches

**WorkspaceForm — two-call architecture**
- allValuesData: fetches all images + pod configs (no imageId context) — used for the Image step, defaults, selectedImage, and summary panel
- filteredValuesData: fetches with imageId context for compatible pod configs — only fires when data.imageConfig is set, gated by kindName to avoid redundant calls
- Defaults (imageConfig/podConfig) are applied once from allValuesData when it first loads for a kind
- Loading spinners shown on Image and Pod Config steps while data loads

**WorkspaceFormSummaryPanel**
- Added imageValues and podConfigValues props for redirect target resolution (previously read from selectedKind inline values)

**Removed deselection useEffect from Image/PodConfig selection components**
- The effect that called onSelect(undefined) when a selected item was hidden by filters caused an infinite loop with the defaults effect (deselect → re-apply default → deselect)
- Selection now persists even when the card is hidden by UI filters — the ID remains valid, the card just isn't rendered
- Better [UX](https://www.patternfly.org/patterns/filters/#filtering-demos) 

**Mock data**
- mockNotebookServiceData.ts: added mockListValuesImages (with CUDA variant) and mockListValuesPodConfigs (with GPU pod config)
- mockNotebookApis.ts: handler now inspects imageConfig.id from request body and filters GPU pod configs for non-GPU images

**Cypress tests**
- Added POST listvalues endpoint to interceptApi type system
- Created interceptListValues helper with dynamic route handler that filters images by imageConfig.id from request body
- Added interceptListValues calls to all 14 Cypress test files that navigate past the Kind step
- Added 3 new listValues test cases in createWorkspace.cy.ts (all options visible, CPU-only options, default auto-selection)

**Unit tests**
- 5 tests for usePodTemplateOptionsListValues hook (API unavailable, kindName undefined, successful call, imageId context, no namespace)

### DIFF
--- a/workspaces/frontend/config/cspell-ignore-words.txt
+++ b/workspaces/frontend/config/cspell-ignore-words.txt
@@ -18,3 +18,6 @@ unhover
 unhovering
 toleration
 tolerations
+podtemplate
+listvalues
+cuda
--- a/workspaces/frontend/src/__tests__/cypress/cypress/support/commands/api.ts
+++ b/workspaces/frontend/src/__tests__/cypress/cypress/support/commands/api.ts
@@ -2,6 +2,7 @@ import type { UserSettings } from 'mod-arch-core';
 import type {
   ApiErrorEnvelope,
   ApiNamespaceListEnvelope,
+  ApiPodTemplateOptionsEnvelope,
   ApiPVCCreateEnvelope,
   ApiPVCListEnvelope,
   ApiSecretCreateEnvelope,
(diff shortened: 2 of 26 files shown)


workspaces/frontend/src/shared/api/apiUtils.ts
```diff
@@ -60,6 +60,7 @@ export function extractConflictCauses(
 
 export function formatValidationErrorMessages(errorEnvelope: ApiErrorEnvelope): string[] {
   const errors = extractValidationErrors(errorEnvelope) ?? [];
+  console.log('errors', errors);
```
paulovmr: Please consider removing this `console.log`

paulovmr (changes requested):

paulovmr (approved):
/lgtm
/approve

PR #1100, opened 2026-05-18
chore: note reason for the role aggregation structure

christian-heusel:
The reason for this has caused multiple people confusion by now which means that we should most likely note it in the code itself. Additionally this way we guard ourselves against reverting / "simplifying" this on accident in the future.

### DIFF
--- a/workspaces/controller/manifests/kustomize/base/manager/user_cluster_roles.yaml
+++ b/workspaces/controller/manifests/kustomize/base/manager/user_cluster_roles.yaml
@@ -1,3 +1,9 @@
+##
+## WARNING: Kubernetes will REMOVE any existing content of `rules` when using `aggregationRule`. 
+##          So, we define both the `aggregate-to-kubeflow-workspaces-edit` and `kubeflow-workspaces-edit` 
+##          even though only `kubeflow-workspaces-edit` is indented to be used in a binding.
+##          https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles
+##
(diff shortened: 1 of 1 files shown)


andyatmiami (approved):
/lgtm

this note helps explain _why this is structured the way it is_ - and important to call out as it caused our working group a little confusion/rework 💯

workspaces/controller/manifests/kustomize/base/manager/user_cluster_roles.yaml
```diff
@@ -1,3 +1,13 @@
+# NOTE: We need to separate out actual rules from aggregation rules due to the
+# following:
+#
+# > "The control plane overwrites any values that you manually specify in the
+# > rules field of an aggregate ClusterRole."
+#
+# https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles
+#
+# The organization of the rules itself is inspired by what kubeflow/pipelines
+# does regarding this.
```
thesuperzapper:
```suggestion
##
## WARNING: Kubernetes will REMOVE any existing content of `rules` when using `aggregationRule`. 
##          So, we define both the `aggregate-to-kubeflow-workspaces-edit` and `kubeflow-workspaces-edit` 
##          even though only `kubeflow-workspaces-edit` is indented to be used in a binding.
##          https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles
##
```

thesuperzapper (changes requested):

PR #1101, opened 2026-05-19
chore: remove gaudi example images for now

thesuperzapper:
superseeds: https://github.com/kubeflow/notebooks/pull/1004

This PR removes the gaudi example images:
- `ghcr.io/kubeflow/kubeflow/notebook-servers/jupyter-pytorch-gaudi`
- `ghcr.io/kubeflow/kubeflow/notebook-servers/jupyter-pytorch-gaudi-full`

This is not a decision to not support gaudi, and we welcome updates to this image (the structure should actually allow any updates to be quite simple), but we need to remove it for now as the builds are failing and we have no way to test this image.

We do this change in a single PR so that its easier to see which code was removed, and revert it in future.


christian-heusel (approved):

thesuperzapper:
Tests are running due to spawner config changes in jupyter web app folder, should be good to merge without waiting for them (but prow might anyway).

/approve

Issue #1102, opened 2026-05-19
[TASK] Un-namespaced WorkspaceKind API call fails for non-admins when editing a workspace


paulovmr:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

When editing a workspace in a manifest-based (with auth) deployment, the frontend makes an un-namespaced call (missing namespace query parameter) to the WorkspaceKind backend API to fetch info. This call fails for non-admin users due to authorization.

Additionally, after the error occurs, the UI appears to crash entirely, requiring a full page refresh — which itself is unexpected behavior. The error should be handled gracefully rather than causing a crash.

  Steps to reproduce:
  1. Deploy with manifest-based auth enabled
  2. Log in as a non-admin user 
  3. Attempt to edit an existing workspace
  4. Observe the failed API call (missing namespace query param) and subsequent UI crash

  Expected behavior:
  - The API call should include the namespace query parameter
  - Even if the call fails, the UI should handle the error gracefully without crashing

### Acceptance Criteria

- [ ] WorkspaceKind info API call includes the correct namespace query parameter when editing a workspace
- [ ] UI handles API errors gracefully (no full crash/blank screen)
- [ ] Verified with a non-admin user in an auth-enabled deployment

andyatmiami:
/close

delivered in #1106

PR #1106, opened 2026-05-19
fix: pass namespace to list workspacekinds on workspace edit

andyatmiami:
Closes: #1102 

When editing a Workspace via the UI, non-admin users received a 403 error. The root cause was that useWorkspaceFormData called listWorkspaceKinds({}) directly — without the namespaceFilter query parameter. Without namespaceFilter, the backend checks for cluster-level "list WorkspaceKinds" permission instead of the namespace-scoped "create Workspaces" permission that regular users have.

The create flow worked because WorkspaceFormKindSelection fetches kinds through useWorkspaceKinds(namespace), which always includes namespaceFilter. The edit flow bypassed that code path entirely.

Rather than just adding namespaceFilter to the raw API call, this fix replaces the direct listWorkspaceKinds({}) call with the shared useWorkspaceKinds(namespace) hook. Both the edit flow (useWorkspaceFormData) and the create flow (WorkspaceFormKindSelection) now go through the same code path, so they cannot diverge on authorization behavior.

### DIFF
--- a/workspaces/frontend/src/app/hooks/useWorkspaceFormData.ts
+++ b/workspaces/frontend/src/app/hooks/useWorkspaceFormData.ts
@@ -24,7 +24,7 @@ const useWorkspaceFormData = (args: {
 }): FetchState<WorkspaceFormData> => {
   const { namespace, workspaceName, workspaceKindName } = args;
   const { api, apiAvailable } = useNotebookAPI();
-  const [workspaceKinds, workspaceKindsLoaded] = useWorkspaceKinds(namespace);
+  const [workspaceKinds, workspaceKindsLoaded, workspaceKindsError] = useWorkspaceKinds(namespace);
 
   const call = useCallback<FetchStateCallbackPromise<WorkspaceFormData>>(async () => {
     if (!apiAvailable) {
@@ -33,6 +33,10 @@ const useWorkspaceFormData = (args: {
 
     if (!namespace || !workspaceName || !workspaceKindName) {
       return EMPTY_FORM_DATA;
+    }
+
+    if (workspaceKindsError) {
+      return Promise.reject(workspaceKindsError);
     }
 
     if (!workspaceKindsLoaded) {
@@ -80,6 +84,7 @@ const useWorkspaceFormData = (args: {
     workspaceKindName,
     workspaceKinds,
     workspaceKindsLoaded,
+    workspaceKindsError,
   ]);
 
   return useFetchState(call, EMPTY_FORM_DATA);
--- a/workspaces/frontend/src/app/hooks/useWorkspaceFormData.ts
+++ b/workspaces/frontend/src/app/hooks/useWorkspaceFormData.ts
@@ -1,6 +1,7 @@
 import { useCallback } from 'react';
-import { FetchState, FetchStateCallbackPromise, useFetchState } from 'mod-arch-core';
+import { FetchState, FetchStateCallbackPromise, useFetchState, NotReadyError } from 'mod-arch-core';
 import { useNotebookAPI } from '~/app/hooks/useNotebookAPI';
+import useWorkspaceKinds from '~/app/hooks/useWorkspaceKinds';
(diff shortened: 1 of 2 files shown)


andyatmiami:
ℹ️ 

After this fix, listWorkspaceKinds is called twice in the edit flow:                                                                                                                                                                         
                                                                                                                                                                                                                                             
1. useWorkspaceFormData — calls useWorkspaceKinds(namespace) on form mount to resolve the full WorkspaceKind object for populating initial form data                                                                                         
2. WorkspaceFormKindSelection — calls useWorkspaceKinds(namespace) when step 1 renders (and again on each re-visit) for the selection UI                                                                                                     
                                                                                                                                                                                                                                             
Both go through the same code path (the bug fix), but they're separate hook instances making separate API calls.                                                                                                                             
                                                                                                                                                                                                                                             
Why It Exists                                                                                                                                                                                                                                
                                                                                                                                                                                                                                             
WorkspaceFormData.kind stores the full WorkspacekindsWorkspaceKindListItem object — not just the name — because the form needs the kind's nested data (available images, pod configs, volume mount paths, defaults) at every subsequent step. The workspace API only returns the kind name, so a separate fetch is needed to resolve the full object.                                                                                                                                     
                                                                                                                                                                                                                                             
How to Eliminate It                                                                                                                                                                                                                          
                                                                                                                                                                                                                                             
PR #1098 moves image and pod config options out of the WorkspaceKind object and into a dedicated listValues API. Once that merges, the form's dependency on the full kind object shrinks significantly. At that point:                       
                                                                                                                                                                                                                                             
- WorkspaceFormData could track just the kind name (which the workspace payload already provides) instead of the full object                                                                                                                 
- The full kind object would only be needed in WorkspaceFormKindSelection for the selection UI                                                                                                                                               
- useWorkspaceFormData would no longer need to call useWorkspaceKinds at all — back to a single API call                                                                                                                                     
                                                                                                                                                                                                                                             
This aligns with the direction of treating the kind name as a reference key and looking up the full object only where the UI needs it.

workspaces/frontend/src/app/hooks/useWorkspaceFormData.ts
```diff
@@ -23,6 +24,7 @@ const useWorkspaceFormData = (args: {
 }): FetchState<WorkspaceFormData> => {
   const { namespace, workspaceName, workspaceKindName } = args;
   const { api, apiAvailable } = useNotebookAPI();
+  const [workspaceKinds, workspaceKindsLoaded] = useWorkspaceKinds(namespace);
```
thaorell: Not blocking, but consider 
`[workspaceKinds, workspaceKindsLoaded, workspaceKindsError]`

workspaces/frontend/src/app/hooks/useWorkspaceFormData.ts
```diff
@@ -33,12 +35,13 @@ const useWorkspaceFormData = (args: {
...
     }
 
-    const [workspaceEnvelope, workspaceKindsEnvelope] = await Promise.all([
-      api.workspaces.getWorkspace(namespace, workspaceName),
-      api.workspaceKinds.listWorkspaceKinds({}),
-    ]);
+    if (!workspaceKindsLoaded) {
+      return Promise.reject(new NotReadyError('Workspace kinds not yet available'));
+    }
+
```
thaorell, line 45: Similarly here, 
`if (workspaceKindsError) {
  Throw some error
}`

thaorell (commented):
Left some non-blocking comments. Please let me know if they make sense

workspaces/frontend/src/app/hooks/useWorkspaceFormData.ts
andyatmiami: this seems like great advice - will handle this (and the related 2 following comments)

thaorell (approved):
/lgtm

paulovmr (approved):
/lgtm
/approve

PR #1107, opened 2026-05-19
fix: pass namespace query param when listing storage classes

andyatmiami:
ℹ️ _NO GH ISSUE_

PR #1097 added a `namespace` query parameter to the backend's GET /storageclasses endpoint. When provided, the backend checks namespace-scoped "create PersistentVolumeClaims" permission instead of cluster-level "list StorageClasses" — allowing regular (non-admin) users to access storage classes within their namespace.

The frontend's useStorageClasses hook was calling listStorageClasses() without the namespace parameter, causing non-admin users to receive a 403 when opening the volume creation or attachment modals in the Workspace Wizard (both Create and Edit flows).

This fix mirrors the approach taken in PR #1106 for workspace kinds: the useStorageClasses hook now accepts an optional namespace argument and passes it as a query parameter to the API. Both VolumesCreateModal and VolumesAttachModal pass selectedNamespace from the namespace selector context.

swagger.version is updated from c9a8a7c6 to 142a30f0 (the PR #1097 commit) so that the generated API client includes the new namespace query parameter on listStorageClasses.


thaorell (approved):
/lgtm

paulovmr (approved):
/lgtm
/approve

christian-heusel (commented):
Also adding to this, I verified this in a environment with auth enabled and can attest that the frontend now calls the right routes and auth succeeds:

```json
{
    "data": [
        {
            "name": "premium-local",
            "displayName": "Premium Local (Local Path)",
            "description": "Simulated premium storage for development. Not enabled for use.",
            "canUse": false
        },
        {
            "name": "standard",
            "displayName": "Standard (Local Path)",
            "description": "Local path provisioner for development. Data is stored on the node and not replicated.",
            "canUse": true
        }
    ]
}

```

http://localhost:8082/workspaces/api/v1/storageclasses?namespace=kubeflow-user-example-com

PR #1110, opened 2026-05-20
feat: add auth in debug menu for developing

andyatmiami:
ℹ️ _NO GH ISSUE_

Backend auth is enabled by default in the Tilt dev environment
(`DISABLE_AUTH=false`). Contributors need a way to set the `kubeflow-userid`
and `kubeflow-groups` request headers from the frontend UI, and the dev
cluster needs RBAC and access logging infrastructure to support
end-to-end auth testing.

Frontend (debug auth menu):
- Add `devAuth.ts` utility with localStorage helpers and Axios interceptor
- Add `DebugAuthSection` component with User/Groups form fields
- Replace Debug page placeholder with the new auth section
- Register interceptor on all API instances in `notebookApisImpl`
- Add unit tests for all devAuth utility functions
- Guard Debug page and nav item behind `DEV_MODE` flag

Dev infrastructure (developing/):
- Add RBAC manifests for two dev users: "`admin`" (cluster-wide via
  `kubeflow-workspaces-admin` + supplement) and "`user`" (namespace-scoped
  in default via `kubeflow-workspaces-edit` + supplement), using the
  controller-defined ClusterRoles for realistic Kubeflow RBAC behavior
- Add Istio access logging with structured JSON output including
  kubeflow-userid and kubeflow-groups headers, using MeshConfig
  extension provider and Telemetry API
- Extract istioctl install flags into IstioOperator values file
  (`developing/manifests/istio-install-values.yaml`)
- Move `kind.yaml `from scripts/ to manifests/ alongside other config
- Update `Tiltfile` with dev-rbac and istio-gateway-logging resources
- Document RBAC users and access logging in `DEVELOPMENT_GUIDE.md`

⚠️ While a `Groups` field was added to the form - this is presently untested behavior (even from a `backend/` perspective) - so future follow up will be done to ensure `kubeflow-groups` behaves correctly in `tilt`


christian-heusel (commented):
Hey @andyatmiami thanks a ton for staring to tackle this! 🔥  

I took a brief look at this (knowing it is still in a draft state) and have the following two comments as of now (which are unrelated to the code itself):
- It is currently unclear what valid values would be for the two fields, we should either make this a dropdown, suggest values or let developers using this know in a help/hover text
- We should most likely also add documentation to the development guide once everything is wired up to let people know of this new "impersonation" feature

workspaces/frontend/src/shared/utilities/devAuth.ts
```diff
@@ -0,0 +1,50 @@
...
+    return '';
+  }
+};
+
+export const setDevAuth = (user: string, groups: string): void => {
+  try {
+    localStorage.setItem(DEV_AUTH_USER_KEY, user);
+    localStorage.setItem(DEV_AUTH_GROUPS_KEY, groups);
+  } catch {
+    // localStorage may be unavailable
```
paulovmr:
```suggestion
    console.error('localStorage is unavailable');
```

paulovmr (changes requested):

DEVELOPMENT_GUIDE.md
```diff
@@ -144,6 +144,40 @@ Access the components through the Istio ingress gateway:
...
+To tail the access logs:
+
+```bash
+kubectl logs -n istio-system -l app=istio-ingressgateway -f
+```
+
+Example log entry:
+
+```json
+{"duration":105,"kubeflow_groups":null,"kubeflow_userid":"admin","method":"GET","path":"/workspaces/api/v1/workspaces/default","response_code":200,"timestamp":"2026-05-21T14:12:11.965Z","upstream":"10.244.1.9:4000"}
```
christian-heusel: I think to showcase the contained information within the log it would make more sense to format this log entry like this:
```suggestion
{
    "duration": 105,
    "kubeflow_groups": null,
    "kubeflow_userid": "admin",
    "method": "GET",
    "path": "/workspaces/api/v1/workspaces/default",
    "response_code": 200,
    "timestamp": "2026-05-21T14:12:11.965Z",
    "upstream": "10.244.1.9:4000"
}
```

developing/scripts/setup-istio.sh
```diff
@@ -24,7 +24,9 @@ if "${ISTIOCTL}" verify-install >/dev/null 2>&1; then
   echo "Istio is already installed"
 else
   echo "Installing Istio with default profile and CNI plugin..."
-  "${ISTIOCTL}" install --set profile=default --set components.cni.enabled=true -y
+  "${ISTIOCTL}" install \
+    -f "${DEVELOPING_DIR}/manifests/istio-install-values.yaml" \
+    -y
```
christian-heusel: I prefer the longopts here for readability:
```suggestion
    --filename "${DEVELOPING_DIR}/manifests/istio-install-values.yaml" \
    --skip-confirmation
```

developing/scripts/setup-kind.sh
```diff
@@ -6,7 +6,8 @@ set -euo pipefail
 
 CLUSTER_NAME="tilt"
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-KIND_CONFIG="${SCRIPT_DIR}/kind.yaml"
+DEVELOPING_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+KIND_CONFIG="${DEVELOPING_DIR}/manifests/kind.yaml"
```
christian-heusel: It's a bit weird how we define those paths (even though I realize that this is the way we did it beforehand, I think a cleaner way would be something like `realpath developing/scripts`, potentially we could modify this to drop the `cd && pwd` pattern. Also `SCRIPT_DIR` is unused with this change but it's a standard definition we seem to put in our scripts, so we could perhaps collapse it into the following:
```suggestion
SCRIPT_DIR="$(realpath "$(dirname "${BASH_SOURCE[0]}")")"
KIND_CONFIG="${SCRIPT_DIR}/../manifests/kind.yaml"
```

developing/manifests/rbac/user.yaml
```diff
@@ -0,0 +1,67 @@
...
+metadata:
+  name: dev-user-resources
+  namespace: default
+rules:
+  - apiGroups: [""]
+    resources: [persistentvolumeclaims]
+    verbs: [create, delete, get, list]
+  - apiGroups: [""]
+    resources: [secrets]
+    verbs: [create, delete, get, list, update]
```
christian-heusel: I think in the other manifests we usually unroll the arrays here:
https://github.com/kubeflow/notebooks/blob/24ce51e5b19adcef194aa989e9df6bbee3b3c1f9/workspaces/controller/manifests/kustomize/base/manager/role.yaml#L6-L16

This would also apply in all other places where this kind of structure is used on the PR 🤗

workspaces/frontend/src/app/pages/Debug/DebugAuthSection.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+    if (!trimmedUser) {
+      setError('User is required.');
+      return;
+    }
+
+    setError(null);
+    setDevAuth(trimmedUser, trimmedGroups);
+    notification.success('Auth headers saved.');
+  };
```
christian-heusel, line 33: AI Review comment 🤖

(I can't verify / falsify it due to lack of frontend knowledge, however I thought I'd leave it here nonetheless)

> **UX: Input fields don't sync to trimmed values after save**
>
> `handleSave` trims before calling `setDevAuth`, but the React state (user, groups) still holds the original untrimmed string. If a user types admin (with spaces), the input will still show the padded value after saving even though admin is what's persisted and used.
>
> **Fix:** call `setUser(trimmedUser)` and `setGroups(trimmedGroups)` in `handleSave` after saving.

workspaces/frontend/src/app/pages/Debug/DebugAuthSection.tsx
```diff
@@ -0,0 +1,88 @@
...
+            helperTextNode={<HelperText>Comma-separated list of group IDs.</HelperText>}
+          >
+            <TextInput
+              id="debug-auth-groups"
+              data-testid="debug-auth-groups-input"
+              value={groups}
+              onChange={(_event, value) => {
+                setGroups(value);
+                setError(null);
+              }}
```
christian-heusel, line 75: AI Review comment 🤖

(I can't verify / falsify it due to lack of frontend knowledge, however I thought I'd leave it here nonetheless)

---

> **UX: Error cleared on groups change, but error is only about user**
>
> `onChange` for the groups field calls `setError(null)`, so typing in groups clears "User is required" even though user is still empty. Either only clear the error in the user's `onChange`, or use per-field error state.
>
> In `DebugAuthSection.tsx`, the groups `onChange` handler calls `setError(null)`:
>
>
> ```ts
> onChange={(_event, value) => {
>   setGroups(value);
>   setError(null);  // ← clears the error
> }}
> ```
> 
> But the only validation error that can be set is 'User is required.', which is triggered when the user field is empty. So if someone leaves the user field blank, hits Save (error appears), and then starts typing in the groups field — the error disappears even though user is still empty and would still fail validation on the next save attempt.
> 
> The fix is simply to remove `setError(null)` from the groups `onChange`, since groups has no validation of its own. Keep it only in the user `onChange`, where it's meaningful (typing in the user field is directly addressing the error).

christian-heusel (commented):
Hey andy, really awesome that you're pushing this (as always: everything, everywhere, all at once 🚀 ) because it will greatly improve the developer experience and allow us to catch entire categories of bugs upfront. 🔥

I currently don't have access to my developer machine, so I can only leave the following (minor) feedback so far and my review is code-only 😄

workspaces/frontend/src/app/pages/Debug/DebugAuthSection.tsx
```diff
@@ -0,0 +1,88 @@
...
+import ThemeAwareFormGroupWrapper from '~/shared/components/ThemeAwareFormGroupWrapper';
+import { getDevAuthUser, getDevAuthGroups, setDevAuth } from '~/shared/utilities/devAuth';
+
+export const DebugAuthSection: React.FC = () => {
+  const notification = useNotification();
+
+  const [user, setUser] = useState(() => getDevAuthUser());
+  const [groups, setGroups] = useState(() => getDevAuthGroups());
+  const [error, setError] = useState<string | null>(null);
+
```
thaorell, line 19: Consider using `useBrowserStorage` hook like in `WithValidImage.tsx` to interact with browser's local storage. That could eliminate the need for the helper functions

thaorell (commented):
Thank you for tackling this, @andyatmiami

developing/scripts/setup-kind.sh
andyatmiami: my own personal preferences (worded through Claude)
  
  cd && pwd is the more portable/robust choice:
  - POSIX-compliant everywhere, always available as a shell builtin
  - Works on macOS/BSD out of the box
  - No external dependencies
  
  realpath is cleaner but has a portability gap:
  - Not available by default on macOS (which matters for this project — contributors develop on macOS)
  - External command, not a builtin
  - Only standardized since POSIX.1-2008
  
  I'm going to ignore this suggestion in the context of this PR - but happy to discuss it more in general - as this pattern (my bad 😎 ) appears throughout multiple scripts - and if we want to change.. we should change at once across all call sites.

workspaces/frontend/src/app/pages/Debug/Debug.tsx
```diff
@@ -1,30 +1,17 @@
...
-      <EmptyStateBody>
-        This represents an the empty state pattern in Patternfly 6. Hopefully it&apos;s simple
-        enough to use but flexible enough to meet a variety of needs.
-      </EmptyStateBody>
-      <EmptyStateFooter>
-        <Button variant="primary">Primary Action</Button>
-      </EmptyStateFooter>
-    </EmptyState>
-  </PageSection>
+  <>
```
thaorell: We should hide it or render `<NotFound />` when DEV_MODE isn't set or false

```
{
   DEV_MODE ?
       <>
           <PageSection data-testid="debug-page">
               <Title headingLevel="h1">Debug Settings</Title>
            </PageSection>
            <PageSection>
               <DebugAuthSection />
            </PageSection>
    </> : 
    <NotFound />
}
```

thaorell (approved):
/lgtm

paulovmr (approved):
/lgtm
/approve

developing/istio-developing.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: install.istio.io/v1alpha1
```
thesuperzapper, line 1: Lets move this file directly under the `developing/` folder rather than manifests, similar to the `testing/istio-cni.yaml`.

We could probably also call it `istio-developing.yaml`.

developing/scripts/setup-kind.sh
thesuperzapper: Lets also move this `kind.yaml` under the `developing/` folder directly, and rename to `kind-1-35.yaml` to show which version.

developing/Tiltfile
```diff
@@ -62,6 +62,37 @@ k8s_resource(
...
+        "dev-user-global-read:clusterrole",
+        "dev-user-global-read:clusterrolebinding",
+        "dev-user-resources:role",
+        "dev-user-resources:rolebinding",
+    ],
+    new_name="dev-rbac",
+    resource_deps=["workspaces-controller"],
+    pod_readiness="ignore",
+    labels=["rbac"],
+)
```
thesuperzapper, line 110: Is the reason this is here to ensure the role aggregation has finished? Or something else?

developing/Tiltfile
```diff
@@ -94,7 +125,7 @@ for o in backend_objects:
             "SWAGGER_SCHEME": "https",
             "SWAGGER_HOST": "localhost:8443",
             "SWAGGER_BASE_PATH": "/workspaces/api/v1",
-            "DISABLE_AUTH": "true",
+            "DISABLE_AUTH": "false",
```
thesuperzapper: Probably mark with TODO to ensure we remember to remove once we remove `DISABLE_AUTH` from backend.

DEVELOPMENT_GUIDE.md
```diff
@@ -144,6 +144,49 @@ Access the components through the Istio ingress gateway:
...
 
+### Tilt - Authentication & RBAC
+
+Tilt deploys RBAC bindings for two dev users from [`developing/manifests/rbac/`](developing/manifests/rbac/).
+The backend authenticates requests via `kubeflow-userid` and `kubeflow-groups` headers, then authorizes each API call with a Kubernetes [SubjectAccessReview](https://kubernetes.io/docs/reference/access-authn-authz/authorization/#checking-api-access).
+
+| User | Scope | Description |
+|------|-------|-------------|
+| `admin` | Cluster-wide | Bound to `kubeflow-workspaces-admin` (workspaces) plus a supplement for WorkspaceKinds, PVCs, secrets, and namespaces. Full access to all resources. |
+| `user` | `default` namespace | Bound to `kubeflow-workspaces-edit` in `default` plus a supplement for PVCs and secrets. Cannot manage WorkspaceKinds or access other namespaces. |
```
thesuperzapper: We can probably simplify these to just be descriptions which say "typical workspaces user permissions", and "typcial cluster admin permissions".

With some YAML link or way to find the specifics (e.g. kubectl command)?

thesuperzapper (changes requested):

developing/Tiltfile
andyatmiami: exactly - the key bit here is this:
- `resource_deps=["workspaces-controller"]`

i.e. making sure the `workspaces-controller` manifests have been applied...

christian-heusel (approved):
Thanks a lot for pushing this through, this will ease development a lot! 🔥 

I reviewed the changes again and they look good to me, all other things that arise can come up can be addressed in follow-up PRs I guess 🤗

thesuperzapper:
Thanks @andyatmiami great to have this!

/approve
/lgtm

PR #1113, opened 2026-05-20
ci: automatically approve test for org members

christian-heusel:
The idea behind this is that they could set the 'ok-to-test' label anyways via prow and it eases the burden of the maintainers having to do this manually for org members.

The code is taken from the way that `kubeflow/sdk` does it for their equivalent job.

Link: https://github.com/kubeflow/sdk/blob/main/.github/workflows/gh-workflow-approve.yaml


thesuperzapper:
Thanks!

/lgtm
/approve

PR #1115, opened 2026-05-21
chore: remove RStudio references from the codebase

manaswinidas:
RStudio is no longer a supported workspace kind. This removes all references including the sample WorkspaceKind manifest, CRD examples, Go type annotations, cypress test mock data, and README mention.


christian-heusel (changes requested):
> RStudio is no longer a supported workspace kind.

Where did you get that idea from? 🤔 Please reference an issue by one of the Kubeflow Notebooks Maintainers that confirm this deprecation, because so far I didn't hear about it and I'm fairly active in the working group.

The images are still being built:

https://github.com/kubeflow/notebooks/blob/17ab7e113cce2facc76947df36339bbedb00a4f8/.github/workflows/example_notebook_servers_publish.yaml#L33-L42

Additionally, the example images are always to be understood as _example_ images and people can create custom images if they'd like to.

/hold

manaswinidas:
Yeah apologies and thanks @christian-heusel 

I'm closing the PR

PR #1118, opened 2026-05-22
ci: define basic e2e tests

andyatmiami:
Add a self-contained Cypress e2e test suite under testing/e2e/ that validates core user workflows against a real Kubernetes cluster. The suite runs two specs: a user creating a Workspace through the wizard, and an admin creating a WorkspaceKind via YAML upload.

Kubernetes environment setup (namespace, RBAC, seed data) is handled programmatically via @kubernetes/client-node in Cypress plugin tasks rather than shell scripts and static YAML manifests. A global before() hook calls setupE2e() before any spec runs, making the test project self-contained with all dependencies co-located. The JupyterLab WorkspaceKind seed data references the controller's sample YAML directly to avoid maintaining duplicate fixtures.

Authentication is simulated by injecting the kubeflow-userid header via cy.intercept(), with separate loginAsAdmin and loginAsUser helpers that map to distinct RBAC policies (cluster-scoped admin vs namespace-scoped user).

Custom Cypress commands (k8sGet, k8sDelete, k8sWaitForResource) provide a thin wrapper over the Kubernetes API for in-test assertions and cleanup. Each spec deletes resources it creates in afterEach() hooks.

Page objects encapsulate UI interactions for the workspaces list, workspace creation wizard, workspace kinds list, and workspace kind creation form.

The CI workflow deploys all three components (controller, backend, frontend) to a Kind cluster with Istio and cert-manager, runs a sanity check to verify endpoints respond, then executes Cypress with the gateway port-forwarded. The frontend is built in standalone deployment mode via .env.production.local so it serves without the central dashboard. Test artifacts (videos, screenshots) are uploaded on failure.

The Makefile provides an `e2e` convenience target that chains setup-cluster, deploy-all, sanity-check, and local-e2e for local development.


.github/workflows/ws-e2e-test.yml
```diff
@@ -11,32 +11,52 @@ on:
...
+
+      - name: Upload test artifacts
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        if: failure()
+        with:
+          name: e2e-test-results
+          path: |
+            testing/e2e/cypress/videos/
+            testing/e2e/cypress/screenshots/
+          retention-days: 7
```
christian-heusel, line 62: Do we need to upload _all_ videos and screenshot in the case of a test failure? 🤔

.github/workflows/ws-e2e-test.yml
```diff
@@ -11,32 +11,52 @@ on:
...
     paths:
-      - workspaces/**
+      - 'workspaces/**'
+      - 'testing/**'
 
 jobs:
   e2e-test:
     runs-on: ubuntu-latest
+    env:
+      TERM: xterm-256color
```
christian-heusel, line 21: What is this used for any why do we need it? 🤔 Please add a comment that explains why it is there in case that it is important 😅

testing/e2e/package.json
christian-heusel, line 1: This file should most likely also be added to the dependabot config in order to ensure that it stays up-to-date:
https://github.com/kubeflow/notebooks/blob/9bd04005e13b18c1a6cb4aa25cd78c677a648eeb/.github/dependabot.yml#L46

testing/Makefile
```diff
@@ -124,3 +134,5 @@ check-kubectl: ## Verify that kubectl is available in PATH.
 clean: ## Remove downloaded tool binaries.
 	rm -rf $(LOCALBIN)
 	@echo "INFO: '$(LOCALBIN)' successfully cleaned."
+	rm -rf $(FRONTEND_ENV_LOCAL)
```
christian-heusel, line 137: Since this is not a directory we can also drop the `-r` argument (recursive `rm` is scary 👻 ):
```suggestion
	rm -f $(FRONTEND_ENV_LOCAL)
```

testing/scripts/run-e2e.sh
```diff
@@ -0,0 +1,44 @@
...
+  if curl -sk "https://localhost:${GATEWAY_PORT}/workspaces/api/v1/healthcheck" >/dev/null 2>&1; then
+    echo "✓ Gateway port-forward ready"
+    break
+  fi
+  if [[ $i -eq 10 ]]; then
+    echo "✗ ERROR: Gateway port-forward failed to become ready"
+    exit 1
+  fi
+  sleep 2
+done
```
christian-heusel, line 33: Same review comment as on the other PR, we can do this part purely in curl:
```bash
if ! curl --silent --output /dev/null --fail \                                                          
     --retry 10 --retry-delay 2 --retry-all-errors \                                                    
     "http://localhost:${local_port}${probe_path}"; then                                                
  echo "✗ ERROR: ${label} endpoint not reachable after 10 attempts"                                     
  return 1                                                                                              
fi
```

https://github.com/kubeflow/notebooks/pull/763#discussion_r3098845984

testing/e2e/cypress/tests/user-creates-workspace.cy.ts
```diff
@@ -0,0 +1,63 @@
...
+    // Assert: redirected to workspaces list, new workspace is visible
+    cy.url().should('not.include', '/create');
+    workspacesPage.assertWorkspaceExists(WORKSPACE_NAME);
+
+    // Assert: Workspace CR exists in the cluster
+    cy.k8sGet(K8S_PARAMS).then((ws) => {
+      expect(ws.metadata.name).to.equal(WORKSPACE_NAME);
+      expect(ws.metadata.namespace).to.equal(NAMESPACE);
+      expect(ws.spec.kind).to.equal('jupyterlab');
+    });
```
christian-heusel, line 61: How does this part of the code handle waiting for the resource? 🤔 Maybe I'm misunderstanding how cypress works, but as far as I understand there is a chance for a flake/race here given that the K8S api takes a bit to process the workspace we submitted 🤔

testing/e2e/cypress/pages/createWorkspace.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  // Step 2: Image selection — selects the first visible card
+  selectFirstImage(): void {
+    cy.get('[class*="pf-v6-c-card"][id]').first().click();
+  }
+
+  // Step 3: Pod config selection — selects the first visible card
+  selectFirstPodConfig(): void {
+    cy.get('[class*="pf-v6-c-card"][id]').first().click();
+  }
```
christian-heusel, line 31: I think that is a rather brittle select 😅 Maybe we can match on something like `data-testid="kind-card-*"` especially once the changes from #1344 lands this is also atleast guaranteed to be the same one each time 🤔

testing/e2e/cypress/pages/createWorkspaceKind.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  uploadYamlFile(filePath: string): void {
+    cy.readFile(filePath, 'utf-8').then((content: string) => {
+      this.findFileInput().selectFile(
+        {
+          contents: Cypress.Buffer.from(content),
+          fileName: 'workspace-kind.yaml',
+          mimeType: 'application/x-yaml',
+        },
+        { force: true },
```
christian-heusel, line 26: What does `force: true`  mean in this context and why do we need it? 🤔

testing/e2e/cypress/pages/workspaces.ts
```diff
@@ -0,0 +1,30 @@
+class WorkspacesPage {
+  visit(): void {
+    cy.visit('/');
+  }
+
+  selectNamespace(namespace: string): void {
+    cy.get('.kubeflow-u-namespace-select').within(() => {
+      cy.get('button').first().click();
+    });
+    cy.get('.pf-v6-c-menu__list-item').contains(namespace).click();
```
christian-heusel, line 10: We should also give them a proper id, this does not seem right to me 😅

testing/e2e/cypress/support/tasks/k8s-client.ts
```diff
@@ -0,0 +1,112 @@
...
+  return kc.makeApiClient(k8s.CustomObjectsApi);
+}
+
+function isNotFoundError(err: unknown): boolean {
+  if (err instanceof Error && err.message.includes('HTTP-Code: 404')) {
+    return true;
+  }
+  const httpErr = err as { statusCode?: number };
+  return httpErr.statusCode === 404;
+}
```
christian-heusel, line 27: I think we could also lean on the more specific upstream type here:
https://github.com/kubernetes-client/javascript/blob/d8ac4ace456815b5668706971e89c510bbaf40e3/src/gen/apis/exception.ts#L6-L7

testing/e2e/cypress/support/tasks/k8s-client.ts
```diff
@@ -0,0 +1,112 @@
...
+    return resp;
+  }
+  const resp = await api.getClusterCustomObject({
+    group: params.group,
+    version: params.version,
+    plural: params.plural,
+    name: params.name,
+  });
+  return resp;
+}
```
christian-heusel, line 48: I think this can be simplified by using variadic parameter expansion & destructuring (applies to a few other places aswell):
```suggestion
async function k8sGet(params: K8sResourceParams): Promise<object> {
  const api = getClient();
  const { namespace } = params;
  if (namespace) {
    return api.getNamespacedCustomObject({ ...params, namespace });
  }
  return api.getClusterCustomObject(params);
}
```

testing/e2e/cypress/support/tasks/k8s-client.ts
```diff
@@ -0,0 +1,112 @@
...
+    if (!isNotFoundError(err)) {
+      throw err;
+    }
+  }
+  return null;
+}
+
+function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
```
christian-heusel, line 79: Sleep functions are genererally a bit suspicious, why do we need it here and can we replace it with a "native" timeout? 🤔

christian-heusel: However the 🤖 says its a pragmatic way of approaching the issue, so maybe we can just leave it as-is:

Is there an alternative way of implementing k8sWaitForResource() that relies on upstream primitives and timeouts?

> Yes — two upstream primitives worth considering instead of the hand-rolled Date.now() polling loop:
>
> 1. k8s.Watch (from @kubernetes/client-node) opens a watch connection and streams ADDED/MODIFIED events, so you resolve as soon as the resource appears instead of polling every 2s. It's event-driven and typically faster/lighter, but adds real complexity here: you still need an initial GET (the resource may already exist before the watch starts), you have to handle the watch's own reconnect/error semantics, and cluster-vs-namespaced-scoped watching means building a different path string per case — essentially duplicating the k8sGet branching one level down.
> 2. AbortSignal.timeout(ms), passed as abortSignal into the API call options, would let you drop your manual deadline/Date.now() bookkeeping and instead just catch the abort error — but that only replaces the timeout arithmetic, not the poll loop itself, since you'd still need to retry the GET on 404/abort until success.
> 
> My take: the current poll-and-isNotFoundError loop is the pragmatic choice for a test helper — it's ~15 lines, easy to reason about, and "wait up to N seconds for a CRD to exist" isn't performance-sensitive in e2e tests. I'd only reach for Watch if polling interval latency were actually causing flaky/slow tests. Want me to prototype the Watch-based version so you can compare?

testing/e2e/cypress/support/tasks/k8s-setup.ts
```diff
@@ -0,0 +1,329 @@
+import * as k8s from '@kubernetes/client-node';
+import * as fs from 'fs';
+import * as path from 'path';
+
+const E2E_NAMESPACE = 'e2e-test';
+const CONTROLLER_SAMPLES = path.resolve(
+  __dirname,
+  '../../../../../workspaces/controller/manifests/kustomize/samples',
+);
```
christian-heusel, line 9: Maybe it's simpler to somehow figure out the repository root to spare the `../..`? 🤔 

Not sure if I like this though:
```suggestion
function findRepoRoot(startDir: string): string {
  let dir = startDir;
  while (!fs.existsSync(path.join(dir, '.git'))) {
    const parent = path.dirname(dir);
    if (parent === dir) {
      throw new Error(`Could not find repository root above ${startDir}`);
    }
    dir = parent;
  }
  return dir;
}

const CONTROLLER_SAMPLES = path.join(
  findRepoRoot(__dirname),
  'workspaces/controller/manifests/kustomize/samples',
);
```

testing/e2e/cypress/support/tasks/k8s-setup.ts
```diff
@@ -0,0 +1,329 @@
...
+
+function getClients() {
+  const kc = new k8s.KubeConfig();
+  kc.loadFromDefault();
+  return {
+    core: kc.makeApiClient(k8s.CoreV1Api),
+    rbac: kc.makeApiClient(k8s.RbacAuthorizationV1Api),
+    custom: kc.makeApiClient(k8s.CustomObjectsApi),
```
christian-heusel, line 17: Do we perhaps also need https://kubernetes-client.github.io/javascript/api-reference/workloads/AppsV1Api?_highlight=state#stateful-set-1 here? 🤔

testing/e2e/cypress/support/tasks/k8s-setup.ts
```diff
@@ -0,0 +1,329 @@
...
+  return httpErr.statusCode === 409;
+}
+
+function isNotFoundError(err: unknown): boolean {
+  if (err instanceof Error && err.message.includes('HTTP-Code: 404')) {
+    return true;
+  }
+  const httpErr = err as { statusCode?: number };
+  return httpErr.statusCode === 404;
+}
```
christian-heusel, line 35: Same comment as above (https://github.com/kubeflow/notebooks/pull/1118/changes#r3856211135) we could potentially use the same solution here if we like it 😊

christian-heusel (changes requested):
Hey @andyatmiami thanks a lot for driving this forward, I think E2E tests will be more and more important as we move forward! 🚧 

I have left you a few review comments, feel free to implement or challenge as you deem sensible!

Also could you rebase your changes? A lot has changed since you have initially posted this 😅

christian-heusel:
@siyuanfoundation could you take a look at the approach this PR takes to e2e testing? 😊 I saw that you opened https://github.com/kubeflow/notebooks/issues/1418 regarding the controller e2e tests, perhaps you also have opinions for this 😄

thisis-Shitanshu:
Hi @andyatmiami! After going through the framework, I think the PR demonstrates the right E2E path with Cypress.

There is already a useful separation in the Makefile between cluster setup/deployment and `local-e2e`. My main architectural concern is that some environment-specific responsibilities still leak into the product E2E runner itself.

For example:

- `local-e2e` currently requires a kind context
- `run-e2e.sh` assumes an Istio ingress gateway and builds the test URL through a local port-forward
- the Cypress global `before()` calls `setupE2e()`, which creates the namespace, RBAC/identities, ServiceAccount, PVC, and baseline WorkspaceKind

So there is effectively an environment contract today, but it is implicit across the Makefile, runner, Cypress setup, and tests.

I wonder if we could make that handoff explicit and let the reusable product suite assume a **reachable, prepared Workspaces deployment**. A small environment profile could describe things such as:

- base URL
- test namespace
- authentication strategy / test identities
- known baseline WorkspaceKind
- optional storage/resource fixtures

The product tests would consume those capabilities without needing to know how the environment created them.

Conceptually, I see the responsibilities as:

**Environment preparation**
- cluster/deployment exists
- routing is reachable
- identities / namespace / RBAC are configured
- baseline WorkspaceKind / storage / resources are available

**Workspaces product E2E**
- authenticate as user/admin
- exercise canonical Workspace / WorkspaceKind workflows
- wait for domain states such as Created / Ready / Deleted
- connect and verify actual Workspace usability
- optionally use Kubernetes access for stronger assertions and diagnostics

This would let kind/Tilt/KCD or other downstream environments prepare themselves differently while still running the same Workspaces behavioral contract.

I also think ordinary test authors should mostly work with Workspaces domain helpers such as `workspace.waitForReady()` / `workspace.waitForDeleted()` rather than generic CRD parameters. Those helpers can still use the existing bounded Kubernetes polling underneath.

This separation would also help with future isolation and parallelism. I believe the current fixed namespace and shared cluster-scoped fixtures could eventually run into the same kind of shared-state contention described in the controller E2E suite in #1418.

If this direction makes sense, I'd be happy to prototype the boundary on my review branch. That should help us see the smallest practical refactor, and then we can decide whether it fits best in this PR or as a prerequisite change.

PR #1131, opened 2026-05-27
test: increase e2e test timeout

siyuanfoundation:
increase the timeout in controller e2e test from 1min to 3min.

When the container image is not cached, it could take more than 2min to get the pod running.

### DIFF
--- a/workspaces/controller/test/e2e/e2e_test.go
+++ b/workspaces/controller/test/e2e/e2e_test.go
@@ -50,7 +50,7 @@ const (
 	curlImage = "curlimages/curl:8.9.1"
 
 	// how long to wait in "Eventually" blocks
-	timeout = time.Second * 60
+	timeout = time.Second * 180
 
 	// how long to wait in "Consistently" blocks
 	duration = time.Second * 10 //nolint:unused
@@ -247,7 +247,9 @@ var _ = Describe("controller", Ordered, func() {
 				}
 				return nil
 			}
+			startTime := time.Now()
 			Eventually(verifyWorkspaceState, timeout, interval).Should(Succeed())
+			_, _ = fmt.Fprintf(GinkgoWriter, "workspace reached 'Running' state in %v\n", time.Since(startTime))
 
 			By("validating that the workspace pod is running as expected")
 			verifyWorkspacePod := func(g Gomega) {


workspaces/controller/test/e2e/e2e_test.go
```diff
@@ -50,7 +50,7 @@ const (
 	curlImage = "curlimages/curl:8.9.1"
 
 	// how long to wait in "Eventually" blocks
-	timeout = time.Second * 60
+	timeout = time.Second * 180
```
andyatmiami, line 53: nitpick:

I think your commit message is out of sync

> increase the timeout in controller e2e test from 1min to 2min.

as it looks like we are setting this to 3min

workspaces/controller/test/e2e/e2e_test.go
```diff
@@ -247,7 +247,9 @@ var _ = Describe("controller", Ordered, func() {
 				}
 				return nil
 			}
+			startTime := time.Now()
 			Eventually(verifyWorkspaceState, timeout, interval).Should(Succeed())
+			_, _ = fmt.Fprintf(GinkgoWriter, "It took %v to validate that the workspace has 'Running' state\n", time.Since(startTime))
```
andyatmiami: most other log messages are much more terse/action-oriented.. how about something like:

```suggestion
			_, _ = fmt.Fprintf(GinkgoWriter, "workspace reached 'Running' state in %v\n", time.Since(startTime))
```

andyatmiami (changes requested):
Thank you @siyuanfoundation for proposing this fix.

Have a couple small nits hopefully you don't find controversial - but totally agree this is the easiest path forward to avoid this flake 💯 

---

longer term - we probably (maybe?) wanna consider pre-pulling images **outside** the Ginkgo framework to avoid being subjected to timeouts for things like image pulls.. and/or when we aren't interested in the actual Workspace `image` itself (which would be almost always 😇 ) - we could potentially use some more smaller `minimal` style jupyter image..
- the current `scipy` image in use now seems to be ~1.2GB (compressed) which seems excessive for sake of this test 😎 

but either of those options require a little more _technical coordination_ and the small improvements you are making here I would still claim hold up even if we more comprehensively improve the framework.

workspaces/controller/test/e2e/e2e_test.go
siyuanfoundation: good catch! updated the description

andyatmiami (approved):
/lgtm

thesuperzapper:
Thanks, looks good to merge.

/approve

PR #1135, opened 2026-05-28
chore: do not rely on type/immutable fields from secret list

andyatmiami:
ℹ️ _NO GH ISSUE_ 

The upcoming Secrets API implementation (https://github.com/kubeflow/notebooks/pull/955) removes `type` and `immutable` from `SecretListItem` because the list endpoint will use a metadata-only cache (PartialObjectMetadata) that does not have access to these fields. This commit prepares the frontend so that PR is not a breaking change.

Key changes:

- SecretsAttachModal: remove Type label and Immutable badge from the secret picker dropdown, as these were sourced from the list response

- WorkspaceFormPropertiesSecrets: remove Immutable badge from the secrets table, simplify edit-action disable logic from `isImmutable || !canUpdate` to `!canUpdate` (immutability enforcement now happens inside the edit modal itself)

- Rename useSecretContents to useSecret: expand the hook to return `immutable` and `type` from the single-secret GET response (SecretUpdate), which still carries these fields. The hook was already making this API call but was discarding everything except key-value pairs.

- SecretsCreateModal: source immutable state from the useSecret hook (GET response) instead of from secretToEdit (list item). The immutable switch disable guard and submit button disable guard are preserved, just re-wired to the GET response data. This is important because canUpdate (a label) and immutable (a K8s Secret field) are independent — the backend does not enforce that immutable secrets have canUpdate=false.

### DIFF
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/properties/secrets/SecretsAttachModal.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/properties/secrets/SecretsAttachModal.tsx
@@ -181,12 +181,6 @@ export const SecretsAttachModal: React.FC<SecretsAttachModalProps> = ({
               <Stack>
                 <StackItem>
                   <LabelGroup>
-                    <Label isCompact>Type: {secret.type}</Label>
-                    {secret.immutable && (
-                      <Label color="orange" isCompact>
-                        Immutable
-                      </Label>
-                    )}
                     {!secret.canMount && <Label isCompact>Unmountable</Label>}
                   </LabelGroup>
                 </StackItem>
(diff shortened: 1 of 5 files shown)


andyatmiami:
Tested this out locally in Tilt - I think the following screenshots capture/demonstrate I didn't horribly break anything 😇 

image

image

image

workspaces/frontend/src/app/pages/Workspaces/Form/properties/secrets/SecretsAttachModal.tsx
```diff
@@ -181,12 +181,6 @@ export const SecretsAttachModal: React.FC<SecretsAttachModalProps> = ({
               <Stack>
                 <StackItem>
                   <LabelGroup>
```
thaorell: Consider removing the `<LabelGroup>` since there is now only one `<Label>`

thaorell (commented):
Just a minor comment from my intuition. If it produces any unpredictable behavior, please discard the comment

paulovmr (approved):
/lgtm
/approve

PR #1141, opened 2026-06-03
ci: only publish semver-tagged images from release branches

christian-heusel:
Guard `tag_with_semver` with `startsWith(github.ref, 'refs/heads/v')` to
further ensure that these image are only published from release branches.

See https://github.com/kubeflow/notebooks/pull/987 for reference.


andyatmiami (approved):
/lgtm
/approve

consistent with #987 which was already reviewed/merged...

PR #1151, opened 2026-06-04
feat: allow ingress from istio-system to workspace pods

richabanker:
This PR addresses a routing/access issue when deploying Kubeflow Workspaces/Notebooks v2 with the Istio component overlay.

The namespace base configuration defines a [default-deny-ingress](https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/controller/manifests/kustomize/base/namespace/network-policy-default-deny.yaml) NetworkPolicy, which blocks all incoming traffic to all pods. Currently, there is no companion `NetworkPolicy` configured under the `istio` component to allow ingress traffic from the Istio Ingress Gateway to the dynamic workspace notebook pods.

### DIFF
--- a/workspaces/controller/manifests/kustomize/components/istio/kustomization.yaml
+++ b/workspaces/controller/manifests/kustomize/components/istio/kustomization.yaml
@@ -3,6 +3,7 @@ kind: Component
 
 resources:
 - network-policy.yaml
+- network-policy-workspaces.yaml
 
 patches:
 - path: deployment_patch.yaml
--- a/workspaces/controller/manifests/kustomize/components/istio/network-policy-workspaces.yaml
+++ b/workspaces/controller/manifests/kustomize/components/istio/network-policy-workspaces.yaml
@@ -1,0 +1,19 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: workspaces-ingress-allow
+spec:
+  podSelector:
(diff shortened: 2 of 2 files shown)


christian-heusel (changes requested):
The instantiations of the Workspaces are hosted in the user-defined namespaces ("Kubeflow Profiles") so they are not in the same namespace as the controller (which lives in in the ns `kubeflow-workspaces`), so I don't think adding this NetworkPolicy is neccessary 🤔

Additionally: How did you test this change? I wondered quite a bit about how we can make sure that our network policies are correct when dealing with the review comment here: https://github.com/kubeflow/manifests/pull/3455#discussion_r3226676475, so if you have any good process around that it would be good to know!

richabanker:
> The instantiations of the Workspaces are hosted in the user-defined namespaces ("Kubeflow Profiles") so they are not in the same namespace as the controller (which lives in in the ns kubeflow-workspaces), so I don't think adding this NetworkPolicy is neccessary

Ah that makes sense, thanks for clarifying that the actual workspace pods are not created in kubeflow-workspaces which has the default-deny ingress policy. I tried to create my workspace in the "kubeflow" namespace `kubectl apply -f workspaces/controller/manifests/kustomize/samples/jupyterlab_v1beta1_workspace.yaml -n kubeflow` and that seems to not run into any issues when I hit connect. 

I'll close this PR.

/close

PR #1152, opened 2026-06-04
chore: update notebook server images

thesuperzapper:
This PR updates all the `example-notebook-servers` images.

I have tested all images locally on AMD64 and ARM (except TensorFlow CUDA image, as they dont publish wheels for ARM + CUDA upstream). 

Note, that we removed the Intel Gaudi images in https://github.com/kubeflow/notebooks/pull/1101 because we have no way to test the required updates. In future Intel is welcome to re-add these.

# Changes

## Python Version

The minor version of Python has been upgraded to `3.12` from the previous `3.11`.

## Kubeflow SDK

We now include the [Kubeflow SDK](https://github.com/kubeflow/sdk) python package at version [`0.4.0`](https://github.com/kubeflow/sdk/releases/tag/0.4.0) in all of our "full" images (and also the "scipy" Jupyter image).

## JupyterLab (TensorFlow - CUDA)

We have updated TensorFlow to `2.21.0`. 

Because support for NVIDIA TensorRT was removed in `2.18.0`, we no longer install it in the image, which means this version will have significant performance regressions. Also note, the latest version of TensorFlow does not support NVIDIA Blackwell Datacenter or Consumer GPUs.

## Code Server

We now set `XDG_DATA_HOME` to `/tmp/data-${NB_USER}` in the codeserver image, as this is where codeserver places a number of configs (e.g. extensions) and by default `XDG_DATA_HOME` is on the users home directory (which will have a PVC mounted and so users may not end up running the extensions we expect). Also state files like [heartbeat](https://github.com/coder/code-server/blob/main/docs/FAQ.md#what-is-the-heartbeat-file) would be on the PVC which is not desirable. 

After this change the heartbeat file is at: `/tmp/data-jovyan/code-server/heartbeat`

## RStudio

Similar to the change in Code Server, we now set `RSTUDIO_DATA_HOME` which is where user-specific runtime state is stored, which defaults to `${XDG_DATA_HOME}/rstudio` should not live on the PVC.

Additionally, we set the CRAN repo for ARM64 images to one which has pre-built binaries for ARM64 so installs of common packages should no longer require builds, similar to how the AMD64 images worked before.

# Version Changes

## Binaries

- kubectl: `1.31.6` -> `1.35.5`

## IDEs

- CodeServer: `4.96.4` -> `4.121.0`
- JupyterLab: `4.3.5` -> `4.5.7`
- RStudio: `2024.12.1-563` -> `2026.05.0-218`

## Python

- Python: `3.11.11` -> `3.12.13`
- Pip: `24.3.1` -> `26.1.1`
- Miniforge: `24.11.3-0` -> `26.3.2-2`

### Kubeflow SDK

- kubeflow: `0.4.0` (NEW)

### Kubeflow Pipelines

- kfp: `2.11.0` -> `2.16.1`

### PyTorch

- PyTorch: `2.5.1` -> `2.12.0`

### TensorFlow

- TensorFlow: `2.17.1` -> `2.21.0`

## R

- r-base: `4.4.2` -> `4.5.3`
- r-reticulate: `1.40.0` -> `1.46.0`
- r-shiny: `1.10.0` -> `1.13.0`


components/example-notebook-servers/codeserver/Dockerfile
```diff
@@ -9,7 +9,7 @@ ARG TARGETARCH
 
 # args - software versions
 # https://github.com/coder/code-server/releases
-ARG CODESERVER_VERSION=4.96.4
+ARG CODESERVER_VERSION=4.121.0
```
christian-heusel, line 12: Any reason we're not using the latest version here? https://github.com/coder/code-server/releases/tag/v4.123.0

components/example-notebook-servers/codeserver-python/Dockerfile
```diff
@@ -65,10 +65,13 @@ RUN echo "ipykernel ==${IPYKERNEL_VERSION}" >> ${CONDA_DIR}/conda-meta/pinned \
 
 # install - requirements.txt
 COPY --chown=${NB_USER}:${NB_GID} requirements.txt /tmp
-RUN python3 -m pip install -r /tmp/requirements.txt --quiet --no-cache-dir \
+RUN PIP_NO_CACHE_DIR=off python3 -m pip install -r /tmp/requirements.txt --quiet --no-cache-dir \
```
christian-heusel, line 68: Why has the `PIP_NO_CACHE_DIR=off` variable been added here? It contradicts the `--no-cache-dir` option later in the line 🤔 
https://pip.pypa.io/en/latest/cli/pip/#cmdoption-no-cache-dir

christian-heusel (approved):
I test-built all images and gave all the (non-cuda) ones a smoke-test on my local installation and all worked fine 🤗 

Image

I have posted two small remarks below, but given the urgency of landing these updates and that they don't have any functional impact I don't view them as blockers.

components/example-notebook-servers/codeserver-python/Dockerfile
thesuperzapper: Strangly `off` actually means on in this case, its possible that `on` also means on now in newer versions of pip (based on [this docstring](https://github.com/pypa/pip/blob/31d7d168953668aad85154d6121879d07fbeac27/src/pip/_internal/cli/cmdoptions.py#L931-L937), but its pretty unclear and the `PIP_NO_CACHE_DIR=off` value deffinitly works in all versions of pip.

components/example-notebook-servers/codeserver/Dockerfile
thesuperzapper: Its just the lastest version when I was testing these images. We probably can updated in a future PR as these versions were only releawsed in the last few days.

PR #1158, opened 2026-06-04
fix: show friendly workspace kinds access message

### LINKED ISSUE #1114: [TASK] Show a friendly message for non-admin users on the WorkspaceKind page
### Certification

### Description

Currently, when a non-admin user navigates to the WorkspaceKind page, they see an error (likely a 403 or failed API response rendered as an error state). Until the dashboard supports conditionally hiding the WorkspaceKind navigation item for non-admins, we should handle this gracefully.

Display a friendly informational message (not an error) such as: "WorkspaceKind management is restricted to administrators. Contact your admin if you need changes to workspace configurations."

 This should use an appropriate PatternFly empty state pattern (e.g., EmptyState with a lock or info icon) rather than an error alert.

### Acceptance Criteria

- [ ] Non-admin users see a friendly, non-error message on the WorkspaceKind page
- [ ] No error alerts or broken UI states are shown
- [ ] The page is still navigable (back button, sidebar links work)

yanmo42:
Fixes #1114.

## Summary
- replace the raw forbidden alert with a calmer info state for non-admin users
- explain that WorkspaceKind visibility depends on permissions
- add WorkspaceKinds page coverage for the access-message state

## Validation
- git diff --check
- npm --prefix workspaces/frontend run test:jest -- WorkspaceKinds.spec.tsx --runInBand
- npm --prefix workspaces/frontend run test:type-check

## AI Assistance
This contribution was prepared with AI assistance from OpenAI Codex. The commit includes an Assisted-by trailer.


workspaces/frontend/src/app/pages/WorkspaceKinds/__tests__/WorkspaceKinds.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+  typeof useWorkspaceCountPerKind
+>;
+
+describe('WorkspaceKinds', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockUseWorkspaceCountPerKind.mockReturnValue({ workspaceCountPerKind: {}, error: null });
+  });
+
+  it('shows a friendly message for users who cannot manage workspace kinds', () => {
```
Snehadas2005: Hi @yanmo42!

I have a quick suggestion! If we’re changing the `data-testid` to "`workspace-kinds-access-empty-state`" and updating the text, we’ll also need to update the Cypress integration tests. 

The `workspaceKinds.cy.ts` file is still looking for the old `data-testid="access-restricted-empty-state"`, so it would be great to update that to prevent any CI issues.

workspaces/frontend/src/app/pages/WorkspaceKinds/WorkspaceKinds.tsx
```diff
@@ -92,6 +95,8 @@ export const WorkspaceKinds: React.FunctionComponent = () => {
   }, [navigate]);
   const [workspaceKinds, workspaceKindsLoaded, workspaceKindsError] = useWorkspaceKinds();
   const workspaceCountResult = useWorkspaceCountPerKind();
+  const { user } = useAppContext();
+  const canManageWorkspaceKinds = Boolean(user?.clusterAdmin);
```
Snehadas2005: I noticed that `useWorkspaceKinds()` and `useWorkspaceCountPerKind()` are firing before we verify `canManageWorkspaceKinds`. This can lead to unauthorized network requests for non-admin users, resulting in a lot of unhandled `403 Forbidden` errors in the console.

Could we check if these hooks can accept an options/config object to conditionally disable them when `canManageWorkspaceKinds` is false? Something like `{ enabled: canManageWorkspaceKinds }` could help if the hook supports it.

workspaces/frontend/src/app/pages/WorkspaceKinds/WorkspaceKinds.tsx
```diff
@@ -264,6 +269,27 @@ export const WorkspaceKinds: React.FunctionComponent = () => {
...
+          <EmptyState
+            headingLevel="h4"
+            titleText="Workspace kinds are managed by administrators"
+            icon={LockIcon}
+            status="info"
+            data-testid="workspace-kinds-access-empty-state"
+          >
+            <EmptyStateBody>
+              You do not have permission to manage workspace kinds. You can still create workspaces
+              from the workspace kinds available to you.
```
Snehadas2005: The text you added is very clear and friendly! However, I wanted to point out that the original issue description (#1114) requested a specific phrasing:

_"WorkspaceKind management is restricted to administrators. Please contact your admin if you need changes to workspace configurations."_

In my opinion, adhering closely to the original wording will make this much more appropriate and effective.

Snehadas2005:
I really appreciate this refactor! From my perspective, dividing the component into `WorkspaceKindsAdminView` is an excellent approach; it effectively manages the permissions gate while avoiding unnecessary network requests.

However, I noticed that the adjustments made for the non-admin layout, specifically changing the target data-testid to `"workspace-kinds-access-empty-state"` and removing the baseline table view text,  will cause the E2E integration tests to break in CI.

In my opinion, we just need to update the describe block labeled `'Non-admin access'` in `workspaces/frontend/src/__tests__/cypress/cypress/tests/mocked/workspaceKinds/workspaceKinds.cy.ts`. Aligning that section to look for the new test ID and the revised string content should be the last step here to keep the pipeline green.

yanmo42:
Thanks for the read — appreciated. Before changing the cypress file I wanted to verify the premise, and a couple of things don't line up. Sharing what I found so we're working from the same picture:

### 1. There is no `'Non-admin access'` describe block in `workspaceKinds.cy.ts`

```
$ grep -rnE "Non-admin|non-admin|workspace-kinds-access" workspaces/frontend/
(no matches)
```

The current top-level `describe` blocks in that file are: `Basic`, `Empty state`, `Filter`, `Sorting`, `Pagination`, `Actions` (with nested `View Details`, `Edit`). No `'Non-admin access'` block exists on this PR branch or on `notebooks-v2`.

### 2. Cypress can't reach the non-admin branch yet

`workspaces/frontend/src/app/context/AppContext.tsx` hardcodes:

```ts
const userSettings: UserSettings = useMemo(
  () => ({
    userId: 'kubeflow-user',
    clusterAdmin: true,
  }),
  [],
);
```

…with the existing `TODO: replace userSettings with const { configSettings, userSettings } = useSettings() once integrated with users`. Cypress runs the real `AppContextProvider`, so `user?.clusterAdmin` is always truthy and the new `EmptyState` path is unreachable from cypress until `clusterAdmin` becomes overrideable. All existing cypress cases continue to hit `WorkspaceKindsAdminView` and pass.

### 3. CI is green; `tide` pending is a label gate, not a test failure

```
DCO              pass
Validate Title   pass
ok-to-test       pass
tide             pending — Not mergeable. Needs approved, lgtm labels.
```

The `tide` state is waiting on `approved` + `lgtm`, not on a broken pipeline.

### 4. Non-admin behavior is already covered

The new Jest spec added in this PR — `workspaces/frontend/src/app/pages/WorkspaceKinds/__tests__/WorkspaceKinds.spec.tsx` — asserts:

- `data-testid="workspace-kinds-access-empty-state"` is rendered,
- the title `"WorkspaceKind management is restricted to administrators"`,
- the body `"Please contact your admin if you need changes to workspace configurations."`,
- and that `useWorkspaceKinds` / `useWorkspaceCountPerKind` are **not** called for non-admins.

### Proposed path forward

Given the above, I'd rather not edit `workspaceKinds.cy.ts` in this PR — there's nothing currently failing there to align, and any new `describe('Non-admin access', ...)` block would be a no-op until `clusterAdmin` is mockable from cypress.

Happy to do cypress-level coverage as a focused follow-up once `AppContextProvider` supports a non-admin override (either via the upcoming `useSettings()` integration or a small test-only hook). If you'd prefer I bundle that work into this PR instead, let me know and I'll scope it.

Snehadas2005:
Thank you so much for the deep dive and the detailed breakdown, @yanmo42!

You are completely right, from my perspective, I must have been referencing an unmerged local draft/workspace setup when I looked for that describe(`'Non-admin access'`) block! Your point about `AppContext.tsx` hardcoding `clusterAdmin: true` is a fantastic catch too; it makes total sense why a Cypress test would just be a no-op right now.

In my opinion, your new Jest unit tests cover the logic perfectly for this stage of the feature. Let's definitely leave the Cypress coverage for a follow-up issue once` clusterAdmin` becomes mockable or overrideable.

This looks amazing and completely addresses the core requirements. I'm happy to sign off on this. Everything looks good to go from my side! Code is clean. 

/lgtm

cc: @thaorell could you please take a look, it's great to have your feedback.

workspaces/frontend/src/app/pages/WorkspaceKinds/__tests__/WorkspaceKinds.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+    jest.clearAllMocks();
+    mockUseWorkspaceCountPerKind.mockReturnValue({ workspaceCountPerKind: {}, error: null });
+  });
+
+  it('shows the restricted-access message for non-admin users', () => {
+    mockUseAppContext.mockReturnValue({
+      config: null,
+      user: { userId: 'kubeflow-user', clusterAdmin: false },
+    });
+    mockUseWorkspaceKinds.mockReturnValue([[], true, new Error('Forbidden'), jest.fn()]);
```
thaorell: The first test sets up `mockUseWorkspaceKinds.mockReturnValue([[], true, new Error('Forbidden'), jest.fn()])`, but this hook is never called for non-admin users since  `WorkspaceKindsAdminView` doesn't render. This mock setup is misleading — it suggests the test exercises an error path when it actually tests the admin gate. Remove this line to make the test intent clearer (the second test already explicitly asserts the hooks aren't called).

workspaces/frontend/src/app/pages/WorkspaceKinds/__tests__/WorkspaceKinds.spec.tsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
```
thaorell, line 1: The tests cover `clusterAdmin: false` but not the edge case where `user` is `null` (e.g., before user settings are loaded). Since `Boolean(null?.clusterAdmin)` evaluates to `false`, a  null user will see the restricted-access view — consider adding a test to document this behavior explicitly so it doesn't regress.

thaorell (changes requested):

thaorell:
Hi @yanmo42, can you provide some screenshots of the before and after?

thaorell (commented):
@yanmo42 Any update on this?

yanmo42:
Hi @thaorell — sorry for the delay! Here are the before/after screenshots, and an update on the review feedback.

## Screenshots

**Before** (non-admin users saw the full WorkspaceKinds management table — including Create/Edit actions they can't actually use):

Before

**After** (non-admin users see a clear, friendly access message):

After

## Review feedback addressed

Both of your review comments from June 8 are addressed in commit `37686a7a`:

1. **Misleading mock removed** — the first test no longer sets up `mockUseWorkspaceKinds.mockReturnValue([[], true, new Error('Forbidden'), jest.fn()])`, since `WorkspaceKindsAdminView` never renders for non-admin users and the hook is never called.

2. **Null-user edge case added** — a third test covers `user: null` (user context not yet loaded), documenting that `Boolean(null?.clusterAdmin)` → `false` and the restricted-access view is shown in that case too (hooks not called).

Happy to make any further adjustments!

andyatmiami:
/close

closing due to inactivity.. please re-open if you feel this was done in err and we plan to progress this PR

christian-heusel:
@Snehadas2005 @thisis-shitanshu would one of you be interested in revisiting this PR? Not neccessarily from the code-perspective, but I believe this would potentially be a good UX improvement 😊

Snehadas2005:
Hi @christian-heusel! I will take this forward. Will rebase these changes, open a new PR referencing #1114, and tag you for review soon! 🙌

PR #1166, opened 2026-06-07
refactor: Add omitempty to slice fields in backend types

christian-heusel:
Split out from the original PR (which also introduces the custom linter) so it is easier to review / merge separately.

Related-to: https://github.com/kubeflow/notebooks/pull/976

Related-to: https://github.com/kubeflow/notebooks/pull/980

### DIFF
--- a/workspaces/backend/internal/models/pvcs/funcs.go
+++ b/workspaces/backend/internal/models/pvcs/funcs.go
@@ -59,17 +59,8 @@ func NewPVCListItemFromPVC(pvc *corev1.PersistentVolumeClaim, pv *corev1.Persist
 		}
 	}
 
-	// build the list of PodInfo for pods that mount this PVC, if any.
-	podList := make([]PodInfo, 0)
-	if pvcToPodInfoList[pvc.Name] != nil {
-		podList = pvcToPodInfoList[pvc.Name]
-	}
-
-	// build the list of WorkspaceInfo for workspaces that reference this PVC, if any.
-	workspaceList := make([]WorkspaceInfo, 0)
-	if pvcToWorkspaceInfoList[pvc.Name] != nil {
-		workspaceList = pvcToWorkspaceInfoList[pvc.Name]
-	}
+	podList := pvcToPodInfoList[pvc.Name]
+	workspaceList := pvcToWorkspaceInfoList[pvc.Name]
 
 	return PVCListItem{
 		Name:       pvc.Name,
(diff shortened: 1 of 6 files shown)


christian-heusel:
/cc @thaorell @paulovmr 
Regarding the frontend implications

workspaces/backend/internal/models/pvcs/types_write.go
```diff
@@ -27,7 +27,7 @@ import (
 // PVCCreate represents the request and response body for creating a PVC.
 type PVCCreate struct {
 	Name             string                              `json:"name"`
-	AccessModes      []corev1.PersistentVolumeAccessMode `json:"accessModes"`
+	AccessModes      []corev1.PersistentVolumeAccessMode `json:"accessModes,omitempty"`
```
andyatmiami: **NOTHING** is ever as easy as it seems 🤣 

Our desire to have a blanket "all slices end up treated as `omitempty` in JSON" isn't feasible - with this `AccessModes` case highlighting why...

Adding `omitempty` causes `swag` to remove `accessModes` from the OpenAPI `required` array. However, `Validate()` at line 46 still enforces `len(p.AccessModes)   == 0 → error.`  This means the Swagger spec now says the field is optional, but the server rejects requests without it.

`PVCCreate.AccessModes` should not have `omitempty`. It's a semantically required field backed by both Kubernetes admission and explicit server-side  validation.  We probably want an inline comment on this field explaining why `omitempty` is... for lack of a better word.. omitted. 😎 

This would really **ONLY** ever be a _possibility_ on `x_write.go` models... such that in cases where the array is in fact **required** to be populated (like in this `AccessModes` case) - then the field **MUST** be marked `required` in our OpenAPI

andyatmiami (changes requested):
Thanks for this PR @christian-heusel ... in general - well structured and consistent...

Unfortunately it highlighted an edge case I don't think any of us fully considered as we "dreamed" of a nice, simple "use `omitempty` in ALL THE PLACES" outcome 😎

workspaces/backend/internal/models/pvcs/types_write.go
thesuperzapper: Unless I'm forgetting something, I think we only were saying it was a requirement that the RETURN models used ommitempty on slices?

The WRITE models obviously are populated by callers, so having required (no ommitempty) is actually desirable.

thesuperzapper: It's probably fine to just lint all of them, and then ignore the lint for any specific case like this. 

Because it's kind of problematic to only apply the lint to files that don't end in `_write.go`

christian-heusel: Reverted that diff, thanks for the feedback! 🤗

PR #1157, opened 2026-06-04
fix: show unknown label for missing workspace state

yanmo42:
Fixes #1104.

## Summary
- render a neutral Unknown state when a workspace has no current status
- avoid leaving the state cell empty for missing/undefined status values
- add WorkspaceTable coverage for the missing-state fallback

## Validation
- git diff --check
- npm --prefix workspaces/frontend run test:jest -- WorkspaceTable.spec.tsx --runInBand
- npm --prefix workspaces/frontend run test:type-check

## AI Assistance
This contribution was prepared with AI assistance from OpenAI Codex. The commit includes an Assisted-by trailer.

Issue #1104, opened 2026-05-19
[TASK] Default workspace state should display as "unknown" for newly created workspaces


paulovmr:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

When a workspace is first created and has no status yet, the workspace table shows a circle icon with no text, which looks like a rendering bug. The default state should display as "unknown" with appropriate styling (e.g., a grey status indicator) until the controller reconciles and sets an actual state.

### Acceptance Criteria

- [ ] Newly created workspaces without a status show "Unknown" state text with a neutral/grey indicator
- [ ] The status icon is never rendered without accompanying text

RajGautam2004:
Hi! I noticed this issue is assigned. If it's available or the assignee is no longer working on it, I'd be happy to help. Please let me know. Thanks!

thaorell:
Thank you @RajGautam2004, but I have opened a PR for this issue #1260

PR #1167, opened 2026-06-07
fix: default empty workspace state to Unknown in API responses

yanmo42:
Refs #1104. Replaces #1157.

## Summary
- Default `ws.Status.State` to `WorkspaceStateUnknown` in `NewWorkspaceListItemFromWorkspace` so the list-workspaces API never returns `"state": ""`.
- Add a `defaultWorkspaceState` helper plus a unit test.

## Root cause
`workspaces/backend/internal/models/workspaces/funcs.go` was copying `ws.Status.State` verbatim. `WorkspaceState` is a string-aliased Go type (zero value `""`), and the JSON tag is `json:"state"` (no `omitempty`), so a workspace whose controller had not yet reconciled a status was serialised with an empty `state` field. This contradicts the generated OpenAPI/TS contract, which declares the field as the non-empty `V1Beta1WorkspaceState` enum, and is the source of the empty-label "circle icon" symptom in #1104.

Fixing it server-side keeps the contract honest and removes the need for defensive fallbacks in every client surface (table cell, PVC tooltip, etc.).

## Context
Originally raised as a frontend fallback in #1157; @thaorell correctly flagged the fallback as redundant against the type contract. Tracing the symptom in the backend surfaced the real source.

## Validation
- `go test ./workspaces/backend/internal/models/workspaces/...`
- `go vet ./workspaces/backend/internal/models/workspaces/...`
- `gofmt -d` clean

## AI Assistance
Prepared with AI assistance from Claude. The commit includes an `Assisted-by` trailer.

### DIFF
--- a/workspaces/backend/internal/models/workspaces/funcs.go
+++ b/workspaces/backend/internal/models/workspaces/funcs.go
@@ -86,7 +86,7 @@ func NewWorkspaceListItemFromWorkspace(cfg *config.EnvConfig, ws *kubefloworgv1b
 		Paused:         ptr.Deref(ws.Spec.Paused, false),
 		PausedTime:     ws.Status.PauseTime,
 		PendingRestart: ws.Status.PendingRestart,
-		State:          ws.Status.State,
+		State:          defaultWorkspaceState(ws.Status.State),
 		StateMessage:   ws.Status.StateMessage,
 		PodTemplate: PodTemplate{
 			PodMetadata: PodMetadata{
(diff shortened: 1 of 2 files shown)


Snehadas2005 (commented):
I have a few suggestions regarding the core fix. I think it effectively addresses the root cause in `NewWorkspaceListItemFromWorkspace`, which helps prevent empty-state values in the primary workspace list views. 

To ensure complete coverage and avoid identical symptoms across other parts of the UI, it would be highly beneficial to address one more internal mapper that currently copies `ws.Status.State` directly. Including it here will maintain strict structural consistency across the repository.

1. **Synchronise Workspace Status in PVC References:** When querying PVC details, the `buildWorkspaceInfoMap` helper directly copies the workspace state. Setting it to `WorkspaceStateUnknown` helps avoid duplicate missing tooltips or empty-state UI artefacts in the PVC overlay views.  Could you please add some similar adjustments in the file `workspaces/backend/internal/repositories/pvcs/repo.go`?
```
state := ws.Status.State
if state == "" {
    state = kubefloworgv1beta1.WorkspaceStateUnknown
}

wsInfo := models.WorkspaceInfo{
    Name:         ws.Name,
    State:        state,
    StateMessage: ws.Status.StateMessage,
}
```

2. **Accompanying PVC Handler Test Update:** Updating the mapper above means the mock workspace setup in the PVC handler test suite will expect the explicit "Unknown" enum alignment instead of a raw empty string. Can you make changes like the ones below in the file `workspaces/backend/api/pvcs_handler_test.go`?
```
Workspaces: []models.WorkspaceInfo{
    {
        Name:  workspaceName1,
        State: kubefloworgv1beta1.WorkspaceStateUnknown,
    },
},
```
Once these data boundaries are aligned, this task is fully future-proofed.

thaorell (commented):
This will be replaced by #1260 due to staleness

PR #1140, opened 2026-06-02
chore: integrate prometheus in tilt

andyatmiami:
ℹ️ _NO GH ISSUE_

Add opt-in Prometheus support to the KinD/Tilt development environment so developers can verify that metrics are properly exposed, ServiceMonitor manifests are valid, and the full scrape pipeline works end-to-end.

## Development environment (developing/)

- Add `setup-prometheus.sh` script that installs the Prometheus Operator via upstream `bundle.yaml` and deploys a minimal Prometheus instance in `prometheus-system` namespace. Follows the same idempotent pattern as `setup-cert-manager.sh` and `setup-istio.sh`.

- Add `manifests/prometheus/prometheus.yaml` with the Prometheus CR, ServiceAccount, ClusterRole, and ClusterRoleBinding. Configured with minimal resources (128Mi/256Mi), 2h retention, and wildcard ServiceMonitor discovery across all namespaces.

- Add `manifests/overlays/controller-prometheus/kustomization.yaml`, a dev-only kustomize overlay that layers the existing prometheus component on top of the istio overlay. This avoids creating a new overlay in the controller's published manifests — enabling prometheus is a dev-environment concern, not a vendoring decision.

- Update `Makefile` with `setup-prometheus` target, conditionally invoked by `tilt-up` when `ENABLE_PROMETHEUS=true`.

- Update `Tiltfile` with `ENABLE_PROMETHEUS` env var, conditional overlay selection, and a `prometheus-ui` local_resource that port-forwards to localhost:9090.

### NetworkPolicy for metrics scraping

Add `network-policy.yaml` to the controller's prometheus kustomize component. Without this, the existing `default-deny-ingress` policy blocks Prometheus from reaching the controller's metrics port. Uses `from: []` (allow from any source) scoped to the metrics port, matching the same pattern used by the istio component's webhook NetworkPolicy.

### Remove `managed-by` from CRD selectors

The `common` kustomize component previously used a single `labels` block with `includeSelectors: true` for all three standard labels, including `app.kubernetes.io/managed-by: kustomize`. This injected `managed-by` into every selector field — including CRD-specific selectors like `ServiceMonitor.spec.selector` and `AuthorizationPolicy.spec.selector`.

[...]

PR #1168, opened 2026-06-07
feat: add custom Prometheus metrics to the controller

### LINKED ISSUE #1163: [TASK] Investigate and extend Prometheus metrics for the Notebooks controller
### Certification

### Description

PR #1140 introduces opt-in Prometheus support to the Tilt development environment, wiring up a full scrape pipeline so developers can verify metrics end-to-end.

Now that the infrastructure is (about to be) in place, we should:

1. **Audit the default `controller-runtime` metrics** and understand which metrics are exposed out of the box (reconcile duration, workqueue depth, etc.) and whether the set is configurable.
2. **Add at least one custom domain-specific metric** to the Notebooks controller, using the kubebuilder custom metrics functionality (https://book.kubebuilder.io/reference/metrics.html#publishing-additional-metrics).

Example candidate for custom metrics:
- Number of `Workspace` resources grouped by `WorkspaceKind`
- `WorkspaceKind` option usage (how many Workspaces use each pod/image config option)

### Acceptance Criteria

- [x] At least one custom metric is implemented in the controller and visible in the Prometheus UI.
- [x] Metric names and label conventions follow the kubebuilder guidelines.
- [x] The custom metric(s) are covered by at least basic tests.

christian-heusel:
Implements three GaugeVec metrics (Workspace count per kind, imageConfig usage and podConfig usage).

Depends on #1140 for testing
Fixes https://github.com/kubeflow/notebooks/issues/1163

### DIFF
--- a/workspaces/controller/internal/controller/metrics.go
+++ b/workspaces/controller/internal/controller/metrics.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	workspaceKindWorkspaceCount = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "notebooks",
+			Name:      "workspacekind_workspace_count",
+			Help:      "Number of Workspace resources managed by each WorkspaceKind.",
+		},
+		[]string{"workspace_kind"},
+	)
+
+	workspaceKindImageConfigWorkspaceCount = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "notebooks",
+			Name:      "workspacekind_image_config_workspace_count",
+			Help:      "Number of Workspace resources using each imageConfig option, grouped by WorkspaceKind.",
+		},
+		[]string{"workspace_kind", "image_config"},
+	)
+
+	workspaceKindPodConfigWorkspaceCount = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "notebooks",
+			Name:      "workspacekind_pod_config_workspace_count",
+			Help:      "Number of Workspace resources using each podConfig option, grouped by WorkspaceKind.",
+		},
+		[]string{"workspace_kind", "pod_config"},
+	)
+)
+
+func init() {
+	ctrlmetrics.Registry.MustRegister(
+		workspaceKindWorkspaceCount,
+		workspaceKindImageConfigWorkspaceCount,
+		workspaceKindPodConfigWorkspaceCount,
+	)
+}
+
+// clearWorkspaceKindMetrics removes all metric time series for a deleted WorkspaceKind.
+func clearWorkspaceKindMetrics(kindName string) {
+	workspaceKindWorkspaceCount.DeleteLabelValues(kindName)
+	workspaceKindImageConfigWorkspaceCount.DeletePartialMatch(prometheus.Labels{"workspace_kind": kindName})
+	workspaceKindPodConfigWorkspaceCount.DeletePartialMatch(prometheus.Labels{"workspace_kind": kindName})
+}
--- a/workspaces/controller/internal/controller/workspacekind_controller.go
+++ b/workspaces/controller/internal/controller/workspacekind_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -75,6 +76,7 @@ func (r *WorkspaceKindReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 			// Owned objects are automatically garbage collected.
 			// For additional cleanup logic use finalizers.
 			// Return and don't requeue.
+			clearWorkspaceKindMetrics(req.Name)
 			return ctrl.Result{}, nil
 		}
 		log.Error(err, "unable to fetch WorkspaceKind")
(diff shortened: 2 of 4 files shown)


workspaces/controller/internal/controller/workspacekind_controller.go
```diff
@@ -143,6 +145,18 @@ func (r *WorkspaceKindReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// update Prometheus metrics
+	workspaceKindWorkspaceCount.WithLabelValues(workspaceKind.Name).Set(float64(numWorkspace))
+	// we call metric.DeletePartialMatch() to ensure that there are no stale stale time series
+	workspaceKindImageConfigWorkspaceCount.DeletePartialMatch(prometheus.Labels{"workspace_kind": workspaceKind.Name})
```
christian-heusel, line 151: Just as a reminder to myself, Andy and I were talking about this code section in an in-call review and we're both wanted to think about ways of avoiding the `DeletePartialMatch()` call or make it more specific to avoid the race condition where the Metrics are queried after the Delete but before being set again.

christian-heusel:
cc @richabanker (since it's something related to metrics) 🤗

workspaces/controller/internal/controller/workspacekind_controller.go
richabanker: Have we considered using a custom `prometheus.Collector` for this instead of trying to manage the metric state ourselves which is prone to races? (For ref, [this](https://github.com/kubernetes/kubernetes/blob/cf7bfb2dca32e5cb9e5379c004a3ff7cd62603ae/pkg/controller/volume/persistentvolume/metrics/metrics.go#L82 ) is a pattern used in upstream K8s which reads from the informer cache on scrape rather than mutating gauges inside the reconcile loop) 

Since `WorkspaceKindReconciler` already computes and persists these counts in `WorkspaceKind.Status` (`status.workspaces`, `status.podTemplateOptions.imageConfig`, etc.), how about implementing a standard `prometheus.Collector` (`Describe` and `Collect`):

- When Prometheus scrapes, `Collect()` lists `WorkspaceKind` from the in-memory cache and emits `prometheus.MustNewConstMetric()`
- Register it once with controller-runtime's registry via `ctrlmetrics.Registry.MustRegister(NewWorkspaceKindCollector(mgr.GetClient()))` (per the [Kubebuilder Metrics Guide](https://book.kubebuilder.io/reference/metrics.html#publishing-additional-metrics))

That way the stateless collector automatically solves the race condition by generating metrics atomically on every scrape. If a WorkspaceKind is deleted or an option is removed from the spec, it automatically drops off the next scrape, so no need for `DeletePartialMatch()` or explicit deletion cleanup. And added benefit of the reconciler logic staying focused without needing metric state management.

Gemini generated snippet of changes:
```
type WorkspaceKindCollector struct {
	client           client.Reader
	workspacesDesc   *prometheus.Desc
	imageConfigsDesc *prometheus.Desc
	podConfigsDesc   *prometheus.Desc
}

func (c *WorkspaceKindCollector) Describe(ch chan<- *prometheus.Desc) {
	ch <- c.workspacesDesc
	ch <- c.imageConfigsDesc
	ch <- c.podConfigsDesc
}

func (c *WorkspaceKindCollector) Collect(ch chan<- prometheus.Metric) {
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()
	wskList := &kubefloworgv1beta1.WorkspaceKindList{}
	if err := c.client.List(ctx, wskList); err != nil {
		return
	}

	for _, wsk := range wskList.Items {
		ch <- prometheus.MustNewConstMetric(c.workspacesDesc, prometheus.GaugeValue, float64(wsk.Status.Workspaces), wsk.Name)
		for _, img := range wsk.Status.PodTemplateOptions.ImageConfig {
			ch <- prometheus.MustNewConstMetric(c.imageConfigsDesc, prometheus.GaugeValue, float64(img.Workspaces), wsk.Name, img.Id)
		}
		for _, pod := range wsk.Status.PodTemplateOptions.PodConfig {
			ch <- prometheus.MustNewConstMetric(c.podConfigsDesc, prometheus.GaugeValue, float64(pod.Workspaces), wsk.Name, pod.Id)
		}
	}
}
```

workspaces/controller/internal/controller/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+var (
+	workspaceKindWorkspaceCount = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "notebooks",
+			Name:      "workspacekind_workspace_count",
+			Help:      "Number of Workspace resources managed by each WorkspaceKind.",
+		},
+		[]string{"workspace_kind"},
```
richabanker, line 31: CI failures seem to not like the repetitive use of the label (and Namespace: "notebooks"), can make them consts.

workspaces/controller/internal/controller/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	workspaceKindWorkspaceCount = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "notebooks",
+			Name:      "workspacekind_workspace_count",
```
richabanker, line 28: Acc to [Prometheus naming best practices](https://prometheus.io/docs/practices/naming/#metric-names), gauges should use the plural noun. The _count suffix is reserved for Summaries and Histograms. (Like `notebooks_workspacekind_workspaces` instead of `workspacekind_workspace_count`. Same for the other 2 metrics)

workspaces/controller/internal/controller/workspacekind_controller_test.go
```diff
@@ -400,4 +401,75 @@ var _ = Describe("WorkspaceKind Controller", func() {
...
+
+			By("asserting imageConfig metric drops to zero")
+			Expect(testutil.ToFloat64(workspaceKindImageConfigWorkspaceCount.WithLabelValues(workspaceKindName, "jupyterlab_scipy_180"))).To(Equal(float64(0)))
+
+			By("deleting the WorkspaceKind")
+			Expect(k8sClient.Delete(ctx, workspaceKind)).To(Succeed())
+			Expect(k8sClient.Get(ctx, workspaceKindKey, workspaceKind)).NotTo(Succeed())
+
+			By("asserting metrics are cleared after WorkspaceKind deletion")
+			Expect(testutil.ToFloat64(workspaceKindWorkspaceCount.WithLabelValues(workspaceKindName))).To(Equal(float64(0)))
```
richabanker, line 472: Calling `.WithLabelValues(...)` on a GaugeVec and asserting the value to be 0 to catch the case that the metric is deleted is a false-positive.

`.WithLabelValues(...)` will return a new (initialized to 0) gauge if the provided label combination does not exist in the vector (See [WithLabelValues() calls GetMetricWithLabelValues()](https://github.com/prometheus/client_golang/blob/57bb367da472cadc62ac8210c55ea2cd83143e5c/prometheus/gauge.go#L238) which [creates and initializes a new gauge with value 0](https://github.com/prometheus/client_golang/blob/57bb367da472cadc62ac8210c55ea2cd83143e5c/prometheus/gauge.go#L182-L183) if the label-value combination is not found)

Asserting 0 works for checking that the actual value (workspace count) dropped to zero, but cannot test that the metric was cleared upon resource deletion. To truly assert the metric being gone, we should do `testutil.CollectAndCount() == 0` ([ref](https://github.com/prometheus/client_golang/blob/57bb367da472cadc62ac8210c55ea2cd83143e5c/prometheus/testutil/testutil.go#L130), which works in this test since it creates exactly one WorkspaceKind)

workspaces/controller/internal/controller/workspacekind_controller.go
```diff
@@ -143,6 +145,18 @@ func (r *WorkspaceKindReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// update Prometheus metrics
+	workspaceKindWorkspaceCount.WithLabelValues(workspaceKind.Name).Set(float64(numWorkspace))
+	// we call metric.DeletePartialMatch() to ensure that there are no stale stale time series
```
richabanker, line 150: Typo - "stale stale"

workspaces/controller/internal/controller/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+			Help:      "Number of Workspace resources managed by each WorkspaceKind.",
+		},
+		[]string{"workspace_kind"},
+	)
+
+	workspaceKindImageConfigWorkspaceCount = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "notebooks",
+			Name:      "workspacekind_image_config_workspace_count",
```
richabanker, line 37: Can use `Subsystem: "workspacekind"` in prometheus.GaugeOpts (and make "workspacekind" a const) instead of prefixing it to Name for all the metrics.

Issue #1105, opened 2026-05-19
[TASK] Allow Admin to Edit WorkspaceKind YAML in the UI


thaorell:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

Allow admins to edit WorkspaceKind configuration directly in the UI with multiple view modes. Currently, WorkspaceKind editing is limited to a structured form. We want to add the ability to view and edit the underlying YAML, with the long-term goal of a split view that shows form and YAML side-by-side.
The preferred approach is a toggle that lets the user switch between **Form**, **Split**, and **YAML** views. The implementation should use the existing expandable form layout (not a drawer) since a split view is far easier to achieve with that pattern.

### Acceptance Criteria

- [x] Add `@patternfly/react-code-editor` as a dependency
- [x] Optional: Configure the monaco editor to accept `WorkspacekindsWorkspaceKindUpdate` as schema

**View Modes**
  - [x] User can toggle between two view modes: **Form** and **YAML** using Tabs
  - [x] The toggle is visible and accessible on the WorkspaceKind edit page

  **YAML View**
  - [x] YAML view displays the full WorkspaceKind configuration as valid YAML
  - [x] YAML is editable with a code editor (syntax highlighting, monospace font) 
  - [x] Changes made in YAML view are submitted when the user clicks Save

  **Form View**
  - [x] Existing expandable form continues to work as-is

  **Save & Validation**
  - [x] Save button remains disabled until changes are detected (existing `hasChanges` behavior)
  - [x] Invalid YAML displays a clear error message before or on submit
  - [x] Successful save calls `PUT /workspacekinds/:kind` and navigates back to the WorkspaceKinds list

  **General**
  - [x] The form view sees minimal changes to its code
  - [x] Accessible to keyboard and screen reader users
  - [x] Covered by Cypress E2E tests for the new view toggle and YAML editing flow
  
  **Future Work**
   - [x] Supports adding a **Split** (form + YAML side-by-side) view mode in the future without major refactoring
   - [ ] Changes made in Form view are reflected if the user switches to YAML view (or vice versa)
   - [ ] Split View: Editing the WorkspaceKind's YAML will reflect changes in the Form and vice versa
  - [ ] The design should not block a future split view where the form and YAML editor are shown side-by-side
  - [ ] Split view should render defaults and required fields from the form into the YAML pane (live or on open)

PR #1171, opened 2026-06-09
feat: Allow Admin to Edit WorkspaceKind Via Yaml

thaorell:
closes: #1105

### YAML Editor
  - Add Form/YAML tab toggle on the edit page using PatternFly Tabs with proper TabContent for a11y compliance
  - New WorkspaceKindYamlEditor component wrapping @patternfly/react-code-editor with monaco-yaml for autocomplete, hover docs, and inline
  validation
  - JSON Schema extracted from backend/openapi/swagger.json provides type-aware suggestions for all WorkspaceKindUpdate fields including nested
  Kubernetes types

 ### Form→YAML Sync
  - Switching to the YAML tab generates YAML from the current form state
  - Switching back to the Form tab preserves form state independently (no lossy round-trip)
  - Preserve original spec fields for imageConfig/podConfig values during Form→Update conversion to prevent phantom changes on in-use configs

###  Revert Button
  - Add "Revert" button in the footer that resets both form and YAML state to the original API data

###  Monaco Setup
  - Configure @monaco-editor/react loader to use local monaco-editor package instead of CDN
  - Set up MonacoEnvironment.getWorker for the YAML worker at app entrypoint
  - Add monaco-editor to webpack CSS includes for dev and prod builds

###  Schema Generation
  - Add scripts/generate-schema.js to extract and bundle the JSON Schema from the Swagger spec
  - Integrate into scripts/generate-api.sh so the schema regenerates alongside the TypeScript API client

### DIFF
--- a/workspaces/frontend/src/app/pages/WorkspaceKinds/Form/WorkspaceKindForm.tsx
+++ b/workspaces/frontend/src/app/pages/WorkspaceKinds/Form/WorkspaceKindForm.tsx
@@ -454,8 +454,9 @@ export const WorkspaceKindForm: React.FC = () => {
                 eventKey={YAML_TAB_KEY}
                 activeKey={activeTabKey}
                 hidden={activeTabKey !== YAML_TAB_KEY}
+                style={{ flex: 1 }}
               >
-                <TabContentBody>
+                <TabContentBody style={{ height: '100%' }}>
                   <WorkspaceKindYamlEditor
                     value={editYamlValue}
                     onChange={handleYamlChange}
--- a/workspaces/frontend/src/app/pages/WorkspaceKinds/Form/WorkspaceKindForm.tsx
+++ b/workspaces/frontend/src/app/pages/WorkspaceKinds/Form/WorkspaceKindForm.tsx
@@ -7,6 +7,7 @@ import { Flex, FlexItem } from '@patternfly/react-core/dist/esm/layouts/Flex';
 import { PageGroup, PageSection } from '@patternfly/react-core/dist/esm/components/Page';
 import { Stack, StackItem } from '@patternfly/react-core/dist/esm/layouts/Stack';
 import { Tabs, Tab, TabTitleText } from '@patternfly/react-core/dist/esm/components/Tabs';
+import { UndoIcon } from '@patternfly/react-icons/dist/esm/icons/undo-icon';
 import { useNotification } from 'mod-arch-core';
 import useGenericObjectState from 'mod-arch-core/dist/utilities/useGenericObjectState';
 import useWorkspaceKindByName from '~/app/hooks/useWorkspaceKindByName';
@@ -195,6 +196,19 @@ export const WorkspaceKindForm: React.FC = () => {
     },
     [activeTabKey, data, initialFormData],
   );
+
+  const handleRevert = useCallback(() => {
+    if (initialFormData) {
+      const converted = convertToFormData(initialFormData);
+      replaceData(converted);
+      setOriginalFormData(converted);
+      const yamlStr = yaml.dump(initialFormData, { noRefs: true });
+      setOriginalYaml(yamlStr);
+      setEditYamlValue(yamlStr);
+      setYamlParseError(null);
+      setError(null);
+    }
+  }, [initialFormData, replaceData]);
 
   const handleYamlChange = useCallback((value: string) => {
     setEditYamlValue(value);
@@ -445,6 +459,14 @@ export const WorkspaceKindForm: React.FC = () => {
               {mode === 'create' ? 'Create' : 'Save'}
             </Button>
           </FlexItem>
+          {mode === 'edit' && (
+            <FlexItem>
+              <Button variant="link" onClick={handleRevert} data-testid="revert-button">
+                <UndoIcon />
+                Revert
+              </Button>
+            </FlexItem>
+          )}
           <FlexItem>
             <Button variant="link" onClick={cancel} data-testid="cancel-button">
               Cancel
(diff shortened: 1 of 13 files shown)


paulovmr (changes requested):
Hi @thaorell , great feature! During testing, I verified that a change in the form is reflected in the YAML when I switch tabs, but the opposite does not happen, could you check that please? Thanks!

thaorell:
@paulovmr Yes. that is intended (we discussed it during a WG meeting that a one way conversion is sufficient for now and we will open a ticket to support the opposite direction in the future). This aligns with Patternfly's guidance for dual Form/Code Editor toggle support as well:
Phase 1: No conversion (Both Form and YAML manages their own states)
Phase 2: Form -> YAML (we are here)
Phase 3: Bidirectional

What I can do is including the message about the behavior in the UI, and subsequently remove it when a future PR is open for Phase 3. I have explored going all the way to Phase 3 but it seems too big for an introductory feature.

workspaces/frontend/src/app/pages/WorkspaceKinds/Form/yamlEditor/WorkspaceKindYamlEditor.tsx
christian-heusel, line 1: It seems like the radio buttons are a bit big in this overview:

Image

christian-heusel, line 1: What does the revert button do when nothing was changed?
Perhaps it should only be activated when there was a change (same as the "save" button already does)

Image

christian-heusel, line 1: It seems like the edit currently seems to assume that an image (in the form of a URL) is mandatory:
Image

However the CRD accepts either an image or a config map Key:
https://github.com/kubeflow/notebooks/blob/f3d8b803dc74366d98d197960c23986d133d9298/workspaces/controller/manifests/kustomize/samples/jupyterlab_v1beta1_workspacekind.yaml#L27-L36

IMO it's fine to only support the URL-based one in the wizard, but maybe there should be a note somewhere? 🤔

christian-heusel, line 1: It's quite non-obvious if an edit is failing server-side validation, as one first needs to manually scroll up:

Image

Maybe a error state on the button or automatic scroll up would be good?

christian-heusel: The error that caused this was setting a label without value, maybe that could be addressed separately?

christian-heusel, line 1: This currently doesn't seem to do anything, but also does not fail:

Image

I wonder whether the volume mount modal is even the right thing to use here given that this is only about a path..?
https://github.com/kubeflow/notebooks/blob/f3d8b803dc74366d98d197960c23986d133d9298/workspaces/controller/manifests/kustomize/samples/jupyterlab_v1beta1_workspacekind.yaml#L134-L140

christian-heusel (commented):
@thaorell This is awesome stuff, thanks a lot for working on this feature 🔥 

I left a few review comments for things that confused me as part of my user testing, for the lack of a better place I just commented on `workspaces/frontend/src/app/pages/WorkspaceKinds/Form/yamlEditor/WorkspaceKindYamlEditor.tsx` 😄

But overall this seems to work surprisingly well, great work 🚀

workspaces/frontend/src/app/pages/WorkspaceKinds/Form/yamlEditor/WorkspaceKindYamlEditor.tsx
thaorell: This is out of scope for the PR

thaorell: This is also out of scope and could be in a follow up.

thaorell: Same here

workspaces/frontend/src/app/pages/WorkspaceKinds/Form/WorkspaceKindForm.tsx
```diff
@@ -341,6 +482,19 @@ export const WorkspaceKindForm: React.FC = () => {
               {mode === 'create' ? 'Create' : 'Save'}
             </Button>
           </FlexItem>
+          {mode === 'edit' && (
+            <FlexItem>
+              <Button
+                variant="link"
+                isDisabled={!canSubmit}
```
paulovmr: The Revert button uses `isDisabled={!canSubmit}`, which ties it to the Save button's logic (including validation state and YAML parse errors). This means Revert is disabled even when there are actual changes if, for example, the YAML has a parse error — but the user likely wants to revert precisely in that situation. Revert should be disabled based on `!hasChanges` instead of `!canSubmit`, so users can revert invalid states back to the original.

paulovmr (changes requested):

paulovmr (approved):
/lgtm

workspaces/frontend/src/app/pages/WorkspaceKinds/Form/yamlEditor/WorkspaceKindYamlEditor.tsx
thaorell: Hey @christian-heusel, I've put everything together in this issue: https://github.com/kubeflow/notebooks/issues/1232
However, the button seems properly sized according to Material UI's guidance (https://mui.com/material-ui/react-radio-button), so I left this out of the issue.

Issue #1172, opened 2026-06-09
RStudio WorkspaceKind sample shows 404 on connect


christian-heusel:
### Checks

- [x] I have searched the [existing issues](https://github.com/kubeflow/notebooks/issues).
- [x] My issue is related to one of the components in the [`kubeflow/notebooks`](https://github.com/kubeflow/notebooks) repository.

### Kubeflow Notebooks Version

v2.0.0-alpha.2 post (02c5bd25f81d95345334335dcdd87293761df931)

### Kubeflow Platform

Tilt

### Kubernetes Distribution

KinD

### Kubernetes Version

```shell
Client Version: v1.36.1
Kustomize Version: v5.8.1
Server Version: v1.35.0
```

### Description

When setting up the WorkspaceKind samples via the `make tilt-up` functionality the RStudio one currently shows a 404 Not Found when pressing the "connect" button.

Steps to reproduce:
1. `make tilt-up` in `./developing`
2. Apply the samples `kubectl kustomize workspaces/controller/manifests/kustomize/samples/ | kubectl apply -f -`
3. Create a Workspace of type RStudio and go to "connect"
4. Observe the 404

<img width="2560" height="1380" alt="Image" src="https://github.com/user-attachments/assets/2d4fe177-1536-49b9-bd6e-7cb9e906ba51" />

This is the rendered VirtualService for the Workspace:
```yaml
kubectl get VirtualServices ws-test123-j8f62 -o yaml
apiVersion: networking.istio.io/v1
kind: VirtualService
metadata:
  creationTimestamp: "2026-06-09T23:10:04Z"
  generateName: ws-test123-
  generation: 1
  labels:
    notebooks.kubeflow.org/workspace-name: test123
  name: ws-test123-j8f62
  namespace: default
  ownerReferences:
  - apiVersion: kubeflow.org/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: Workspace
    name: test123
    uid: 693be2a0-53e4-4c8a-aeba-a250d78a5916
  resourceVersion: "1678"
  uid: 05005cd0-bb67-4303-a425-d10dc72e298a
spec:
  gateways:
  - kubeflow/kubeflow-gateway
  hosts:
  - '*'
  http:
  - headers:
      request:
        set:
          X-RStudio-Root-Path: /workspace/connect/default/test123/rstudio/
    match:
    - uri:
        prefix: /workspace/connect/default/test123/rstudio/
    route:
    - destination:
[...]

PR #1173, opened 2026-06-09
fix: strip path prefix in VirtualService when removePathPrefix is true

christian-heusel:
The rewrite logic in `generateVirtualServiceHTTPRoute()` was accidentially inverted: When `removePathPrefix` was `true`, no rewrite was emitted, so Istio forwarded the full '/workspace/connect/{namespace}/{name}/{port_id}/' path to the upstream container unchanged.

Now a `rewrite: {uri: "/"}` is emitted when `removePathPrefix` is `true`, so the matched prefix is actually stripped. When `removePathPrefix` is false, no rewrite is needed at all, since Istio forwards the original URI by default (the previous rewrite of the matched prefix to itself was a no-op).

Fixes https://github.com/kubeflow/notebooks/issues/1172

### DIFF
--- a/workspaces/controller/internal/controller/workspace_controller.go
+++ b/workspaces/controller/internal/controller/workspace_controller.go
@@ -992,10 +992,12 @@ func (r *WorkspaceReconciler) generateVirtualServiceHTTPRoute(
 	matchUriPrefix := getWorkspaceConnectPath(workspace.Namespace, workspace.Name, imageConfigPort.Id)
 
 	// determine rewrite configuration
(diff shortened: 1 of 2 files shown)


andyatmiami (approved):
/lgtm
/approve

Appreciate you fixing this faulty logic @christian-heusel .

Confirmed via Tilt that an RStudio workspace can now be spun up and successfully connected to (which was not possible with the current codebase on `HEAD` of `notebooks-v2` branch)

Image

Image

PR #1174, opened 2026-06-10
chore: bump frontend Node.js from 20 to 24 LTS

christian-heusel:
Node 24 entered LTS in October 2025 and ships with a new enough npm
that the manual upgrade step in CI is no longer needed.

### DIFF
--- a/.github/workflows/ws-frontend-test.yml
+++ b/.github/workflows/ws-frontend-test.yml
@@ -32,16 +32,6 @@ jobs:
           node-version-file: workspaces/frontend/package.json
           cache-dependency-path: workspaces/frontend/package-lock.json
 
-      # TODO: Remove this once it is possible to specify a npm version for
-      # `actions/setup-node` or the node version that we're using is new enough
-      # to ship with npm >= 11
-      #
-      # https://github.com/actions/setup-node/issues/529#issuecomment-4179457319
-      - name: Upgrade npm
-        working-directory: workspaces/frontend
-        # npm pkg get engines.npm -> "^11.10.0"
-        run: npm install --global npm@$(npm pkg get engines.npm | tr --delete '"')
-
       - name: Install dependencies
         working-directory: workspaces/frontend
         run: npm ci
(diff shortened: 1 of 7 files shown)


paulovmr (changes requested):
Hi @christian-heusel , changes lgtm, only one comment though, please consider including the changes in the `workspaces/frontend/package-lock.json` after the `npm install` is executed in your branch, changing the node version. Thanks!

thaorell (commented):
/lgtm

I was able to run `npm run start:dev` on my local environment, but like @paulovmr mentioned, there is a change in `package-lock.json`

thesuperzapper:
@christian-heusel if the `package-lock.json` file gets changed by running `npm install` / `npm ci` why is the porcelain check not triggering?

paulovmr:
@thesuperzapper @christian-heusel I believe the `npm ci` used by the workflow does not update the `package-lock.json` file because it was a change in the engine, and not in a dependency. Then there was no changes for the porcelain check to detect.

`npm ci` does a clean install from the lock file:

  1. Removes `node_modules/` entirely
  2. Installs dependencies exactly as specified in `package-lock.json` — no resolution, no upgrades
  3. Fails if `package-lock.json` is out of sync with `package.json` (dependency mismatches)
  4. Never writes to `package-lock.json`

That last point is why CI didn't catch it. Unlike `npm install` (which resolves, updates, and rewrites the lock file), `npm ci` treats the lock file as read-only. So there's no diff for the "uncommitted changes" check to find.

christian-heusel:
I have run `npm install` and added the missing changes to `packages-lock.json`, sorry for missing that initially! 🤗 

I think generally it's more sane to use `npm ci` in the context of CI, but as pointed out by Paulo this leads us to miss this kind of situation. Maybe we should just externalize it from the main CI into another check that does just `npm install` and the porcelain check?

paulovmr (approved):
/lgtm
/approve

andyatmiami (approved):
/lgtm
/approve

Relying on @paulovmr for the `/frontend` review and signoff.

I am comfortable with the `ws-frontend-test.yml` workaround being removed as node24 always shipped with some version of `npm` at v11 

Image
- https://nodejs.org/en/about/previous-releases

Issue #690, opened 2025-10-29
[TASK] Support spec.rules[] in WorkspaceKind CRD for compatibility selectors


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

We want to update the WorkspaceKind CRD to allow specifying a `spec.rules[]` element as outlined in #683 .

The controller only needs to verify the data structure (if present) is _syntactically correct_ .  It is the `backend` component that will be utilizing the information contained within `spec.rules[]` (and be handled in a separate PR).

### Acceptance Criteria

[ ]`controller` can validate/ reconcile a WorkspaceKind CRD that contains `spec.rules[]` as outlined in https://github.com/kubeflow/notebooks/issues/633

andyatmiami:
💡 I had Claude take a spin out outlining a possible implementation plan - this is **NOT A HARD REQUIREMENT TO IMPLEMENT VERBATIM** - but rather can be thought of as a _directionally correct guide_ to help contributors new to the repo understand the layout of the codebase and be aware of the general area where code contributions will be required...

---

All work is in `workspaces/controller/`.

1. Add types to `api/v1beta1/workspacekind_types.go`

Add a FilterRules field to WorkspaceKindSpec:

```
// +kubebuilder:validation:Optional
FilterRules []FilterRule `json:"filterRules,omitempty"`
```

Define the supporting types with kubebuilder markers. Follow the existing patterns in the file:

```
// +kubebuilder:validation:Enum=WORKSPACE_KIND;POD_CONFIG;IMAGE_CONFIG
type FilterRuleScope string

const (
    FilterRuleScopeWorkspaceKind FilterRuleScope = "WORKSPACE_KIND"
    FilterRuleScopePodConfig     FilterRuleScope = "POD_CONFIG"
    FilterRuleScopeImageConfig   FilterRuleScope = "IMAGE_CONFIG"
)

type FilterRule struct {
    Scope  FilterRuleScope    `json:"scope"`
    Effect FilterRuleEffect   `json:"effect"`
    // +kubebuilder:validation:MinItems:=1
    Match  []FilterRuleMatch  `json:"match"`
}

type FilterRuleEffect struct {
    // +kubebuilder:validation:Optional
    UI  *FilterRuleEffectUI  `json:"ui,omitempty"`
    // +kubebuilder:validation:Optional
[...]

PR #1175, opened 2026-06-12
feat: add spec.filterRules to WorkspaceKind CRD for compatibility selectors

Jefftree:
closes: #690

Adds the spec.filterRules[] field and supporting types to the WorkspaceKind CRD so the backend API can drive compatibility-selector filtering. The controller persists and validates the rules but does not evaluate them, per the issue.

The webhook validates each label selector. Structural constraints (scope enum, at-least-one match condition per entry, at-least-one effect, denyMessage only with deny) are enforced by the CRD schema via kubebuilder markers and CEL.


christian-heusel:
@Jefftree why did this PR get closed? 🤗

Jefftree:
someone else will be taking over 😃

Issue #1184, opened 2026-06-13
Disable cpu and memory selection when GPUs are selected


hahahannes:
### Checks

- [x] I have searched the [existing issues](https://github.com/kubeflow/notebooks/issues).
- [x] My request is related to one of the components in the [`kubeflow/notebooks`](https://github.com/kubeflow/notebooks) repository.

### Motivation

At the moment, users can select their own cpu and memory resources when creating a new notebook via the UI form.  This is not compatible with a setup where the platform administrators want to set these values. To disable them, the readonly configuration can be set but this will disable the possibility to set them completely. Instead it would be nice if they could be disabled based on whether a GPU is selected or not. This would be helpful in the case when cpu and memory resources are preset to a ratio of the node's resources based on the number of requested GPUs. But it will still allow selecting cpu and memory resources when the notebook runs on nodes for cpu workloads. 

### Implementation

I would add a configuration value to allow platform administrators to disable cpu and memory selection in the frontend when a GPU is selected. To not confuse the user, the fields should be greyed out.

For the users, this could look like this:

<img width="796" height="348" alt="Image" src="https://github.com/user-attachments/assets/91a9011f-41f6-4a8e-a316-648034f27d1d" />

For the platform administrators, the config could look like this:
```yaml
 cpu:
    readOnly: false
    value: '0.5'
    limitFactor: '1.2'
    disableIfGPU: true
  memory:
    readOnly: false
    value: 1.0Gi
    limitFactor: '1.2'
    disableIfGPU: true
```
  

### Are you willing & able to help?

- [x] I am able to submit a PR!
- [x] I can help test the feature!

kr1xh:
Hey @hahahannes , I’d like to work on this issue. I’ve gone through the proposed behavior and would like to investigate the existing resource selection flow and implement the disableIfGPU behavior, along with the necessary tests

PR #1185, opened 2026-06-13
feat: allow disabling cpu and memory selection if a gpu is selected during notebook creation; closes: #1184

hahahannes:
This PR will add the possibility to disable cpu and/or memory selection during notebook creation when a GPU is selected. See #1184 for more details. The change should be non breaking as it adds two additional configuration values `disableIfGPU` in the cpu and memory sections.


hahahannes:
Looking to notebooks-v2 as well at the moment.

Issue #989, opened 2026-03-25
[TASK] UI improvements in the WorkspaceForm Properties section


thaorell:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

After the demo on the WorkGroup on March 19th, we identified the following areas of improvements in the WorkspaceForm Properties section:

- ReadOnly is visible on Home Volume modals
- Access modes are currently radio buttons, which is hard for user to navigate
- State message is missing on Workspace table

### Acceptance Criteria

- [ ] Hide ReadOnly on Home Volume modals
- [ ] Make access mode selection a dropdown
- [ ] Wrap the state with a tooltip to display state message on WS table
- [ ] Write tests

PR #1219, opened 2026-06-25
feat: UI improvements for WorkspaceForm and WorkspaceTable

thaorell:
closes: #989

Descriptions
This PR targets a few UI improvements that were discussed in the working group meeting
* Hide Read Only Access from Home Volume Modals
* Convert Access Mode Select from Radio to DropDown
Screenshot 2026-06-25 at 6 11 30 PM

* Wrap the state with a tooltip to display state message on WS table
Screenshot 2026-06-25 at 6 10 58 PM

### DIFF
--- a/workspaces/frontend/src/__tests__/cypress/cypress/tests/mocked/workspaces/volumesManagement.cy.ts
+++ b/workspaces/frontend/src/__tests__/cypress/cypress/tests/mocked/workspaces/volumesManagement.cy.ts
@@ -261,7 +261,7 @@ describe('Volumes Management - Attach and Create', () => {
       volumesManagement.clickCreateVolume();
       volumesCreateModal.assertModalVisible();
 
-      volumesCreateModal.assertAccessModeChecked('ReadWriteOnce');
+      volumesCreateModal.assertAccessModeSelected('ReadWriteOnce (RWO)');
     });
 
     it('should create a volume and display it in the table', () => {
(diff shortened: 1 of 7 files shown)


christian-heusel (approved):
Just did a functional test of this, can't comment on the code itself 🤗

workspaces/frontend/src/__tests__/cypress/cypress/tests/mocked/workspaces/volumesManagement.cy.ts
```diff
@@ -421,4 +421,26 @@ describe('Volumes Management - Attach and Create', () => {
...
+
+      volumesCreateModal.assertReadOnlySwitchExists();
+      volumesCreateModal.clickCancel();
+    });
+
+    it('should hide read-only toggle when editing a home volume', () => {
+      // The Home Volume section is always expanded above the Data Volumes section.
+      // The mock workspace update returns the mount path as the home PVC identifier,
+      // which the form reconstructs as pvcName.
+      cy.findAllByTestId('volumes-table').first().find('[aria-label="plain kebab"]').click();
```
paulovmr: This test writes UI interaction logic directly rather than using a page object method. The `find('[aria-label="plain kebab"]')` call also uses a CSS attribute selector instead of `data-testid`. The page object already has  `openRowKebabMenu()` and `clickEditAction()` — consider adding a method to the page object for the home volume table (the existing `findVolumesTable()` targets the last table, but this test needs the first), then call that method here instead of inlining the interaction.

paulovmr (changes requested):

paulovmr (approved):
/lgtm
/approve

Issue #1224, opened 2026-06-26
[TASK] Integration of `kubeflow/dashboard` in the Tilt setup


christian-heusel:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

As a intermediate step towards https://github.com/kubeflow/notebooks/issues/1205 (as proposed via #300) we should setup integration of `kubeflow/dashboard` as a first step.

### Acceptance Criteria

- [ ] add a `./developing/manifests/kubeflow-dashboard` folder, and include it in the tilt defitions (same as this from kcd: https://github.com/kubeflow/community-distribution/blob/master/applications/dashboard/overlays/istio/kustomization.yaml)
- [ ] Gate it via an env-var (called `ENABLE_DASHBOARD`) which is enabled by default, and when enabled also adds a `/` link in the port-forwarded gateway in tilt
- [ ] fail if `ENABLE_FRONTEND` is not true, do something, like warn or fail

christian-heusel:
Done with the merge on #1225 ✔️

Issue #1205, opened 2026-06-21
[FEATURE] Setup optional Development environment that closely mirrors the Community Distribution Setup


christian-heusel:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Motivation

In order to facilitate developers with an environment where they can easily develop against a "more complete" setup we should add an option to start tilt with the components from the Kubeflow Community Distribution. This allows us to take advantage of all the neat parts of using tilt regarding the developer experience (autorebuild, ease of setup etc.) while enabling easily reproduce some of the bugs reported against the downstream manifest setup.

### High Level Design / Mock-ups

- New environment for `make tilt-up` (similar to the `ENABLE_PROMETHEUS` from #1140)
- Include the following components (most likely via URL):
    - Dashboard
    - oauth2-proxy
    - dex

### Acceptance Criteria

- [ ] There is an option to deploy tilt with the Kubeflow Community Distribution components for dex, oauth2-proxy and dashboard
- [ ] The components are all pinned to a fixed release to ensure reproducibility across developer setups
- [ ] There is documentation on how to use the setup in the Developer Guide
- [ ] The previously existing workflows for tilt stay functionally equivalent

PR #1225, opened 2026-06-26
feat: Setup `kubeflow/dashboard` in Tilt

christian-heusel:
Closes https://github.com/kubeflow/notebooks/issues/1224
Related-to https://github.com/kubeflow/notebooks/issues/1205


developing/Tiltfile
thesuperzapper: Lets move this to the end of the file (not because it will do anything, but because it will always be the last thing applied, as it depends on dashboard being up when its enabled)

thesuperzapper: For BOTH of these we will need to apply these common resources in each of the profile namespaces when dashboard is enabled.

This will probably be the `example-profile-{1,2,3,4}` as created in the other new manifest.

NOTE, because the profile namespaces wont exist until after the profiles have been applied AND reconciled by the dashboard, we need to somehow wait for this in tilt before applying these.

developing/manifests/kubeflow-dashboard/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
thesuperzapper, line 1: Lets make a new `developing/manifests/profiles` which contains 4 profiles with the `anonymous@kubeflow.org` as the owner, to avoid the user-creation flow and to allow us to pre-populate the namespaces created by the profiles with test resources.

These resources MUST depend on the kubeflow-dsahboard ones being already ready (so the CRD exists).

```yaml
apiVersion: kubeflow.org/v1
kind: Profile
metadata:
  name: example-profile-1
spec:
  owner:
    kind: User
    name: anonymous@kubeflow.org
  resourceQuotaSpec: {}
```

thesuperzapper (changes requested):

developing/Tiltfile
```diff
@@ -338,6 +305,51 @@ local_resource(
...
+        profile["metadata"]["name"] = new_name
+        k8s_yaml(encode_yaml(profile))
+        profile_names.append(new_name)
+
+    k8s_resource(
+        objects=[p for p in profile_names],
+        new_name="profiles",
+        labels=['dashboard'],
+        resource_deps=["profiles-deployment", "profile-crd-established"],
+    )
```
thesuperzapper: Rather than supporting additional profiles in the tiltfile, lets instead make the `manifests/profiles` into a kustomize target with 4 profile definitions, so that people can add/change profiles in their dev without understanding tilt (and also lets us have different profile definitions).

After applying them from this kustomize, we can then create a "wait" gate on the namespace being created like this (but looping over the profile yaml output to get each namespace name, which is the same as the profile name):

```
    local_resource(
        "profile-namespaces-ready",
        cmd=" && ".join([
            "kubectl wait --for=create namespace/%s --timeout=120s" % ns
            for ns in XXXXXXXXX
        ]),
        resource_deps=["profiles", "profile-crd-established"],
        labels=["???????"],
    )
```

developing/Tiltfile
```diff
@@ -351,3 +363,134 @@ if enable_prometheus:
...
+#
+# Common sample resources (ServiceAccount, PVCs, Secret, ConfigMap) applied to
+# one namespace per generated profile so workspaces can be created out of the
+# box; falls back to the "default" namespace when the dashboard (and thus
+# profile generation) is disabled.
+
+load('ext://namespace', 'namespace_inject')
+
+rbac_path = os.path.join(tilt_root, "manifests/rbac")
+common_samples_path = os.path.join(controller_dir, "manifests/kustomize/samples/common")
```
thesuperzapper: Lets split this into its own "section" of the tiltfile called "profile-resources", which will either:
1. If the dashboard is enabled, we ONLY apply these common resources in the `example-profile-1` namespace.
2. If the dashboard is not enabled, we apply these common resources in the `default` namespace (like we currently do).

developing/Tiltfile
```diff
@@ -351,3 +363,134 @@ if enable_prometheus:
...
+# resources the backend also checks (PVCs, secrets, etc).
+#
+# Common sample resources (ServiceAccount, PVCs, Secret, ConfigMap) applied to
+# one namespace per generated profile so workspaces can be created out of the
+# box; falls back to the "default" namespace when the dashboard (and thus
+# profile generation) is disabled.
+
+load('ext://namespace', 'namespace_inject')
+
+rbac_path = os.path.join(tilt_root, "manifests/rbac")
```
thesuperzapper: Lets create two new manifests folders under `developing/manifests`:

1. `rbac-standalone`
      - (just rename the current `manifets/rbac/` folder to this)
2. `rbac-dashboard`
     - `admin.yaml` (same as before, because its a clusterrolebinding)
     - `user.yaml` (hard-coded to have role bindings reference `example-namespace-1`, that is, the user only has access to `example-namespace-1`)

Create a new tilt section called "RBAC" which applies `rbac-standalone` when dashboard is disabled, and `rbac-dashboard` when its enabled.

These resources should be gated on `profile-namespaces-ready`.

thesuperzapper (changes requested):

christian-heusel:
@andyatmiami this now has the changes from @thesuperzapper applied, could you also give it a review given that you know the tilt setup quite well? 😊

PR #1226, opened 2026-06-27
chore: use server-side apply for CRDs in tilt and makefile

andyatmiami:
ℹ️ _NO GH ISSUE_

---

The WorkspaceKinds CRD (~316K) exceeds the 262KB
last-applied-configuration annotation limit for client-side kubectl apply. This hasn't caused failures yet because Tilt's k8s_yaml applies resources through its own mechanism, but the CRD will continue to grow and the controller Makefile's install and deploy targets use kubectl apply directly.

Adopt the pattern used by external-secrets-operator: split CRDs from non-CRD resources at the Tiltfile level and apply them separately with server-side apply.

## Development environment (developing/)

- Update Tiltfile to decode the rendered kustomize output and separate CRDs from non-CRD resources using decode_yaml_stream. CRDs are written to a scratch file in .tilt-tmp/ and applied via a local_resource using kubectl apply --server-side --force-conflicts, followed by kubectl wait --for=condition=Established to block until CRDs are registered. Non-CRD resources continue through standard k8s_yaml.

- Add apply-crds as a resource dependency on workspaces-controller, ensuring the controller deployment does not start until CRDs are fully established. This eliminates transient errors on cold starts where webhooks or RBAC referencing CRD types could be applied before the CRDs existed.

- Add .tilt-tmp/ directory to .gitignore for scratch files generated during Tiltfile evaluation.

## Controller Makefile (workspaces/controller/)

- Update make install to use --server-side --force-conflicts when applying CRDs from base/crd, preventing annotation size limit failures when installing CRDs standalone.

- Update make deploy to pipe kustomize build output through kubectl apply --server-side --force-conflicts instead of using kubectl apply -k (which uses client-side apply). This is functionally equivalent but avoids the annotation size limit for all resources in the overlay.

### DIFF
--- a/developing/Tiltfile
+++ b/developing/Tiltfile
@@ -51,9 +51,32 @@ else:
     controller_kustomize_path = os.path.join(controller_dir, "manifests/kustomize/overlays/istio")
 manifests = kustomize(controller_kustomize_path, kustomize_bin=kustomize_bin)
 
-# Apply manifests
+# Split CRDs from controller manifests for server-side apply.
+# The WorkspaceKinds CRD (~316K) exceeds the 262KB last-applied-configuration
+# annotation limit for client-side apply, so CRDs must use server-side apply.
+controller_objects = decode_yaml_stream(manifests)
+controller_crds = [o for o in controller_objects if o.get("kind") == "CustomResourceDefinition"]
+controller_non_crds = [o for o in controller_objects if o.get("kind") != "CustomResourceDefinition"]
+
+# Write CRDs to a scratch file for kubectl to apply.
+# This file is regenerated each time the Tiltfile re-evaluates (triggered by
+# kustomize() watching the manifests directory for changes).
+if controller_crds:
+    local("mkdir -p .tilt-tmp", quiet=True)
+    local("cat > .tilt-tmp/crds.yaml", stdin=encode_yaml_stream(controller_crds), quiet=True)
+
+# Apply CRDs via server-side apply and wait for establishment.
+local_resource(
+    "apply-crds",
+    cmd="kubectl apply --server-side --force-conflicts -f .tilt-tmp/crds.yaml"
+        + " && kubectl wait --for=condition=Established crd --all --timeout=60s",
+    deps=[".tilt-tmp/crds.yaml"],
+    labels=["controller"],
+)
+
+# Apply non-CRD controller resources via standard k8s_yaml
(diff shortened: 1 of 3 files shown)


developing/Tiltfile
```diff
@@ -51,9 +51,32 @@ else:
...
+controller_objects = decode_yaml_stream(manifests)
+controller_crds = [o for o in controller_objects if o.get("kind") == "CustomResourceDefinition"]
+controller_non_crds = [o for o in controller_objects if o.get("kind") != "CustomResourceDefinition"]
+
+# Write CRDs to a scratch file for kubectl to apply.
+# This file is regenerated each time the Tiltfile re-evaluates (triggered by
+# kustomize() watching the manifests directory for changes).
+if controller_crds:
+    local("mkdir -p .tilt-tmp", quiet=True)
+    local("cat > .tilt-tmp/crds.yaml", stdin=encode_yaml_stream(controller_crds), quiet=True)
```
christian-heusel, line 66: Any reason we're not using the established `.output` directory here? 🤔 
Compare i.e. https://github.com/kubeflow/notebooks/blob/f3d8b803dc74366d98d197960c23986d133d9298/workspaces/controller/Makefile#L150-L157

developing/Tiltfile
```diff
@@ -51,9 +51,32 @@ else:
...
+
+# Apply CRDs via server-side apply and wait for establishment.
+local_resource(
+    "apply-crds",
+    cmd="kubectl apply --server-side --force-conflicts -f .tilt-tmp/crds.yaml"
+        + " && kubectl wait --for=condition=Established crd --all --timeout=60s",
+    deps=[".tilt-tmp/crds.yaml"],
+    labels=["controller"],
+)
```
christian-heusel, line 75: How sure are you that we actually need this? 🤔 

From what I can tell there is some recent mechanism in tilt that should be able to deal with thin on it's own:
https://github.com/tilt-dev/tilt/pull/6680

https://docs.tilt.dev/api.html#api.update_settings

christian-heusel: It seems like this is first included with [`v0.37.0`](https://github.com/tilt-dev/tilt/releases/tag/v0.37.0) (released in 2026-03):
```
$ git tag --contains 95c1ec6b878557b1d6550294a230de456335dd9c | cat
v0.37.0
v0.37.1
v0.37.2
v0.37.3
v0.37.4
```

Issue #1222, opened 2026-06-25
[FEATURE] Workspace Pod Metrics


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Motivation

Bella needs to see how her workspaces are consuming CPU and memory to optimize resource allocation
for her models and experiments. Without this visibility, she's flying blind — she has no way to
know if she requested too many resources (wasting cluster capacity) or too few (risking OOM kills
or throttling) until something breaks.

Joel needs resource utilization visibility to identify capacity issues before they impact users.
When Bella reports "my workspace is slow," Joel's first question is whether the workspace is
resource-constrained — and today he has to leave the Notebooks UI and use separate tooling to
answer that.

This feature introduces resource usage metrics to the workspace UI, querying a suitable Kubernetes
metrics source and displaying point-in-time CPU and memory usage in the workspace detail view.
When no metrics source is available, the UI degrades gracefully with a clear explanation.

### High Level Design / Mock-ups

**Key constraints guiding the design:**
- Metrics source must not impose burdensome infrastructure requirements on consumers adopting
  Notebooks v2
- Point-in-time metrics are the starting point; historical/time-series data can layer on later
  (see Epic #884)
- Graceful degradation is required — workspaces must be fully functional without a metrics source
- Pod-level metrics only (node-level and cluster-aggregate metrics are out of scope)

### Acceptance Criteria

- [ ] Backend provides a mechanism to query workspace pod resource usage (CPU, memory)
- [ ] Backend detects metrics source availability and returns clear status when unavailable
- [ ] Frontend displays resource usage with request/limit context in the workspace detail view
- [ ] Frontend gracefully handles the metrics source not being available
- [ ] Metrics display does not block or slow down workspace page loading
- [ ] Solution does not require consumers to deploy additional infrastructure beyond what is
      commonly already available in Kubernetes clusters
- [ ] Unit tests cover metrics retrieval, availability detection, and error handling

richabanker:
Sharing an initial proposal [here](https://docs.google.com/document/d/1vlrz0ARQVoa_hYftMoDkSXvpAZRVvD3UOTMSBEQivQ0/)

andyatmiami:
/close

all work delivered and acceptance criteria met

(subsequent feature will be opened up to track work for node metrics)

PR #1228, opened 2026-07-01
feat: integrate metrics-server in tilt

christian-heusel:
Add opt-in metrics-server support to the Kind/Tilt development
environment, gated by the `ENABLE_METRICS_SERVER` env var, so that the
metrics.k8s.io APIService can be verified end-to-end.

Add `setup-metrics-server.sh` tries to follow the same pattern as
`setup-prometheus.sh` and by adding a kustomize overlay that patches the
upstream `components.yaml` with `--kubelet-insecure-tls` (required since
Kind's kubelet serving certs aren't valid for metrics-server to verify).

Related to https://github.com/kubeflow/notebooks/issues/1222 (cc @richabanker)

---

Verified via:
```
$ kubectl get --raw "/apis/metrics.k8s.io/v1beta1/nodes" | jq ".items[].usage"
{
  "cpu": "87224497n",
  "memory": "1757708Ki"
}
{
  "cpu": "144742938n",
  "memory": "4121184Ki"
}

$ kubectl top pod
NAME                              CPU(cores)   MEMORY(bytes)
ws-jupyterlab-workspace-v5nr9-0   0m           138Mi
ws-test123-sd4ff-0                0m           145Mi
```

### DIFF
--- a/developing/manifests/metrics-server/kustomization.yaml
+++ b/developing/manifests/metrics-server/kustomization.yaml
@@ -1,0 +1,17 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.8.1/components.yaml
+
+patches:
+# Kind's kubelets serve metrics over TLS certificates that aren't valid for
+# the address metrics-server connects to, so verification must be disabled.
+# This is only safe for local development, never for production.
+- target:
+    kind: Deployment
+    name: metrics-server
+  patch: |-
+    - op: add
+      path: /spec/template/spec/containers/0/args/-
+      value: --kubelet-insecure-tls
--- a/developing/manifests/metrics-server/kustomization.yaml
+++ b/developing/manifests/metrics-server/kustomization.yaml
@@ -8,10 +8,30 @@ patches:
 # Kind's kubelets serve metrics over TLS certificates that aren't valid for
 # the address metrics-server connects to, so verification must be disabled.
 # This is only safe for local development, never for production.
+#
+# Selects the container by name (rather than by array index) so the patch
+# doesn't break if upstream reorders or adds containers to the Deployment.
+#
+# Args are modified from upstream deployment, those might require updates when
+# the metrics-server component is being updated
+# https://github.com/kubernetes-sigs/metrics-server/blob/9ce65a47f9262e19e482c259ace815f01135a898/manifests/base/deployment.yaml#L20-L33
 - target:
     kind: Deployment
     name: metrics-server
   patch: |-
-    - op: add
-      path: /spec/template/spec/containers/0/args/-
-      value: --kubelet-insecure-tls
+    apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+      name: metrics-server
+    spec:
+      template:
+        spec:
+          containers:
+          - name: metrics-server
+            args:
+              - --cert-dir=/tmp
+              - --secure-port=10250
+              - --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname
+              - --kubelet-use-node-status-port
+              - --metric-resolution=15s
+              - --kubelet-insecure-tls
--- a/developing/scripts/setup-metrics-server.sh
+++ b/developing/scripts/setup-metrics-server.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+# Setup script for metrics-server
+# This script checks if metrics-server is installed and installs it if needed.
+# It lets developers verify `kubectl top` and the metrics.k8s.io APIService
+# work end-to-end.
+
+set -euo pipefail
+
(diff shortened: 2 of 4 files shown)


developing/manifests/metrics-server/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
...
+patches:
+# Kind's kubelets serve metrics over TLS certificates that aren't valid for
+# the address metrics-server connects to, so verification must be disabled.
+# This is only safe for local development, never for production.
+- target:
+    kind: Deployment
+    name: metrics-server
+  patch: |-
+    - op: add
+      path: /spec/template/spec/containers/0/args/-
```
andyatmiami: I think we should use the "selector syntax" similar to:
- https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/controller/manifests/kustomize/overlays/istio/kustomization.yaml#L41C8-L41C51

developing/scripts/setup-metrics-server.sh
```diff
@@ -0,0 +1,38 @@
...
+else
+  echo "Installing metrics-server..."
+  kubectl apply -k "${DEVELOPING_DIR}/manifests/metrics-server"
+fi
+
+echo "Waiting for metrics-server to be ready..."
+kubectl wait --for=condition=available deployment/metrics-server \
+  --namespace=kube-system \
+  --timeout=120s
+
```
andyatmiami, line 38: Maybe also after `deployment` wait on `pod` (?)

andyatmiami (changes requested):
minor tweaks - but overall this looks really good.

developing/manifests/metrics-server/kustomization.yaml
christian-heusel: I didn't find a way to use the syntax used there but selected the container by name instead 🤗

developing/manifests/metrics-server/kustomization.yaml
```diff
@@ -0,0 +1,33 @@
...
+    kind: Deployment
+    metadata:
+      name: metrics-server
+    spec:
+      template:
+        spec:
+          containers:
+          - name: metrics-server
+            args:
+              - --cert-dir=/tmp
```
richabanker, line 32: Should we add a comment that is list of args is copied over from the defaults of [metrics-server v0.8.1](https://github.com/kubernetes-sigs/metrics-server/releases/tag/v0.8.1) and might need refreshing if doing a version bump ?

developing/scripts/setup-metrics-server.sh
```diff
@@ -0,0 +1,42 @@
...
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+DEVELOPING_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+# Do nothing if ENABLE_METRICS_SERVER is not set to true
+if [[ "${ENABLE_METRICS_SERVER:-false}" != "true" ]]; then
+  echo ""
+  echo ""
+  echo "INFO: metrics-server setup is disabled. Set ENABLE_METRICS_SERVER=true to enable."
```
richabanker, line 17: Why not just do a re-apply `kubectl apply -k` ? Would help reconcile the manifest if changed ?

richabanker:
Minor comments, but I tested this on my own local and it worked for me, so happy with the addition! Thanks for the proactive help in setting up a tilt-compatible version of the metrics-server!

developing/scripts/setup-metrics-server.sh
christian-heusel: Are you referencing the code a bit further down?
https://github.com/christian-heusel/notebooks/blob/90209aaca5445239782ee31cb97cfa0bfa21affc/developing/scripts/setup-metrics-server.sh#L23-L28

richabanker: oh yes, didn't tag the right lines for the comment, sorry about that. But yes basically when we detect that a deployment is up already, we may still want to apply an updated config

developing/manifests/metrics-server/kustomization.yaml
christian-heusel: Done in the latest changeset ✅

developing/scripts/setup-metrics-server.sh
andyatmiami: the logic implemented here is consistent with how the various other dependencies we install function.. so i'd lean towards _consistency_ right now (although we can certainly discuss this more outside the context of _this particular PR_ )

given tilt is a "lightweight" development environment - i'm (personally) comfortable with the fact people can simply blow away their tilt cluster and re-apply to get updates..

andyatmiami (approved):
/lgtm
/approve

---

verified both `make tilt-up` and `ENABLE_METRICS_SERVER=true make tilt-up` do _what they are supposed to do_ .

excited to have this foundation in so @richabanker can have her work more easily surfaced to community/users 💯

Issue #1123, opened 2026-05-22
[TASK] Add activity `status` fields to Workspace CRD


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

🛑 Blocked by:  activityProbe schema task (#1122 ), activityRules schema task (#840 )                                                                                                                                       
                                                          
Expands the `Workspace` CRD status fields to support the activity and culling information defined in the [Workspace Culling design 
doc](https://github.com/kubeflow/notebooks/issues/839#issuecomment-4270494091). These fields will be populated by the controller in
issues mentioned above; this task defines the schema only.                                                                                
                                                          
The new status fields enable:                                                                                                      
- **Probe result tracking** — last probe start/end time, result (Success/Failure/Timeout), and a human-readable message
- **Rule evaluation state** — the `eligibleAfter` timestamp for the `pauseWorkspace` effect, computed from the first matching      
activity rule                                                                                                                      
- **Running duration tracking** — `lastRunningTime` records when the Workspace last entered a Running state, used by               
`activityRules` to enforce `minRunningSeconds` grace periods                                                                       
                                                          
All timestamp fields use **UNIX epoch in milliseconds**. The existing `LastActivity`, `LastUpdate`, and `PauseTime` fields         
currently use seconds — these must be updated to milliseconds for consistency (acceptable since v2 is pre-release with no
production users).                                                                                                                 
                                                          
#### Scope

**In scope:**
- Go type definitions in `workspace_types.go`
- Kubebuilder validation annotations
- Update any existing tests or code that reference the changed timestamp fields                                                    
- Regenerated CRDs, DeepCopy methods, and manifests                                                                                
                                                                                                                                   
**Out of scope:**                                                                                                                  
- Controller logic to populate these fields 
- Backend API changes to expose these fields                                                            
- Frontend UI changes

### Acceptance Criteria

- [ ] Workspace CRD status is expanded with `lastRunningTime`, `activity.lastProbe`, and `activity.rules` fields as outlined in the agreed   
proposal                                                                                                                           
- [ ] all timestamp fields across WorkspaceStatus use UNIX epoch in milliseconds consistently (including existing `pauseTime`,           
`activity.lastActivity`, `activity.lastUpdate`)                                                                                        
- [ ] any existing controller code or tests referencing the updated timestamp fields are updated to reflect the millisecond semantics

Issue #1191, opened 2026-06-18
[FEATURE] Structured Workspace Status


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Motivation

The current Workspace status model (`state` enum + freetext `stateMessage`) does not scale. As the
controller grows to support Activity Rules, filter rules, and other operational signals, the single
flat enum becomes increasingly inadequate for both machine consumption and human understanding.

This feature restructures Workspace status to:

1. **Introduce structured sub-states** on the CRD status that independently track distinct aspects
   of Workspace health (pod scheduling, container readiness, config validity, etc.)
2. **Derive a summary state deterministically** from the sub-states, maintaining a single top-level
   enum the frontend can always rely on
3. **Make the state generation process testable** by separating I/O (event fetching, API calls) from
   pure state derivation logic
4. **Provide a consistent pattern** for future signals (Activity Rules probe status, filter rule
   effects) to feed into the status without requiring rework of the state machine each time

#### Design Goals

- **Summary enum stays** — the frontend MUST have a single `state` field with a sensible,
  small enum it can switch on for primary rendering (colors, icons, basic labels)
- **Sub-states are on the CRD** — computed by the controller, visible via `kubectl get ws -o yaml`,
  consumed by the backend as-is (no re-derivation needed)
- **Not necessarily a Conditions array** — Kubernetes-style `[]Condition` is one option, but named
  fields or a hybrid may be cleaner depending on the actual sub-states identified. The implementer
  should evaluate what shape best fits the problem.

#### Conceptual Structure (illustrative, not prescriptive)

The implementer should evaluate the actual sub-states needed, but the following illustrates the
*kind* of structure we're looking for:

```yaml
status:
  state: Running              # summary enum (primary frontend signal)
  stateMessage: "..."          # human-readable summary (kept for backward compat)

  # structured sub-states (shape TBD by implementer)
  # each tracks an independent signal contributing to the summary
  #
  # examples of things that might be sub-states:
  #   - config validity (is WK resolvable? are options valid?)
  #   - resource generation (did STS/Service/VS generate correctly?)
  #   - pod scheduling (scheduled? unschedulable? gated?)
  #   - container readiness (running? CrashLoopBackOff? ImagePullBackOff?)
  #   - pending restart (options diverged? WK template changed?)
  #
  # future additions (not in this feature's scope):
[...]

PR #1230, opened 2026-07-07
feat: add structured status sub-states to Workspace CRD

### LINKED ISSUE #1192: [TASK] Define Workspace Status Sub-State Schema on CRD
### Certification

### Description

Defines the CRD status schema for structured Workspace sub-states. This is the foundational
schema task — it establishes the types and fields, but does NOT change the controller's state
generation logic (that's the follow-up task).

#### What to Evaluate

Audit the current state generation to identify the distinct signals that should become sub-states.
Today, state is derived from:

**Config-level signals** (from `Reconcile()` early returns, lines 146–224):
- Is the referenced WorkspaceKind resolvable?
- Is the imageConfig option valid (exists, redirect chain resolves)?
- Is the podConfig option valid?
- Did StatefulSet generation succeed?
- Did Service generation succeed?
- Did VirtualService generation succeed?
- Are there duplicate owned resources (multiple StatefulSets/Services/VirtualServices)?
- Did controller reference setting succeed?

**Runtime signals** (from `generateWorkspaceState()`, lines 1201–1376):
- Is the workspace paused?
- Is the pod being deleted (terminating)?
- Is the pod scheduled? If not, why? (Unschedulable, SchedulingGated, SchedulerError)
- Is the pod ready and running?
- What is the main container's state? (CrashLoopBackOff, ImagePullBackOff, etc.)
- Are there warning events on the StatefulSet?
- Are there warning events on the Pod?

**Currently missing signals** (identified from WG meeting notes):
- Volume binding status — PVC bind failures are not surfaced (known bug)
- DRA resource claim binding — future signal for GPU/accelerator claims

**Existing standalone signals:**
- `pendingRestart` (bool) — options or WK template have diverged

The implementer should determine:
1. Which of these signals warrant their own sub-state vs. being grouped
2. What shape each sub-state takes (bool, enum, struct with reason/message, etc.)
3. Whether a Conditions array, named fields, or a hybrid is the best fit
4. How the summary `state` enum maps from the sub-states (decision table or priority rules)
5. Whether the existing 6-value summary enum is sufficient or should be refined (e.g., splitting
   "Error" into config-error vs runtime-error). This is a judgment call for the implementer —
   the constraint is that the frontend must have a manageable enum to switch on.

#### Scope

**In scope:**
- Go type definitions in `workspace_types.go`
- Kubebuilder validation annotations
- Document the summary state derivation rules (which sub-state combinations produce which summary)
- Regenerated CRDs, DeepCopy methods, and manifests
- Consider forward-compatibility with Activity Rules status fields (Epic #545, task #1123) — the
  sub-state pattern established here should be one that activity probe status can naturally extend

**Out of scope:**
- Changing the controller's state generation logic (next task)
- Backend or frontend changes
- Activity Rules status fields themselves (separate task #1123)

### Acceptance Criteria

- [ ] Workspace CRD status includes structured sub-state fields that capture all current state
      derivation signals
- [ ] Summary `state` enum derivation rules are documented (how sub-states map to summary)
- [ ] The sub-state structure is extensible — adding a new signal (e.g., activity probe) should
      be straightforward without restructuring existing sub-states
- [ ] `pendingRestart` is integrated into the sub-state structure (or explicitly kept separate
      with documented rationale)
- [ ] Regenerated CRDs, DeepCopy methods, and manifests are committed
- [ ] Existing tests that reference `WorkspaceStatus` fields compile after the schema change
      (may need type updates even if behavior changes come in the next task)

Jefftree:
Closes #1192.

Adds `status.conditions` to the Workspace CRD so clients read each health dimension independently instead of parsing `stateMessage`. This follows typical Kubernetes convention (a `metav1.Condition` list-map alongside a coarse summary, as `Pod` does with `phase` + `PodScheduled`/`Ready`) and replaces the v1 `Notebook` approach of mirroring raw container state. The summary `state` enum is unchanged and now derives from the conditions, so existing consumers are unaffected.

## Schema

`status.conditions []metav1.Condition` (list-map keyed on `type`). Active dimensions:

| Condition | `False` reasons |
|---|---|
| `ConfigResolved` | `WorkspaceKindNotFound`, `InvalidImageConfig`, `InvalidPodConfig` |
| `ResourcesProvisioned` | `StatefulSetGenerationFailed`, `ServiceGenerationFailed`, `VirtualServiceGenerationFailed`, `DuplicateResources`, `ControllerReferenceFailed` |
| `PodScheduled` | `Unschedulable`, `SchedulingGated`, `SchedulerError` |
| `WorkspaceReady` | `Pending`, `CrashLoopBackOff`, `ImagePullBackOff`, `PodWarningEvent`, `StatefulSetWarningEvent` |

Derivation rules are documented on the `WorkspaceConditionType` constants. New signals (e.g. volume binding, activity probes, DRA resource claims) can be added later as new condition types without restructuring the existing ones.

## Design decisions

Resolving the open questions in #1192:

- **Conditions array over named fields** (Q3): reasoning in Alternatives considered below.
- **Grouped into 4 lifecycle dimensions** (Q1) with the specific cause in `reason` (Q2), rather than one condition per raw signal. The four map to successive stages of bring-up, and each has a distinct owner when `False`: `ConfigResolved` is a user or admin config fix, `ResourcesProvisioned` is a controller or RBAC issue, `PodScheduled` is cluster capacity, `WorkspaceReady` is the image or app. This keeps the summary small while making a failing dimension actionable and routable.
- **Kept `state` at 6 values, did not split `Error`** (Q5): the extra detail lives in the conditions, so more enum values would just repeat it and give the frontend more to handle.
- **`pendingRestart` left as-is, not moved into `conditions`**: it's unrelated to workspace health.

**Open question:** conditions use `metav1.Condition`, so their timestamps are RFC3339 rather than the epoch-ms convention in #1123. Matching epoch-ms would mean defining our own condition struct instead of the standard Kubernetes type. Leaving them as RFC3339 for now, open to suggestions.

## Alternatives considered

**Named fields per dimension** (a bool or small enum for each, like `status.configResolved`). Even as enums, these only tell you what state each dimension is in, not why or since when, so you end up adding a reason and message beside every one and rebuilding a Condition by hand. And every new signal is another schema field instead of a new entry in the array.

**More `state` values** (splitting `Error` into config/scheduling/runtime, etc.). The problem is the enum can only ever say one thing at a time. Once two dimensions matter at once, or something orthogonal like a failing activity probe shows up on an otherwise-Running workspace, a single value has to pick one and drop the rest. Conditions let each dimension speak for itself, and `state` still gives the at-a-glance summary on top.


christian-heusel:
https://github.com/kubeflow/notebooks/issues/1191#issuecomment-5036606842
/hold

Issue #840, opened 2026-01-16
[TASK] Implement Activity Rules feature


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

🛑 Blocked by: #839 

We want a _functional implementation_ of the Activity Rules feature as outlined in #839 .

💡 Please be aware the "Activity Rules" is the name of the feature we will be delivering in Notebooks 2.0.  "Activity Rules" can be configured in such a way to enable functional parity with the "Culling" feature available in Notebooks 1.0 - but are more configurable and flexible to meet future use cases.

Changes are expected to be delivered across the `backend` and `controller` components.

### Acceptance Criteria

- [ ] `activityRules` can be defined on a `WorkspaceKind` CR in such a way to be at feature parity with the Culling feature of Notebooks 1.0
- [ ] `activityRules` configuration (and/or misconfiguration) cannot degrade the performance/stability of the Notebooks 2.0 application.

andyatmiami:
/close

addressed in https://github.com/kubeflow/notebooks/pull/1239

PR #1239, opened 2026-07-09
feat: Add activityRules schema to WorkspaceKind CRD

siyuanfoundation:
Adds the `spec.activityRules` field to the WorkspaceKind CRD

 closes: #840


workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -44,8 +46,72 @@ type WorkspaceKindSpec struct {
...
+	// the conditions under which this rule applies
+	// +kubebuilder:validation:Optional
+	Match *ActivityRuleMatch `json:"match,omitempty"`
+
+	// the action to take when the rule matches and its conditions are met
+	Effect ActivityRuleEffect `json:"effect"`
+}
+
+type ActivityRuleConfig struct {
+	// the number of seconds of inactivity before a Workspace is eligible for this rule's effect
```
andyatmiami, line 77: might be nice to add a comment why `16` is chosen as the default as it is probably non-obvious to a new contributor looking at this codebase...

workspaces/controller/manifests/kustomize/samples/codeserver_v1beta1_workspacekind.yaml
```diff
@@ -16,6 +22,16 @@ spec:
...
     serviceAccount:
       name: "default-editor"
+    activityProbe:
+      minProbeIntervalSeconds: 300
+      probeIntervalSeconds: 3600
+      podExec:
+        timeoutSeconds: 60
+        script: |-
+          #!/usr/bin/env bash
+          # A simple mock probe checking for user processes or file modifications
```
andyatmiami: The comment says "A simple mock probe checking for user processes or file modifications" but the script always reports has_activity: true without actually checking anything. This could confuse users who copy the sample. Consider either:
  - Changing the comment to: # A mock probe that always reports activity (replace with real detection logic)
  - Or adding a note that this is a placeholder

workspaces/controller/manifests/kustomize/samples/rstudio_v1beta1_workspacekind.yaml
```diff
@@ -16,6 +22,16 @@ spec:
...
     serviceAccount:
       name: "default-editor"
+    activityProbe:
+      minProbeIntervalSeconds: 300
+      probeIntervalSeconds: 3600
+      podExec:
+        timeoutSeconds: 60
+        script: |-
+          #!/usr/bin/env bash
+          # A simple mock probe checking for user processes or file modifications
```
andyatmiami: The comment says "A simple mock probe checking for user processes or file modifications" but the script always reports has_activity: true without actually checking anything. This could confuse users who copy the sample. Consider either:
  - Changing the comment to: # A mock probe that always reports activity (replace with real detection logic)
  - Or adding a note that this is a placeholder

workspaces/controller/internal/webhook/workspacekind_webhook_test.go
andyatmiami, line 1: Consider adding a test for `pauseWorkspace: false` as a no-op override.

The first-match-wins semantics allow a rule with `pauseWorkspace: false` to override a later catch-all `pauseWorkspace: true`. There's no test verifying this valid  use case is accepted.

andyatmiami (changes requested):
Thanks for getting this open @siyuanfoundation .

Few trivial nipitcks on the shape of this - otherwise its looking for good from a quick/static review.

I will look to take a deeper look today or tomorrow (and also spend more time looking at the inline comments similar to a pass I did on the probe PR).

workspaces/controller/api/v1beta1/workspacekind_types.go
siyuanfoundation: done

workspaces/controller/manifests/kustomize/samples/codeserver_v1beta1_workspacekind.yaml
siyuanfoundation: done

workspaces/controller/manifests/kustomize/samples/rstudio_v1beta1_workspacekind.yaml
siyuanfoundation: done

workspaces/controller/internal/webhook/workspacekind_webhook_test.go
siyuanfoundation: added some more test cases

workspaces/controller/internal/webhook/workspacekind_webhook.go
```diff
@@ -139,11 +139,16 @@ func (v *WorkspaceKindValidator) ValidateCreate(ctx context.Context, obj runtime
 	allErrs = append(allErrs, validateImageConfigRedirects(imageConfigIdMap, imageConfigRedirectMap)...)
 	allErrs = append(allErrs, validatePodConfigRedirects(podConfigIdMap, podConfigRedirectMap)...)
 
+	// validate activity rules
+	rulesErrs, rulesWarnings := validateActivityRules(workspaceKind)
+	allErrs = append(allErrs, rulesErrs...)
+	warnings := rulesWarnings
```
andyatmiami: I like how in `ValidateUpdate` we declared `warnings` alongside `allErrs` as a `var` - could we apply that same style here for consistency in this function?

```
var warnings admission.Warnings
```

workspaces/controller/internal/webhook/workspacekind_webhook.go
```diff
@@ -202,6 +208,14 @@ func (v *WorkspaceKindValidator) ValidateUpdate(ctx context.Context, oldObj, new
 		allErrs = append(allErrs, validateActivityProbe(newWorkspaceKind.Spec.PodTemplate.ActivityProbe, activityProbePath, podTemplatePortsIdMap)...)
 	}
 
+	// validate activity rules if rules or probe changed
+	if !equality.Semantic.DeepEqual(newWorkspaceKind.Spec.ActivityRules, oldWorkspaceKind.Spec.ActivityRules) ||
+		!equality.Semantic.DeepEqual(newWorkspaceKind.Spec.PodTemplate.ActivityProbe, oldWorkspaceKind.Spec.PodTemplate.ActivityProbe) {
+		rulesErrs, rulesWarnings := validateActivityRules(newWorkspaceKind)
+		allErrs = append(allErrs, rulesErrs...)
+		warnings = rulesWarnings
```
andyatmiami: While fine today - I worry this `warnings` assignment could cause a _future headache_ if we needed to assign/declare warnings _before this block_ .  Might be better to follow the `append(...)` pattern here...

workspaces/controller/internal/webhook/workspacekind_webhook.go
```diff
@@ -795,3 +808,85 @@ func validateActivityProbe(activityProbe *kubefloworgv1beta1.ActivityProbe, path
...
+
+		// selectors validation
+		if rule.Match != nil {
+			if rule.Match.MatchNamespace != nil {
+				errs = append(errs, v1validation.ValidateLabelSelector(&rule.Match.MatchNamespace.Selector, v1validation.LabelSelectorValidationOptions{}, rulePath.Child("match", "matchNamespace", "selector"))...)
+			}
+			if rule.Match.MatchPodConfig != nil {
+				errs = append(errs, v1validation.ValidateLabelSelector(&rule.Match.MatchPodConfig.Selector, v1validation.LabelSelectorValidationOptions{}, rulePath.Child("match", "matchPodConfig", "selector"))...)
+			}
+		}
```
andyatmiami, line 846:
```suggestion
		// selectors validation
		if rule.Match != nil {
		    matchPath := rulePath.Child("match")
			if rule.Match.MatchNamespace != nil {
				errs = append(errs, v1validation.ValidateLabelSelector(&rule.Match.MatchNamespace.Selector, v1validation.LabelSelectorValidationOptions{}, matchPath.Child("matchNamespace", "selector"))...)
			}
			if rule.Match.MatchPodConfig != nil {
				errs = append(errs, v1validation.ValidateLabelSelector(&rule.Match.MatchPodConfig.Selector, v1validation.LabelSelectorValidationOptions{}, matchPath.Child("matchPodConfig", "selector"))...)
			}
		}
```

workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -44,6 +46,79 @@ type WorkspaceKindSpec struct {
...
+
+	// the conditions under which this rule applies
+	// +kubebuilder:validation:Optional
+	Match *ActivityRuleMatch `json:"match,omitempty"`
+
+	// the action to take when the rule matches and its conditions are met
+	Effect ActivityRuleEffect `json:"effect"`
+}
+
+type ActivityRuleConfig struct {
```
andyatmiami, line 76:
```suggestion
// ActivityRuleConfig defines the timing parameters for an ActivityRule
type ActivityRuleConfig struct {
```

workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -44,6 +46,79 @@ type WorkspaceKindSpec struct {
...
+	// filters Workspaces by namespace labels
+	// +kubebuilder:validation:Optional
+	MatchNamespace *NamespaceMatch `json:"matchNamespace,omitempty"`
+
+	// filters Workspaces by the PodConfig option they are using
+	// +kubebuilder:validation:Optional
+	MatchPodConfig *PodConfigMatch `json:"matchPodConfig,omitempty"`
+}
+
+type NamespaceMatch struct {
```
andyatmiami, line 104:
```suggestion
// NamespaceMatch filters Workspaces by namespace labels
type NamespaceMatch struct {
```

workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -44,6 +46,79 @@ type WorkspaceKindSpec struct {
...
+	// the standard Kubernetes label selector to match namespace labels
+	Selector metav1.LabelSelector `json:"selector"`
+}
+
+type PodConfigMatch struct {
```
andyatmiami, line 110:
```suggestion
// PodConfigMatch filters Workspaces by the PodConfig option they are using
type PodConfigMatch struct {
```

workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -21,6 +21,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const DefaultProbeIntervalSeconds int32 = 3600
+
 // Important: Run "make" to regenerate code after modifying this file
```
andyatmiami:
```suggestion
// Important: Run "make" to regenerate code after modifying this file

// DefaultProbeIntervalSeconds is the fallback probe interval used when no activityProbe                                        
// is configured. This must stay in sync with the +kubebuilder:default marker on                                                
// ActivityProbe.ProbeIntervalSeconds. 
const DefaultProbeIntervalSeconds int32 = 3600
```

workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -44,6 +46,79 @@ type WorkspaceKindSpec struct {
...
+// ActivityRuleEffect defines the action to take. Each field represents a different effect type
+// (e.g., pauseWorkspace). Different effect types are evaluated independently by the controller.
+// To allow independent evaluation passes without shielding, a field must be explicitly set (true or false)
+// to be considered configured. Fields left as nil are ignored (fallthrough) during that effect's evaluation pass.
+//
+// For each effect type, an explicit 'false' matches the rule and terminates evaluation, overriding any later
+// rule (no-op override). A 'nil' value skips the rule entirely, letting evaluation fall through to subsequent rules.
+//
+// +kubebuilder:validation:XValidation:message="must specify at least one effect",rule="has(self.pauseWorkspace)"
+type ActivityRuleEffect struct {
```
andyatmiami:
```suggestion
// ActivityRuleEffect defines the action to take when a rule matches.
//
// Each field is a tri-state (`*bool`) controlling one effect type:
//  - `true`: apply the effect (e.g., pause the Workspace)
//  - `false`: do not apply the effect (e.g., exempt matching Workspaces from being paused)
//  - `nil`: skip this effect type entirely; evaluation continues to subsequent rules
//
// Both `true` and `false` terminate evaluation for that effect type.
// Only `nil` causes fallthrough to the next rule.
//
// +kubebuilder:validation:XValidation:message="must specify at least one effect",rule="has(self.pauseWorkspace)"
type ActivityRuleEffect struct {
```

workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -44,6 +46,79 @@ type WorkspaceKindSpec struct {
...
+	// determines if the Workspace should be paused. Requires activityProbe to be configured.
+	// +kubebuilder:validation:Optional
+	PauseWorkspace *bool `json:"pauseWorkspace,omitempty"`
```
andyatmiami:
```suggestion
    // determines if the Workspace should be paused
    //  - the webhook rejects rules with `pauseWorkspace: true`
    //    when no `activityProbe` is configured
	// +kubebuilder:validation:Optional
	PauseWorkspace *bool `json:"pauseWorkspace,omitempty"`
```

workspaces/controller/internal/webhook/workspacekind_webhook.go
```diff
@@ -795,3 +808,85 @@ func validateActivityProbe(activityProbe *kubefloworgv1beta1.ActivityProbe, path
...
+					errs = append(errs, field.Invalid(rulePath.Child("match"), rule.Match, "catch-all (empty match) rule must be the last rule configuring the pauseWorkspace effect"))
+				}
+			}
+		}
+
+		// Warning check: secondsSinceActive < 2 * probeIntervalSeconds
+		if rule.Config.SecondsSinceActive < 2*probeIntervalSeconds {
+			warnings = append(warnings, fmt.Sprintf("spec.activityRules[%d].config.secondsSinceActive (%d) is less than twice the probeIntervalSeconds (%d). This may cause Workspaces to be paused too quickly or prematurely.", i, rule.Config.SecondsSinceActive, probeIntervalSeconds))
+		}
+	}
```
andyatmiami, line 880: The warning text says "may cause Workspaces to be paused too quickly" but fires for ALL rules regardless of `PauseWorkspace` value. A rule with `PauseWorkspace: nil` (fallthrough) or `PauseWorkspace: false` (exemption) will never cause pausing, so the warning is misleading. Suggested fix:

```go
if rule.Effect.PauseWorkspace != nil && *rule.Effect.PauseWorkspace &&
    rule.Config.SecondsSinceActive < 2*probeIntervalSeconds {
```

workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -44,6 +46,79 @@ type WorkspaceKindSpec struct {
...
+	// the action to take when the rule matches and its conditions are met
+	Effect ActivityRuleEffect `json:"effect"`
+}
+
+type ActivityRuleConfig struct {
+	// the number of seconds of inactivity before a Workspace is eligible for this rule's effect.
+	// The minimum value is 16 (secondsSinceActive > 15) to prevent thrashing and culling workspaces
+	// prematurely during startup or transient connection drops.
+	// +kubebuilder:validation:Minimum:=16
+	SecondsSinceActive int32 `json:"secondsSinceActive"`
```
andyatmiami:
```suggestion
    // the number of seconds of inactivity before a Workspace is eligible for this rule's effect
    //  - the minimum value is 16 (`secondsSinceActive` > 15) to prevent thrashing and culling
    //    workspaces prematurely during startup or transient connection drops
	// +kubebuilder:validation:Minimum:=16
	SecondsSinceActive int32 `json:"secondsSinceActive"`
```

workspaces/controller/internal/webhook/workspacekind_webhook.go
```diff
@@ -795,3 +808,85 @@ func validateActivityProbe(activityProbe *kubefloworgv1beta1.ActivityProbe, path
...
+					}
+				}
+				if hasSubsequentPauseWorkspaceRule {
+					errs = append(errs, field.Invalid(rulePath.Child("match"), rule.Match, "catch-all (empty match) rule must be the last rule configuring the pauseWorkspace effect"))
+				}
+			}
+		}
+
+		// Warning check: secondsSinceActive < 2 * probeIntervalSeconds
+		if rule.Config.SecondsSinceActive < 2*probeIntervalSeconds {
```
andyatmiami: `2*probeIntervalSeconds` could overflow `int32` if `ProbeIntervalSeconds` is set to an appropriately absurd high number. There's no upper-bound kubebuilder marker on `ProbeIntervalSeconds` to prevent this. Extreme edge case but could produce an incorrect warning with a wrapped-negative comparison value.  We should define some reasonable upper bound to avoid the overflow potential - but i'm not really opinionated on _the exact value_ to use..

andyatmiami (changes requested):
Hey @siyuanfoundation - thanks for this contribution..

Overall really solid - some random/minor nits on edge cases and styling.. reach out if you disagree with anything!

As always, _words are hard_ - so please don't take my "suggestions" w.r.t. comment strings as hard requirements..  the formatting is important (relatively speaking) - but feel free to play around with any wordings as you see fit!

workspaces/controller/internal/webhook/workspacekind_webhook.go
siyuanfoundation: good point. Added upper bound of 1 year

workspaces/controller/internal/webhook/workspacekind_webhook.go
siyuanfoundation: good catch!

workspaces/controller/internal/webhook/workspacekind_webhook.go
siyuanfoundation: done

workspaces/controller/internal/webhook/workspacekind_webhook.go
siyuanfoundation: added `var warnings admission.Warnings`, but I had to do `warnings = rulesWarnings` instead of `append` to work around the lint error: direct assignment is more efficient than append in this unconditional case

andyatmiami (approved):
/lgtm
/approve

thanks for continuing to flesh out this feature @siyuanfoundation - exciting to see it becoming _more and more real_

lets get these CRD changes in so the real fun (i.e. business logic) can start to be worked!

---

testing notes here:
- [TEST-EVIDENCE-PR-1239.md](https://github.com/user-attachments/files/30150503/TEST-EVIDENCE-PR-1239.md)
    - tests were driven by Claude - but the plan and results were carefully reviewed by me!

PR #1240, opened 2026-07-09
ci: Add slash-commands for blocked-by relations

christian-heusel:
Also replace the manual GraphQL calls in `actions/github-script` with
Octokit calls.

Additionally post just a single result comment after the operations run instead
of posting a preview comment beforehand and linking to it.

---

Example issues on playground:
- https://github.com/christian-heusel/notebooks-playground/issues/127
- https://github.com/christian-heusel/notebooks-playground/issues/129

### DIFF
--- a/.github/workflows/slash-commands.yaml
+++ b/.github/workflows/slash-commands.yaml
@@ -15,6 +15,8 @@ jobs:
       && (
           contains(github.event.comment.body, '/add-sub-issue')
           || contains(github.event.comment.body, '/remove-sub-issue')
+          || contains(github.event.comment.body, '/add-blocked-by')
+          || contains(github.event.comment.body, '/remove-blocked-by')
       )
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/slash-commands.yaml
```diff
@@ -47,148 +49,119 @@ jobs:
...
-                  }
-                }
-              `);
-              return {
-                id: response.repository.issue.id,
-                title: response.repository.issue.title
-              };
+            const getIssue = async (owner, repo, issueNumber) => {
+              const { data } = await github.rest.issues.get({ owner, repo, issue_number: issueNumber });
+              return { id: data.id, title: data.title };
```
andyatmiami: Seems like with the REST API - we don't need this `data.title` attribute anymore - so we can probably drop this....

andyatmiami (changes requested):
Hey @christian-heusel - from a static analysis - this looks **really good** - and the flip from GraphQL to REST API is (imho) so much preferable to read 💯 

One minor nit - and in parallel i'm looking to test these functionalities on my fork just to "see" the behavior to be thorough - but don't anticipate any issues based on this well-structured PR.

andyatmiami (approved):
/lgtm
/approve

---

thanks @christian-heusel for this nice cleanup/refactor as well as introducing new slash commands to help with the "consumability" of our planning boards.  seeing issues that are "ready but blocked" will help alleviate a common pain point for people looking to pick up work.

testing was done on my forked repo and captured here to validate appropriate behavior:
- [PR-1240-TEST-RESULTS.md](https://github.com/user-attachments/files/30193561/PR-1240-TEST-RESULTS.md)
    - claude was used as a "test runner" - but the plan itself and results were validated by me!

Issue #1223, opened 2026-06-25
[TASK] Agree on architecture for workspace resource metrics


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

As a community, we need to agree on the architecture for exposing workspace resource metrics
(CPU, memory) in the Notebooks UI before implementation tasks can be created.

#### Problem

Bella wants to know if her workspace is resource-constrained — is it running out of memory, is CPU
throttled — so she can adjust her resource requests or restructure her workload. Joel wants to see
resource utilization across workspaces so he can identify capacity issues before users hit them.

Today, answering these questions requires leaving the Notebooks UI entirely and using separate
cluster monitoring tools. This feature should bring basic resource visibility directly into the
workspace experience.

#### What Needs to Be Decided

The design owner should propose and reach community consensus on:

1. **Metrics data source** — Where does the backend get CPU/memory usage data? The Kubernetes
   ecosystem has multiple options (Metrics Server, Prometheus, Custom Metrics API, etc.), each
   with different availability, capability, and infrastructure assumptions. The choice should
   consider what consumers of Notebooks v2 are likely to already have deployed.

2. **API design** — How are metrics exposed to the frontend? Consider whether metrics belong in
   existing workspace responses or warrant their own endpoint, and how the response schema should
   be structured.

3. **Graceful degradation** — Not all clusters will have a metrics source available. The design
   must define how the backend detects availability and how the frontend behaves when metrics
   data cannot be retrieved. The workspace UI must remain fully functional without metrics.

4. **Consumer burden** — Notebooks v2 is adopted by a range of consumers, from minimal Kubernetes
   clusters to fully instrumented production environments. The design should not impose
   infrastructure requirements that become an adoption barrier. If new dependencies are needed,
   the rationale should be clearly documented.

#### Constraints

- **Scope is point-in-time metrics** — historical data, time-series charts, and trend analysis
  are out of scope for this feature (see Epic #884 for future OTel/Prometheus work)
- **Pod-level only** — node-level and cluster-aggregate metrics are out of scope
- **No new CRD changes** — the controller already tracks pod status; this feature is backend
  API + frontend
- **RBAC-aware** — the solution must account for clusters where the backend's service account may
  not have permissions to read metrics

#### Expected Deliverable

A written design proposal (posted as a comment on this issue or linked document) covering the
[...]

PR #1241, opened 2026-07-10
feat: implement core metrics provider for workspace resource usage

richabanker:
This PR introduces the core models and repository logic for surfacing live, point-in-time pod resource utilization (CPU and Memory) for workspaces using K8s Metrics Server. This implements the foundation for the Resource Utilization Metrics proposal.

Changes:
- Added `internal/models/metrics` package containing domain models (`WorkspaceResourceUsage`) and schema mappings for the API response
- `internal/repositories/metrics` holds implementation for Repository.GetWorkspaceResourceUsage(), and a memoized `metrics.k8s.io` availability probe
- `internal/helper/k8s.io` Updated `helper.NewManager()` to disable caching for `PodMetrics`, as the aggregated API does not support WATCHes
- `manifests/kustomize/base/rbac.yaml` grants the backend ServiceAccount get/list permissions on `pods.metrics.k8s.io`
- Added unit tests 

Issue: https://github.com/kubeflow/notebooks/issues/1290
[Design](https://github.com/kubeflow/notebooks/issues/1223#issuecomment-5072632630)


richabanker:
I have kept this first PR intentionally scoped to just the data layer for the resource metrics API. Currently it adds no HTTP paths, handlers, Repositories wiring and RBAC changes (which I plan to split in smaller follow-ups). Please let me know if granularity works or if I should split this PR further into smaller chunks? 

Happy to go whichever way makes review easiest.

cc @andyatmiami @thesuperzapper

workspaces/backend/internal/helper/k8s.go
```diff
@@ -46,6 +47,9 @@ func BuildScheme() (*runtime.Scheme, error) {
 	if err := kubefloworgv1beta1.AddToScheme(scheme); err != nil {
 		return nil, fmt.Errorf("failed to add Kubeflow types to scheme: %w", err)
 	}
+	if err := metricsv1beta1.AddToScheme(scheme); err != nil {
+		return nil, fmt.Errorf("failed to add metrics types to scheme: %w", err)
+	}
```
christian-heusel, line 52: I think you also need to do this for the tests:



```
  [FAILED] in [BeforeSuite] - /home/chris/Documents/shared_projects/kion/kubeflow-notebooks/workspaces/backend/api/suite_test.go:132 @ 07/11/26 02:20:34.521
  << Timeline

  [FAILED] Unexpected error:
      <*fmt.wrapError | 0x17a49b88b620>:
      unable to create manager: no kind is registered for the type v1beta1.PodMetrics in scheme "pkg/runtime/scheme.go:100"
      {
          msg: "unable to create manager: no kind is registered for the type v1beta1.PodMetrics in scheme \"pkg/runtime/scheme.go:100\"",
          err: <*runtime.notRegisteredErr | 0x17a49b3c26c0>{
              schemeName: "pkg/runtime/scheme.go:100",
              gvk: {Group: "", Version: "", Kind: ""},
              target: nil,
              t: <*reflect.rtype | 0x2065a40>{
                  t: {Size_: 0x140, PtrBytes: 0x130, Hash: 614274612, TFlag: 7, Align_: 8, FieldAlign_: 8, Kind_: 25, Equal: nil, GCData: 85, Str: 253661, PtrToThis: 5271776},
              },
          },
      }
  occurred
  In [BeforeSuite] at: /home/chris/Documents/shared_projects/kion/kubeflow-notebooks/workspaces/backend/api/suite_test.go:132 @ 07/11/26 02:20:34.521
------------------------------
```

>  **🤖-generated reason**
> 
> `api/suite_test.go` builds its manager's scheme manually (`scheme.Scheme` + `kubefloworgv1beta1.AddToScheme`), not via `helper.BuildScheme()`. `helper.NewManager()` was changed in this PR to unconditionally add `&metricsv1beta1.PodMetrics{}` to `Cache.DisableFor`, which requires the passed-in scheme to know that GVK.

workspaces/backend/internal/models/metrics/funcs_test.go
```diff
@@ -0,0 +1,188 @@
...
+})
+
+func podWithContainer(name string, containers ...corev1.Container) corev1.Pod {
+	return corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: name},
+		Spec: corev1.PodSpec{
+			Containers: containers,
+		},
+	}
+}
```
christian-heusel, line 265: Should be fine to just `//nolint` here given that its in a test file:

```
$ make -C workspaces/backend lint
make: Entering directory '/home/chris/Documents/shared_projects/kion/kubeflow-notebooks/workspaces/backend'
/home/chris/Documents/shared_projects/kion/kubeflow-notebooks/workspaces/backend/bin/golangci-lint run
internal/models/metrics/funcs_test.go:168:23: `podWithContainer` - `name` always receives `"pod-1"` (unparam)
func podWithContainer(name string, containers ...corev1.Container) corev1.Pod {
                      ^
make: *** [Makefile:77: lint] Error 1
make: Leaving directory '/home/chris/Documents/shared_projects/kion/kubeflow-notebooks/workspaces/backend'
```

workspaces/backend/internal/models/metrics/types.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// Reason is a machine-readable code present only when Available is false.
+	Reason string `json:"reason,omitempty"`
+
+	// Message is a human-readable explanation present only when Available is false.
+	Message string `json:"message,omitempty"`
+
+	// Pods holds the per-pod usage.
+	Pods []PodResourceUsage `json:"pods"`
+}
```
christian-heusel: I'm not 100% sure with this, however I think we usually have API design that revolves around error states instead of success states, so the way that this types combines both paths (i.e. via `Available` & `Reason`) feels odd to me. However the last call regarding design decisions is of course with Andy and Mathew so would be good if they'd weigh in on this at some point 🤗 

See for example this snippet:
https://github.com/kubeflow/notebooks/blob/e51895dd4ba0bd43748b20b48b785d3b6dbd39ca/workspaces/backend/internal/models/common/assets/types.go#L17-L33

If we apply this pattern here, we would get something like this:
```golang
type WorkspaceResourceUsage struct {
    // Source identifies which provider produced the data (e.g. "metrics-server").
    // Empty when Error is set.
    Source string `json:"source,omitempty"`

    // Error is set when live usage metrics could not be retrieved.
    Error *ResourceUsageError `json:"error,omitempty"`

    // Pods holds the per-pod usage. Empty when Error is set.
    Pods []PodResourceUsage `json:"pods"`
}

type ResourceUsageError struct {
    Code    ResourceUsageErrorCode `json:"code"`
    Message string                 `json:"message,omitempty"`
}

type ResourceUsageErrorCode string

const (
    ResourceUsageErrorMetricsAPIUnavailable ResourceUsageErrorCode = "METRICS_API_UNAVAILABLE"
    ResourceUsageErrorWorkspaceNotRunning    ResourceUsageErrorCode = "WORKSPACE_NOT_RUNNING"
)
```

Also the `ImageRefErrorCode` values are in 'SCREAMING_SNAKE' casing ("CONFIGMAP_MISSING"), while this PR's constants are 'PascalCase' ("MetricsAPIUnavailable").

christian-heusel (commented):
Hey @richabanker, thanks a lot for driving this and starting out with an implementation of your proposal! 🥳 

I have left a few thoughts below!

/ok-to-test

workspaces/backend/internal/helper/k8s.go
richabanker: yep, I was not registering `metricsv1beta1` to the scheme used by the test client. Added it now.

workspaces/backend/internal/models/metrics/funcs_test.go
richabanker: Actually changed a test to pass a value other than "pod-1" for the name param. That should make it go away.

workspaces/backend/internal/models/metrics/types.go
richabanker: ah good point. I like the proposed schema, can pivot to this to keep things consistent once @andyatmiami or @thesuperzapper confirm. Thanks for this suggestion!

workspaces/backend/internal/repositories/metrics/repo.go
```diff
@@ -0,0 +1,134 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/config"
+	modelsmetrics "github.com/kubeflow/notebooks/workspaces/backend/internal/models/metrics"
+)
+
+const (
+	// LabelWorkspaceName is the label the WorkspaceController sets on every workspace pod
+	// (at the StatefulSet pod-template level).
+	LabelWorkspaceName = "notebooks.kubeflow.org/workspace-name"
```
andyatmiami: Move `LabelWorkspaceName` to `internal/models/common/funcs.go` and import as `modelsCommon.LabelWorkspaceName`

nice to colocate all our "special strings"
- https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/internal/models/common/funcs.go#L26-L37

workspaces/backend/internal/repositories/metrics/repo.go
```diff
@@ -0,0 +1,134 @@
...
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	metricsv1beta1 "k8s.io/metrics/pkg/apis/metrics/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/config"
+	modelsmetrics "github.com/kubeflow/notebooks/workspaces/backend/internal/models/metrics"
```
andyatmiami: per convention - the "primary" `models` package in the repo should be called `models` ... and then other models we may need to reference get a camelCased "suffix qualifier"

```suggestion
	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/metrics"
```

workspaces/backend/internal/repositories/metrics/repo.go
```diff
@@ -0,0 +1,134 @@
...
+	// (at the StatefulSet pod-template level).
+	LabelWorkspaceName = "notebooks.kubeflow.org/workspace-name"
+
+	// TTL for API availability checks.
+	apiAvailabilityTTL = 60 * time.Second
+)
+
+// Repository exposes point-in-time workspace resource utilization, read from the
+// Kubernetes Metrics Server, to the API layer.
+type Repository struct {
```
andyatmiami:
```suggestion
type MetricsRepository struct {
```

⚠️ This rename will then need to be applied to comments, constructor, etc

workspaces/backend/internal/models/metrics/funcs_test.go
andyatmiami, line 1: Might be nice to have a few test cases around `UsageForPod` as well here...

workspaces/backend/internal/models/metrics/funcs_test.go
```diff
@@ -0,0 +1,186 @@
...
+			Expect(got.Error).To(BeEmpty())
+			Expect(got.Containers).To(HaveLen(1))
+
+			cUsage := got.Containers["container-1"]
+			Expect(cUsage.MetricsFromMetricsServer).NotTo(BeNil())
+			Expect(cUsage.MetricsFromMetricsServer.Timestamp).To(Equal("2026-06-20T12:00:00Z"))
+			Expect(cUsage.MetricsFromMetricsServer.Usage.CPU).To(Equal("50m"))
+			Expect(cUsage.MetricsFromMetricsServer.Usage.Memory).To(Equal("100Mi"))
+			Expect(cUsage.Resources.Requests[corev1.ResourceCPU]).To(Equal(qtyCPU))
+			Expect(cUsage.Resources.Limits[corev1.ResourceCPU]).To(Equal(qtyCPU))
```
andyatmiami: `BeComparableTo` might be a little more _friendly to work with_ for these types of multiple asserts on a single "thing"

example: 
- https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/api/healthcheck_handler_test.go#L62-L68
- https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/api/pvcs_handler_test.go#L845-L881

workspaces/backend/internal/models/metrics/funcs.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	containers := make(map[string]ContainerResourceUsage, len(pod.Spec.Containers))
+	for _, c := range pod.Spec.Containers {
+		metrics := usageByContainer[c.Name]
+		containers[c.Name] = ContainerResourceUsage{
+			MetricsFromMetricsServer: metrics,
+			Resources:                c.Resources,
+		}
+	}
```
andyatmiami, line 49: `c.Resources` is copied by value, but its `Requests` and `Limits` fields are `corev1.ResourceList` (which is `map[ResourceName]Quantity`). The map references are shared with the original Pod object. Since this is a read-only response-building path, mutation risk is very low, but a defensive deep copy wouldn't _hurt anything_ here either..

wanted to at least call it out..

andyatmiami (changes requested):
Hey @richabanker - sorry its taken me so long to turn attention to this!

overall it looks pretty good - some minor nitpicks that are hopefully non-controversial to address... reach out to me if you disagree and/or want to discuss.

---

w.r.t. "what is the right PR size" - hehe - a good-but-loaded question... **personally** - I would prefer a PR results in some (positive!) change being evident in the application... for the sake of `*.go` code - that means I can do something new/better within the `tilt/kind` cluster...

now, we should totally just get this PR merged in more or less its current state (once feedback addressed) - but in the future (again, speaking personally) - I'd prefer some external "hooks" be in any PR to really enable "hands on testing"... i.e. CRD fields to set, APIs to call, etc etc

others may feel differently, and I'm happy to accommodate per your preferences.. just sharing my $0.02

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Done

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Done

workspaces/backend/internal/models/metrics/funcs_test.go
richabanker: oh missed that! Thanks for catching, added now.

workspaces/backend/internal/models/metrics/funcs_test.go
richabanker: Yeah definitely much cleaner, incorporated.

workspaces/backend/internal/models/metrics/funcs.go
richabanker: controller-runtime's `cachereader.List` already deep-copies every object, [ref](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/cache/internal/cache_reader.go#L171-L178) unless explicitly disabled, so I think we should be safe here without the additional deep-copy ?

andyatmiami (approved):
/lgtm
/approve

---

thanks @richabanker  for this foundational work and pivoting quickly based on the architecture discussions you drove...  

implementation aligns with repo conventions and the design doc... looking forward to the subsequent PR where we get some API handlers into the mix to really open this up! 💯

PR #1242, opened 2026-07-10
fix: prevent duplicate volumes when same PVC is mounted to multiple paths

Jaehyeon1020:
When the same PVC is mounted to multiple paths via the Jupyter Web App UI, `add_notebook_volume` appends a `V1Volume` with the same name for each mount. This causes a Kubernetes validation error (duplicate volume names in `spec.volumes`) and the Pod fails to start.

**Fix:** Track already-added volume names in the volume creation loop (`post.py`) to skip duplicate `add_notebook_volume` calls. `add_notebook_container_mount` is unchanged since multiple `volumeMounts` referencing the same volume is valid in Kubernetes.

**How to test:**
1. Create a new notebook with the same PVC mounted to two different paths (e.g. `/data` and `/backup`)
2. Verify the Pod starts successfully with one volume entry and two volume mounts

<sub>Re-opening #1023, which was auto-closed by the stale bot due to inactivity. No changes from the original.</sub>

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/default/routes/post.py
+++ b/components/crud-web-apps/jupyter/backend/apps/default/routes/post.py
@@ -57,6 +57,7 @@ def post_pvc(namespace):
         api.create_pvc(pvc, namespace, dry_run=True)
 
     # create the new PVCs and set the Notebook volumes and mounts
+    added_volumes = set()
     for api_volume in api_volumes:
         pvc = volumes.get_new_pvc(api_volume)
         if pvc is not None:
@@ -66,7 +67,10 @@ def post_pvc(namespace):
         v1_volume = volumes.get_pod_volume(api_volume, pvc)
         mount = volumes.get_container_mount(api_volume, v1_volume["name"])
 
-        notebook = volumes.add_notebook_volume(notebook, v1_volume)
+        if v1_volume["name"] not in added_volumes:
+            notebook = volumes.add_notebook_volume(notebook, v1_volume)
+            added_volumes.add(v1_volume["name"])
+
         notebook = volumes.add_notebook_container_mount(notebook, mount)
 
     log.info("Creating Notebook: %s", notebook)
--- a/components/crud-web-apps/jupyter/backend/apps/default/routes/post_test.py
+++ b/components/crud-web-apps/jupyter/backend/apps/default/routes/post_test.py
@@ -1,0 +1,113 @@
+import unittest
+from unittest import mock
+
+from flask import Flask
+
+from apps.default.routes import post
+
+
+FORM_SETTERS = (
+    "set_notebook_image",
+    "set_notebook_image_pull_policy",
+    "set_server_type",
+    "set_notebook_cpu",
+    "set_notebook_memory",
+    "set_notebook_gpus",
+    "set_notebook_tolerations",
+    "set_notebook_affinity",
+    "set_notebook_configurations",
+    "set_notebook_shm",
+)
+
+
+class TestPostPVC(unittest.TestCase):
+    def setUp(self):
+        self.app = Flask(__name__)
+
+    def test_same_pvc_mounted_at_multiple_paths_adds_one_volume(self):
+        api_volumes = [
+            {
+                "mount": "/data",
+                "existingSource": {
+                    "persistentVolumeClaim": {
+                        "claimName": "shared-pvc",
+                    },
+                },
+            },
+            {
+                "mount": "/backup",
+                "existingSource": {
+                    "persistentVolumeClaim": {
+                        "claimName": "shared-pvc",
+                    },
+                },
+            },
+        ]
+        notebook = {
+            "spec": {
+                "template": {
+                    "spec": {
+                        "containers": [{}],
+                    },
+                },
+            },
+        }
+
+        with self.app.test_request_context(
+            "/api/namespaces/test-namespace/notebooks",
+            method="POST",
+            json={"name": "test-notebook"},
+        ):
+            with (
+                mock.patch.object(
+                    post.helpers,
+                    "load_param_yaml",
+                    return_value=notebook,
+                ),
+                mock.patch.object(
+                    post.utils,
+                    "load_spawner_ui_config",
+                    return_value={},
+                ),
+                mock.patch.object(
+                    post.form,
+                    "get_form_value",
+                    side_effect=[api_volumes, None],
+                ),
+                mock.patch.object(
+                    post.authn,
+                    "get_username",
+                    return_value="user@example.com",
+                ),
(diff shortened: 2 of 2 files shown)


Snehadas2005:
The intent of this PR is great and addresses an edge case with multi-path PVC mounting. There is a small logical loop error in the code that needs fixing. Since this modifies the main notebook creation pipeline (`routes/post.py`), it would be ideal to back it up with an automated unit test in the backend to prevent future regressions. A mock payload with one PVC mapping to multiple `api_volumes` paths would be useful. Looking forward to your updates! 🚀

components/crud-web-apps/jupyter/backend/apps/default/routes/post.py
```diff
@@ -66,7 +67,10 @@ def post_pvc(namespace):
         v1_volume = volumes.get_pod_volume(api_volume, pvc)
         mount = volumes.get_container_mount(api_volume, v1_volume["name"])
 
-        notebook = volumes.add_notebook_volume(notebook, v1_volume)
+        if v1_volume["name"] not in added_volumes:
```
Snehadas2005, line 70: Hi @Jaehyeon1020! The architecture strategy to allow multiple `volumeMounts` while deduplicating the underlying `V1Volume` spec makes perfect sense. However, looking at the current loop implementation, there is a small logical oversight. The original line `notebook = volumes.add_notebook_volume(notebook, v1_volume)` is still executing unconditionally right above your new conditional block.  Because of this, the volume will be appended to the notebook array before the deduplication check even evaluates. To make the deduplication active, we should completely remove that unconditional call:

```suggestion
pvc = volumes.get_new_pvc(api_volume)
if pvc is not None:
    v1_volume = volumes.get_pod_volume(api_volume, pvc)
    mount = volumes.get_container_mount(api_volume, v1_volume["name"])

    # Only add the volume structure if it hasn't been appended yet
    if v1_volume["name"] not in added_volumes:
        notebook = volumes.add_notebook_volume(notebook, v1_volume)
        added_volumes.add(v1_volume["name"])
        
    notebook = volumes.add_notebook_container_mount(notebook, mount)
```

Could you please update the block so the `set()` actually guards the mutations? Thanks!

shivamsingh-007:
Welcome @Jaehyeon1020! Thanks for re-opening this fix — preventing duplicate volumes when the same PVC is mounted to multiple paths is a useful edge case to handle. Let me know if you need any help working through the feedback from @Snehadas2005!

Jaehyeon1020:
@Snehadas2005 @shivamsingh-007

Thanks for the feedback!

I've added a backend regression test for mounting the same PVC at multiple paths. It verifies that the Notebook spec contains one volume entry and multiple volume mounts. All backend unit tests are passing 😃

components/crud-web-apps/jupyter/backend/apps/default/routes/post.py
Jaehyeon1020: @Snehadas2005  Thanks for the feedback! However, the logic currently changed in this PR already performs a conditional check to prevent duplicate volumes from being added. Could you please take another look and clarify where you see the `notebook = volumes.add_notebook_volume(notebook, v1_volume)` line being executed unconditionally?

Snehadas2005: Hi @Jaehyeon1020,

Apologies for any formatting confusion in my previous suggestion layout! Here is a quick dry-run the exact execution flow of the current loop layout in the PR branch to clarify where the duplicate append happens::

```python
        # 1. This line runs on EVERY single loop iteration (completely unconditional)
        notebook = volumes.add_notebook_volume(notebook, v1_volume)
        
        # 2. This conditional check only guards the SECOND append and the tracking set
        if v1_volume["name"] not in added_volumes:
            notebook = volumes.add_notebook_volume(notebook, v1_volume)
            added_volumes.add(v1_volume["name"])
            
        notebook = volumes.add_notebook_container_mount(notebook, mount)

```

If a user mounts the exact same PVC (e.g., `"shared-pvc"`) to two different paths, the `for api_volume in api_volumes:` loop runs exactly twice:

1. **First Iteration (Path A):**
	* The first `add_notebook_volume` runs --> `"shared-pvc"` is added to the notebook spec.
	* The `if` condition evaluates to `True`  --> The second `add_notebook_volume` runs (adding it a second time), and `"shared-pvc"` is added to the `added_volumes` tracking set.


2. **Second Iteration (Path B):**
	* The first `add_notebook_volume` runs again unconditionally --> `"shared-pvc"` is added to the notebook spec a **third time**.
	* The `if` condition now evaluates to `False` (since it exists in the set) --> the inner block is skipped.



Because the first statement is completely unnested and outside the `if` guard, it mutates the notebook object on every loop sequence. To fix this and activate the deduplication correctly, the first unconditional call needs to be removed, so the code looks like this:

```python
        if v1_volume["name"] not in added_volumes:
            notebook = volumes.add_notebook_volume(notebook, v1_volume)
            added_volumes.add(v1_volume["name"])

```

Let me know if this tracing makes sense! Love the addition of the new unit tests in `post_test.py`. Once we wrap this call cleanly inside the conditional guard, the whole patch will be solid. 🙌

Jaehyeon1020: @Snehadas2005

Thanks for the detailed explanation. Walking through the execution flow helped me fully understand what you meant.

However, it looks like the unconditional call you’re referring to may be the line removed by this PR, as indicated by the “-” prefix in the diff. That line is no longer present in the current branch.

The only remaining `add_notebook_volume()` call is inside the conditional block:

```python3
if v1_volume["name"] not in added_volumes:
    notebook = volumes.add_notebook_volume(notebook, v1_volume)
    added_volumes.add(v1_volume["name"])
```

Therefore, `add_notebook_volume()` can execute at most once per iteration and is always guarded by the condition.

The regression test also confirms this behavior. If the unconditional call were still present, mounting the same PVC at two paths would produce three `"shared-pvc"` volume entries instead of the single expected entry, causing the test to fail.

Could you please check the current PR branch once more?

Snehadas2005 (commented):
Hi @Jaehyeon1020,

Ah, thank you so much for clarifying and catching that! You are completely right. I misread the git diff block due to the layout alignment in the unified view and thought the original call was still active above the conditional block.

Since the unconditional line is indeed removed and neatly enclosed within your guard statement, the deduplication logic works flawlessly as validated by your test cases. Great job on adding the comprehensive backend regression tests in post_test.py as well!

This looks solid and ready. Appreciate your patience while sorting this out.

Snehadas2005:
cc: @andyatmiami 
It would be great to have your review and approval on this PR.

PR #1246, opened 2026-07-10
refactor: Enable 'modernize' linter

christian-heusel:
Detailed descriptions are in each commit:

- **refactor: Enable the 'modernize' linter in `golangci-lint`**
- **refactor: Utilize `maps.Copy()` instead of `for`**
- **refactor: Replace `interface{}` with `any`**
- **refactor: Convert `ClusterOptionMetrics` to pointer**
- **refactor: Use the more efficient `strings.SplitSeq()`**
- **chore: Suppress `fmt.Appendf()` lint in test**

### DIFF
--- a/workspaces/backend/.golangci.yml
+++ b/workspaces/backend/.golangci.yml
@@ -23,6 +23,7 @@ linters:
     - ineffassign
     - lll
     - misspell
+    - modernize
     - nakedret
     - nolintlint
     - prealloc
--- a/workspaces/backend/api/helpers.go
+++ b/workspaces/backend/api/helpers.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
 	"mime"
 	"net/http"
 	"strings"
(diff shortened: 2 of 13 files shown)


andyatmiami (approved):
/lgtm
/approve

thanks for this nice code cleanup @christian-heusel .  agree with the changes - some notes below for posterity

---

- https://github.com/golang/go/blob/master/src/maps/maps.go#L62-L66
    - `maps.copy` is literally identical functionally to the code it is replacing

- `*ClusterOptionMetrics`
    - acceptable change that is aligned with original vision (that has _yet to be acted on_ - which is why this code appears confusing)
        - original thoughts were that we would have a sibling field `NamespaceMetrics` that would be present when Bella makes API calls (with `?namespaceFilter` query param) that scoped counts to the specific namespace Bella was operating against
        - in this scenario - `ClusterOptionMetrics` actually **SHOULD** be a `ptr` as it could/would be `nil` and not returned in API response
        - For now, without `NamespaceMetrics` defined - it doesn't really matter one way or another - and this change is just as reasonable (and arguably slightly more correct)

Issue #1234, opened 2026-07-08
[TASK] Changes in WorkspaceKind YAML Edit Should Reflect on the Form


thaorell:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

Currently, changes in the Form view are reflected in the YAML view, but not vice versa. As a step towards full functionality, changing the YAML should reflect in the Form for all fields that are rendered in the Form. Any fields edited in the YAML that do not have a corresponding Form control should be preserved silently and included when the user saves, without requiring the Form to render them.

### Acceptance Criteria

- [ ] Changes in YAML will be rendered in the Form when the user clicks the Form tab
- [ ] (Optional) Type errors in the YAML view should disable the Save button
- [ ] Tests

PR #1251, opened 2026-07-13
feat: Add support to convert from yaml to form in WorkspaceKind Edit

thaorell:
closes: #1234

- Added `lastApiObjectRef` to support bidirectional conversion when users toggle between Form and YAML when editing WorkspaceKind
- Added unit tests


paulovmr (approved):
/lgtm
/approve

Issue #1144, opened 2026-06-03
[TASK] Warn users when they select a redirected or hidden option in workspace create/update flow


paulovmr:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

When a user selects a redirected or hidden option and clicks "Next" in the workspace create/update flow, the UI should display a confirmation dialog that:

  1. Warns the user that their selected option has been redirected by an administrator
  2. Shows the redirect chain (e.g., Option A → Option B → Option C) so the user understands what will actually be applied
  3. Asks for confirmation: "Your admin has redirected the option you selected. Do you want to apply the redirect?"
  4. Provides options to confirm (proceed with the redirected option) or cancel (go back to select a different option)

For hidden options, the warning should indicate that the option has been deprecated/hidden and suggest selecting an alternative.

### Acceptance Criteria

- [ ] Clicking "Next" with a redirected option triggers a confirmation dialog
- [ ] The dialog shows the full redirect chain
- [ ] The user can confirm or cancel
- [ ] Hidden options show an appropriate deprecation warning
- [ ] If the user confirms, the final redirected option is used

PR #1254, opened 2026-07-15
feat: add redirect/hidden confirmation modal to workspace form

thaorell:
closes: #1144

Redirect confirmation modal integration
  - Added StepRedirectInfo discriminated union type to model redirect/hidden state for the current form step
  - Integrated WorkspaceFormRedirectConfirmModal into WorkspaceForm with handlers for apply, continue, and cancel actions
  - Extracted transformRedirectMessageLevel from WorkspaceFormOptionCard into the shared resolveRedirectChain utility to avoid duplication

  Cypress E2E tests
  - Added RedirectConfirmModal page object in workspaceForm.ts
  - Added 10 new tests covering image redirect (apply/continue/cancel), image hidden-only, pod config redirect (apply/cancel), and pod config
  hidden-only flows
  - Updated existing create and edit workspace tests to handle the new modal during step navigation

Examples:

Hidden image warning:
Screenshot 2026-07-15 at 1 50 12 PM

Redirection
Screenshot 2026-07-15 at 1 50 21 PM
 warning:

### DIFF
--- a/workspaces/frontend/src/__tests__/cypress/cypress/pages/workspaces/workspaceForm.ts
+++ b/workspaces/frontend/src/__tests__/cypress/cypress/pages/workspaces/workspaceForm.ts
@@ -1,3 +1,51 @@
+class RedirectConfirmModal {
+  find() {
+    return cy.findByTestId('redirect-confirm-modal');
+  }
+
+  assertModalVisible() {
+    return this.find().should('be.visible');
+  }
+
+  assertModalNotExists() {
+    cy.findByTestId('redirect-confirm-modal').should('not.exist');
+  }
+
+  findApplyRedirectButton() {
+    return this.find().findByTestId('apply-redirect-button');
+  }
+
+  clickApplyRedirect() {
+    return this.findApplyRedirectButton().click();
+  }
+
+  assertApplyRedirectButtonExists() {
+    return this.findApplyRedirectButton().should('exist');
+  }
+
+  assertApplyRedirectButtonNotExists() {
+    return this.find().findByTestId('apply-redirect-button').should('not.exist');
+  }
+
+  findContinueButton() {
+    return this.find().findByTestId('continue-button');
+  }
+
+  clickContinue() {
+    return this.findContinueButton().click();
+  }
+
+  findCancelButton() {
+    return this.find().findByTestId('cancel-button');
+  }
+
+  clickCancel() {
+    return this.findCancelButton().click();
+  }
+}
+
+const redirectConfirmModal = new RedirectConfirmModal();
+
 /**
  * Base class for workspace form page objects (create and edit).
  * Contains shared methods for interacting with the WorkspaceForm component.
@@ -282,6 +330,15 @@ class WorkspaceForm {
     return cy.findByTestId('summary-diff-new');
   }
 
+  /**
+   * Clicks Next and dismisses the redirect confirm modal by clicking "Keep Selection" (Continue).
+   * Use when navigating past a step that has a redirected or hidden option selected.
+   */
+  advancePastRedirectModal(): void {
+    this.clickNext();
+    redirectConfirmModal.clickContinue();
+  }
+
   findRedirectSummaryIcon(step: number, suffix: string): Cypress.Chainable<JQuery<HTMLElement>> {
     return cy.findByTestId(`redirect-icon-${step}-${suffix}`);
   }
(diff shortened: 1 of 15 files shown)


caponetto:
Check if these findings make sense:


1. **Description:** resolveRedirectChain() returns a non-undefined finalTarget even when a cycle (A→B→A) breaks the traversal, so the modal's 'Apply Redirect' button stays enabled for a non-terminal target.
**File:** workspaces/frontend/src/app/pages/Workspaces/Form/utils/resolveRedirectChain.ts
**Suggestion:** Return finalTarget as undefined (or a distinct cyclic/unresolved status) when the visited-set break fires, so the modal's isDisabled check can detect it.

1. **Description:** handleApplyRedirect() in WorkspaceForm.tsx updates imageConfig on redirect but never revalidates or clears the currently selected podConfig, allowing a stale, now-incompatible podConfig to be submitted.
**File:** workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
**Suggestion:** After applying an image redirect, re-check the selected podConfig against the refreshed compatible options (awaiting the refetch if needed) and clear/reset it if no longer valid.

1. **Description:** The hidden-option branch of WorkspaceFormRedirectConfirmModal warns that an option is deprecated/hidden but never suggests a replacement, missing issue #1144's acceptance criterion to suggest an alternative.
**File:** workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceFormRedirectConfirmModal.tsx
**Suggestion:** Surface a recommended non-hidden alternative option in the hidden-only modal branch, or link out to the option list filtered to non-hidden choices.

1. **Description:** WorkspaceFormOptionCard.tsx retains its own transformRedirectToChain() with no cycle detection, duplicating and potentially disagreeing with the shared resolveRedirectChain() used by the confirmation modal.
**File:** workspaces/frontend/src/app/pages/Workspaces/Form/shared/WorkspaceFormOptionCard.tsx
**Suggestion:** Replace transformRedirectToChain() with a call to the shared resolveRedirectChain() utility so the popover and confirmation modal render from a single source of truth.

1. **Description:** WorkspaceForm.tsx now mixes page rendering with redirect detection, modal state, and step-gating logic, making the component harder to reason about and extend.
**File:** workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
**Suggestion:** Extract the redirect-confirm flow (state, handlers, step-gating checks) into a dedicated hook, e.g. useRedirectConfirmation.

1. **Description:** The new step-gating state machine (currentStepRedirectInfo, nextStep, handleApplyRedirect, handleContinueWithWarning) has no RTL/unit test coverage and is exercised only via Cypress E2E specs against mocked fixtures.
**File:** workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
**Suggestion:** Add RTL/component tests covering the apply, continue, and cancel transitions of the redirect-confirm flow in isolation.

1. **Description:** The modal's 'Continue' button advances with the user's original (unredirected) selection, but its label doesn't communicate this, risking users misreading it as applying the redirect.
**File:** workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceFormRedirectConfirmModal.tsx
**Suggestion:** Relabel the button to make the outcome explicit, e.g. 'Continue with original selection', to disambiguate it from 'Apply Redirect'.

1. **Description:** Cypress specs repeat the clickNext() plus redirectConfirmModal.clickContinue() sequence across many files instead of a shared helper, increasing coupling to the modal's exact interaction pattern.
**File:** workspaces/frontend/src/__tests__/cypress/cypress/pages/workspaces/workspaceForm.ts
**Suggestion:** Add a shared page-object helper, e.g. advancePastRedirectModal(), that wraps the clickNext + modal-continue sequence for reuse across specs.

thaorell:
@caponetto I agree with most the comments, but I have some thoughts for the following. 
6. We haven't historically been adding unit tests for components. While I agree with stronger guardrails with more tests, I think Cypress E2E tests should be sufficient. Please let me know your thoughts on this
7. I want to avoid buttons with wordy labels, but I can add texts to explain what Continue means

caponetto:
@thaorell sounds good to me!

thaorell:
@caponetto I've revised the code. please re-review

caponetto:
### Findings

1. **Description:** Self-redirect cycles render the modal as a hidden-option warning instead of a redirect misconfiguration, showing wrong copy and buttons.
   **File:** `workspaces/frontend/src/app/pages/Workspaces/Form/utils/resolveRedirectChain.ts`
   **Evidence:**
   ```
   if (visited.has(targetId)) {
     cycleDetected = true;
     break;
   }
   ...
   if (chain.length === 0) {
     return { chain, finalTarget: undefined };
   }
   ...
   const hasRedirect = redirectChain && redirectChain.length > 0;
   ```
   **Suggestion:** Add a cycleDetected flag to the return value so the modal can render a dedicated misconfiguration message instead of silently falling back to the hidden-option branch.

-- Improvement suggestions / not blockers --

2. **Description:** The hook's OptionValue abstraction requires manual type casts at call sites, so the type model does not fully encode its own branching logic.
   **File:** `workspaces/frontend/src/app/hooks/useRedirectConfirmation.ts`
   **Evidence:**
   ```
   if (redirectInfo.optionType === 'image') {
     onImageSelect(redirectInfo.finalTarget as OptionsImageConfigValue);
   } else {
     onPodConfigSelect(redirectInfo.finalTarget as OptionsPodConfigValue);
   }
   ```
   **Suggestion:** Model the return type as a discriminated union so TypeScript narrows finalTarget's type based on optionType without requiring explicit `as` casts.

3. **Description:** Redirect severity presentation is spread across three parallel switch-based helpers, requiring manual sync to keep icon, color, and text consistent.
   **File:** `workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceFormRedirectConfirmModal.tsx`
   **Evidence:**
   ```
   const getLevelIcon = (level?: string) => {
     switch (level) {
   ...
   const getLevelColor = (level?: string): 'blue' | 'orange' | 'red' => {
     switch (level) {
   ...
   const getLevelText = (level?: string): string => {
     switch (level) {
   ```
   **Suggestion:** Consolidate the icon/color/text mappings into a single lookup table keyed by severity level to prevent drift.

thaorell:
@caponetto I applied your suggestions

paulovmr (approved):
/lgtm
/approve

PR #1260, opened 2026-07-17
fix: default Workspace state to Unknown in WorkspaceTable

thaorell:
closes: #1104

default Workspace state to Unknown in WorkspaceTable after create/edit action

Screenshot 2026-07-17 at 10 34 51 AM

### DIFF
--- a/workspaces/frontend/src/app/components/WorkspaceTable.tsx
+++ b/workspaces/frontend/src/app/components/WorkspaceTable.tsx
@@ -365,152 +365,162 @@ const WorkspaceTable = React.forwardRef<WorkspaceTableRef, WorkspaceTableProps>(
               {sortedWorkspaces.length > 0 &&
                 sortedWorkspaces
                   .slice(perPage * (page - 1), perPage * page)
-                  .map((workspace, rowIndex) => (
-                    <Tbody id="workspaces-table-content" key={rowIndex} data-testid="table-body">
-                      <Tr
-                        id={`workspaces-table-row-${rowIndex + 1}`}
-                        data-testid={`workspace-row-${rowIndex}`}
-                        isStriped={rowIndex % 2 === 0}
-                      >
-                        {visibleColumnKeys.map((columnKey) => {
-                          if (columnKey === 'connect') {
+                  .map((workspace, rowIndex) => {
+                    const displayState =
+                      workspace.state.length > 0
+                        ? workspace.state
+                        : V1Beta1WorkspaceState.WorkspaceStateUnknown;
+                    return (
+                      <Tbody id="workspaces-table-content" key={rowIndex} data-testid="table-body">
+                        <Tr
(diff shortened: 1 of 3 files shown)


workspaces/frontend/src/app/components/WorkspaceTable.tsx
```diff
@@ -474,9 +474,15 @@ const WorkspaceTable = React.forwardRef<WorkspaceTableRef, WorkspaceTableProps>(
...
-                                      {workspace.state}
+                                  <Tooltip content={workspace.stateMessage || 'Unknown'}>
+                                    <Label
+                                      color={extractWorkspaceStateColor(
+                                        workspace.state.length > 0
+                                          ? workspace.state
+                                          : V1Beta1WorkspaceState.WorkspaceStateUnknown,
+                                      )}
+                                    >
+                                      {workspace.state.length > 0 ? workspace.state : 'Unknown'}
```
caponetto: Can we use `V1Beta1WorkspaceState.WorkspaceStateUnknown` instead of hardcoding `'Unknown'` twice?

caponetto:
Some findings to consider:

1. The tooltip fallback (workspace.stateMessage || 'Unknown') fires whenever stateMessage is empty, even if workspace.state holds a real, non-empty state. This means a row with a legitimate state but no status message shows a contradictory 'Unknown' tooltip alongside its real state label, giving users misleading status feedback.

Suggestion: Only fall back the tooltip text to 'Unknown' when workspace.state is also empty; otherwise show the real state (or an empty/omitted tooltip) when stateMessage is unset for a known state.


2. No test exercises a WorkspaceTable row with an empty workspace.state, so the new 'Unknown' fallback — the only new behavior introduced by this bug fix — can regress silently. The PR's diffstat confirms no test files were added or modified.

Suggestion: Add an RTL/Jest test asserting that a row with workspace.state === '' renders the text 'Unknown' with the grey WorkspaceStateUnknown-mapped Label color.


3. The cell renderer checks workspace.state.length > 0 inline in two separate places (color lookup and label text) to detect the unset-state sentinel, coupling the view directly to the API's empty-string representation instead of using a single normalized/derived state value.

Suggestion: Compute a single normalized state (e.g., const displayState = workspace.state || V1Beta1WorkspaceState.WorkspaceStateUnknown) once and reuse it for both the color lookup and the label text.

caponetto:
One last thing based on the new changes:

**Description:** State normalization is applied only in the cell renderer, so table filters and sorts still use the raw empty-string value, causing newly displayed 'Unknown' rows to be excluded or sorted incorrectly.
   **File:** `workspaces/frontend/src/app/components/WorkspaceTable.tsx`
   **Suggestion:** Compute the normalized displayState once via a shared helper and use it consistently for filtering, sorting, and rendering.

paulovmr (approved):
/lgtm
/approve

PR #1265, opened 2026-07-21
chore: Update list of authorized users for slash commands and planning labels

christian-heusel:
## `chore: Allowlist feature owners to use planning labels`

Currently some of the feature owners are not able to use the planning labels to complete the currently underspecified part of their assigned features, which creates a artificial dependency on the projects maintainers and core team.

The list of users to be added was obtained like this and verified manually:

    $ gh issue list --label "kind/plan-feature" --json assignees | jq -r ".[].assignees[].login" | sort --uniq

Not all of the feature owners are Kubeflow Organization members yet, however they are already contributing very nicely and are considered trusted both by their employer and the wider K8S ecosystem, so this is most likely more a matter of time until we get their PRs merged.

## `chore: Remove inactive members from slash commands`

It does not really make sense to grant this capability to people who are currently not involved in the project and hence not aware of our planning process and methodology.

---

cc @Jefftree @richabanker @siyuanfoundation

### DIFF
--- a/.github/workflows/slash-commands.yaml
+++ b/.github/workflows/slash-commands.yaml
@@ -11,7 +11,7 @@ jobs:
   handle-slash-command:
     if: |
       github.event.issue.pull_request == null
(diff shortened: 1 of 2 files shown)


andyatmiami (approved):
/lgtm
/approve

Issue #887, opened 2026-02-02
[TASK] Implement Logs Viewing API endpoint


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

Following the agreement on architecture/implementation details in #886, we need to deliver a full,         
production-ready implementation of the Logs Viewing API.                                                   
																										 
At a high level this involves:                                                                             
																										 
- Defining the API contract (routes, request/response models) in                                           
https://github.com/kubeflow/notebooks/tree/main/workspaces/backend/api with appropriate Swagger annotations
- Implementing business logic within                                                                       
https://github.com/kubeflow/notebooks/tree/main/workspaces/backend/internal/repositories to interact with  
Kubernetes API server for pod log retrieval as defined in the agreed architecture                          
- Extending client infrastructure to support Kubernetes pod logs subresource API (current                  
controller-runtime client doesn't natively support log streaming)                                          
- Implementing handler logic to properly:                                                                  
    - Validate input data (namespace, workspace name, container name)                                        
    - Handle response conditions (success, pod not found, container not found, logs unavailable, etc.)       
    - Enforce authorization for log access                                                                   
    - Implementing query parameter support for:                                                                
        - Container selection                                                                                    
        - Line limits (tail)                                                                                     
        - Previous container instance logs                                                                       
        - Timestamp inclusion                                                                                    
- Adding unit tests covering:                                                                              
        - Common successful scenarios                                                                            
        - Error handling and validation cases                                                                    
        - Authorization enforcement     

### Acceptance Criteria

- [ ] Full implementation of the Logs Viewing API logic as defined in the agreed architecture of #886          
- [ ] Container logs can be retrieved for a Workspace's pod via API                                            
- [ ] Specific containers can be selected via query parameter                                                  
- [ ] Log output can be limited via tail parameter                                                             
- [ ] Previous container instance logs can be retrieved                                                        
- [ ] API endpoints enforce appropriate authorization                                                          
- [ ] API endpoints return clear error responses for unavailable logs (pod not running, container not found,   
etc.)                                                                                                                                              
- [ ] Swagger/OpenAPI documentation is generated via make swag

YoyinZyc:
cc @yoyinzyc

YoyinZyc:
@andyatmiami as the pr was merged, are we good to close the issue?

christian-heusel:
[...]

Issue #886, opened 2026-02-02
[TASK] Agree on architecture/implementation details for logs viewing API


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

As a community, we need to agree on how the Logs Viewing API will be designed and implemented. This API    
will enable users to retrieve container logs for their Workspaces through the frontend.                 
																										 
The Notebooks 1.0 implementation provides a proven pattern that should serve as inspiration for the v2     
design. Relevant code references from the notebooks-v1 branch:                                             
																										 
Backend API Layer:                                                                                         
- https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/crud-web-apps/common/backend/kubeflow/
kubeflow/crud_backend/api/pod.py                                                                           
- get_pod_logs(namespace, pod, container, auth=True) — Retrieves logs via Kubernetes                     
read_namespaced_pod_log() with authorization                                                               
																										 
Route Definition:                                                                                          
- https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/crud-web-apps/jupyter/backend/apps/com
mon/routes/get.py                                                                                          
- Endpoint pattern: /api/namespaces/<namespace>/notebooks/<notebook_name>/pod/<pod_name>/logs            
																										 
Key areas requiring alignment for v2:                                                                      
- API endpoint structure and path conventions consistent with existing v2 patterns                         
- Query parameters for container selection, line limits, previous container logs, timestamps               
- Response format (JSON envelope vs plain text)                                                            
- Authorization model (reuse workspace GET permissions or separate log-specific policy)                    
- Error response conventions for various failure scenarios                                                 
																										 
An initial proposal should be drafted and reviewed so that expectations for the Logs Viewing API are       
well-defined and able to be used as a source of truth for implementation.  

### Acceptance Criteria

- [ ] API Proposal for Logs Viewing API is drafted as a Markdown document                                      
- [ ] API Proposal is reviewed and agreed upon by community                                                    
- [ ] API Proposal is architecturally consistent with existing v2 backend patterns

YoyinZyc:
cc @yoyinzyc

andyatmiami:
Proposal available here:
- https://gist.github.com/andyatmiami/ed4173457d97138bc8bb55f7e8a119fc

andyatmiami:
# API Proposal: Workspace Logs Batch API

[...]

PR #1267, opened 2026-07-21
feat: implement endpoint to support fetching workspace pod log

YoyinZyc:
Implement the batch log API proposed in https://github.com/kubeflow/notebooks/issues/886#issuecomment-5072354043
closes: https://github.com/kubeflow/notebooks/issues/887
related: https://github.com/kubeflow/notebooks/issues/886 

### Behavior
A new one-shot endpoint returning a workspace pods container(including the initContainer like istio) logs as a plain text with timestamps. 
 
Query params: container(defaults to first/primary container); tail(default to 1000); previous (default to false); sinceTime(default to none)

#### API
```
GET /api/v1/workspaces/{namespace}/{name}/podtemplate/logs/batch
      ?container=<name>    # optional, defaults to the primary container ("main")
      &tail=<int>          # optional, default 1000
      &previous=<bool>     # optional, default false (previous terminated instance)
      &sinceTime=<string> #optional, default to none
```
#### Response

```
2026-07-23T18:37:19.502427795Z 2026-07-23T18:37:19.502162Z	info	xdsproxy	connected to delta upstream XDS server: istiod.istio-system.svc:15012	id=96
2026-07-23T19:07:20.316168674Z 2026-07-23T19:07:20.315975Z	info	xdsproxy	connected to delta upstream XDS server: istiod.istio-system.svc:15012	id=97
2026-07-23T19:36:29.611434886Z 2026-07-23T19:36:29.611255Z	info	xdsproxy	connected to delta upstream XDS server: istiod.istio-system.svc:15012	id=98
2026-07-23T20:04:05.870345556Z 2026-07-23T20:04:05.870108Z	info	xdsproxy	connected to delta upstream XDS server: istiod.istio-system.svc:15012	id=99
2026-07-23T20:32:49.598960074Z 2026-07-23T20:32:49.598795Z	info	xdsproxy	connected to delta upstream XDS server: istiod.istio-system.svc:15012	id=100
2026-07-23T21:01:53.168136917Z 2026-07-23T21:01:53.167896Z	info	xdsproxy	connected to delta upstream XDS server: istiod.istio-sy
```
### Unit tests
`internal/repositories/logs/repo_test.go`
`workspaces/backend/api/workspace_podtemplate_logs_handler_test.go`

### Tested Manually
BASE_URL=https://localhost:8443/workspaces/api/v1/workspaces/default/jupyterlab-workspace/podtemplate/logs/batch

#### Positive
| Path  | HTTP Code|Result |
| ------------- | ------------- | --------------|
|  empty | 200 | default the main container with tail set to 1000 and previous=false  |
| tail=15  | 200 | 15 line of log  |
| container=main| 200 | return log for main container|
| container=istio-proxy | 200 | return log for istio proxy |
| sinceTime=2026-07-23T18:30:00Z| 200 | return main log since that time |
| container=istio-proxy&sinceTime=2026-07-23T18:30:00Z | 200 | return istio proxy log since that time |
| previous=true | 200 | return the log of previous terminated container|

#### Negative
| Path  | HTTP Code | Error Msg |
| ------------- | ------------- | --------------|
| previous=true(never restarted) | 400 | no logs found for the previous container instance |
| container=<invalid>| 400 | container not found in pod |
| tail=<invalid> / tail=0 / previous=maybe / sinceTime=<invalid time> / container=BAD_NAME! | 422 | field validation error |
| non-existent workspace| 404 | workspace not found |
| non-running container(in waiting state) | 400 | container has not started yet (note: this one is hard to test manually, mostly rely on unit test) |

### DIFF
--- a/workspaces/backend/api/constants/paths.go
+++ b/workspaces/backend/api/constants/paths.go
@@ -24,13 +24,13 @@ const (
 	HealthCheckPath = PathPrefix + "/healthcheck"
 
 	// workspaces
-	AllWorkspacesPath               = PathPrefix + "/workspaces"
-	WorkspacesByNamespacePath       = AllWorkspacesPath + "/:" + NamespacePathParam
-	WorkspacesByNamePath            = AllWorkspacesPath + "/:" + NamespacePathParam + "/:" + ResourceNamePathParam
-	WorkspaceActionsPath            = WorkspacesByNamePath + "/actions"
-	PauseWorkspacePath              = WorkspaceActionsPath + "/pause"
-	WorkspacePodTemplateDetailsPath = WorkspacesByNamePath + "/podtemplate/details"
-	WorkspacePodLogsBatchPath       = WorkspacesByNamePath + "/podtemplate/logs/batch"
+	AllWorkspacesPath                    = PathPrefix + "/workspaces"
+	WorkspacesByNamespacePath            = AllWorkspacesPath + "/:" + NamespacePathParam
+	WorkspacesByNamePath                 = AllWorkspacesPath + "/:" + NamespacePathParam + "/:" + ResourceNamePathParam
+	WorkspaceActionsPath                 = WorkspacesByNamePath + "/actions"
+	PauseWorkspacePath                   = WorkspaceActionsPath + "/pause"
+	WorkspacePodTemplateDetailsPath      = WorkspacesByNamePath + "/podtemplate/details"
+	WorkspacePodTemplatePodLogsBatchPath = WorkspacesByNamePath + "/podtemplate/logs/batch"
 
 	// workspacekinds
 	AllWorkspaceKindsPath            = PathPrefix + "/workspacekinds"
--- a/workspaces/backend/api/constants/paths.go
+++ b/workspaces/backend/api/constants/paths.go
@@ -30,6 +30,7 @@ const (
 	WorkspaceActionsPath            = WorkspacesByNamePath + "/actions"
 	PauseWorkspacePath              = WorkspaceActionsPath + "/pause"
 	WorkspacePodTemplateDetailsPath = WorkspacesByNamePath + "/podtemplate/details"
+	WorkspacePodLogsBatchPath       = WorkspacesByNamePath + "/podtemplate/logs/batch"
 
 	// workspacekinds
 	AllWorkspaceKindsPath            = PathPrefix + "/workspacekinds"
--- a/workspaces/backend/api/response_errors.go
+++ b/workspaces/backend/api/response_errors.go
@@ -185,18 +185,6 @@ func (a *App) notFoundResponse(w http.ResponseWriter, r *http.Request) {
 	a.errorResponse(w, r, httpError)
 }
 
-// HTTP: 404 with a caller-provided message.
-func (a *App) notFoundResponseWithMessage(w http.ResponseWriter, r *http.Request, err error) {
-	httpError := &HTTPError{
-		StatusCode: http.StatusNotFound,
-		ErrorResponse: ErrorResponse{
-			Code:    strconv.Itoa(http.StatusNotFound),
-			Message: err.Error(),
-		},
-	}
-	a.errorResponse(w, r, httpError)
-}
-
 // HTTP: 405
 func (a *App) methodNotAllowedResponse(w http.ResponseWriter, r *http.Request) {
 	httpError := &HTTPError{
--- a/workspaces/backend/api/response_errors.go
+++ b/workspaces/backend/api/response_errors.go
@@ -185,6 +185,18 @@ func (a *App) notFoundResponse(w http.ResponseWriter, r *http.Request) {
 	a.errorResponse(w, r, httpError)
 }
 
+// HTTP: 404 with a caller-provided message.
+func (a *App) notFoundResponseWithMessage(w http.ResponseWriter, r *http.Request, err error) {
+	httpError := &HTTPError{
+		StatusCode: http.StatusNotFound,
+		ErrorResponse: ErrorResponse{
+			Code:    strconv.Itoa(http.StatusNotFound),
+			Message: err.Error(),
+		},
+	}
+	a.errorResponse(w, r, httpError)
+}
+
 // HTTP: 405
 func (a *App) methodNotAllowedResponse(w http.ResponseWriter, r *http.Request) {
 	httpError := &HTTPError{
--- a/workspaces/backend/api/workspace_logs_handler.go
+++ b/workspaces/backend/api/workspace_logs_handler.go
@@ -1,0 +1,144 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+import (
+	"errors"
+	"net/http"
+	"strconv"
+
+	"github.com/julienschmidt/httprouter"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	"github.com/kubeflow/notebooks/workspaces/backend/api/constants"
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/auth"
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/helper"
+	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/workspaces/podtemplate/logs"
+	repository "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/logs"
+)
+
+const (
+	logsContainerQueryParam = "container"
+	logsTailQueryParam      = "tail"
+	logsPreviousQueryParam  = "previous"
+)
+
+// WorkspaceLogsEnvelope is the response envelope for workspace container logs.
+type WorkspaceLogsEnvelope Envelope[models.WorkspaceLogs]
+
+// GetWorkspaceLogsHandler returns a point-in-time snapshot of container logs for a workspace pod.
+//
+//	@Summary		Get workspace container logs (batch)
+//	@Description	Returns a point-in-time snapshot of container logs for the workspace pod as a JSON array of log lines.
+//	@Tags			workspaces
+//	@ID				getWorkspaceLogsBatch
+//	@Produce		json
+//	@Param			namespace	path		string					true	"Namespace of the workspace"	extensions(x-example=kubeflow-user-example-com)
+//	@Param			name		path		string					true	"Name of the workspace"			extensions(x-example=my-workspace)
+//	@Param			container	query		string					false	"Target container name. Defaults to the first (primary) container."
+//	@Param			tail		query		integer					false	"Number of lines from the end of the log to return. Defaults to 1000."
+//	@Param			previous	query		boolean					false	"If true, returns logs from the previous terminated container instance."
+//	@Success		200			{object}	WorkspaceLogsEnvelope	"Successful operation."
+//	@Failure		400			{object}	ErrorEnvelope			"Bad Request. Container not found in workspace pod."
+//	@Failure		401			{object}	ErrorEnvelope			"Unauthorized."
+//	@Failure		403			{object}	ErrorEnvelope			"Forbidden."
+//	@Failure		404			{object}	ErrorEnvelope			"Workspace not found."
+//	@Failure		409			{object}	ErrorEnvelope			"Conflict. Workspace pod is not running."
+//	@Failure		422			{object}	ErrorEnvelope			"Unprocessable Entity. Validation error."
+//	@Failure		500			{object}	ErrorEnvelope			"Internal server error."
+//	@Router			/workspaces/{namespace}/{name}/podtemplate/logs/batch [get]
+func (a *App) GetWorkspaceLogsHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	namespace := ps.ByName(constants.NamespacePathParam)
+	workspaceName := ps.ByName(constants.ResourceNamePathParam)
+
+	// validate path parameters
+	var valErrs field.ErrorList //nolint:prealloc
+	valErrs = append(valErrs, helper.ValidateKubernetesNamespaceName(field.NewPath(constants.NamespacePathParam), namespace)...)
+	valErrs = append(valErrs, helper.ValidateWorkspaceName(field.NewPath(constants.ResourceNamePathParam), workspaceName)...)
+	if len(valErrs) > 0 {
+		a.failedValidationResponse(w, r, errMsgPathParamsInvalid, valErrs, nil)
+		return
+	}
+
+	// parse and validate query parameters
+	opts, valErrs := parseLogOptions(r)
+	if len(valErrs) > 0 {
+		a.failedValidationResponse(w, r, errMsgQueryParamsInvalid, valErrs, nil)
+		return
+	}
+
+	// =========================== AUTH ===========================
+	authPolicies := []*auth.ResourcePolicy{
+		auth.NewResourcePolicy(auth.VerbGet, auth.Workspaces, auth.ResourcePolicyResourceMeta{Namespace: namespace, Name: workspaceName}),
+	}
+	if _, ok := a.requireAuth(w, r, authPolicies); !ok {
+		return
+	}
+	// ============================================================
+
+	logs, err := a.repositories.Logs.GetWorkspaceLogs(r.Context(), namespace, workspaceName, opts)
+	if err != nil {
+		switch {
+		case errors.Is(err, repository.ErrWorkspaceNotFound):
+			a.notFoundResponse(w, r)
+		case errors.Is(err, repository.ErrPreviousLogsNotFound):
+			a.notFoundResponse(w, r)
+		case errors.Is(err, repository.ErrPodNotRunning):
+			a.conflictResponse(w, r, err, nil)
+		case errors.Is(err, repository.ErrContainerNotFound):
+			a.badRequestResponse(w, r, err)
+		case errors.Is(err, repository.ErrContainerNotRunning):
+			a.conflictResponse(w, r, err, nil)
+		default:
+			a.serverErrorResponse(w, r, err)
+		}
+		return
+	}
+
+	responseEnvelope := &WorkspaceLogsEnvelope{Data: logs}
+	a.dataResponse(w, r, responseEnvelope)
+}
+
+// parseLogOptions parses and validates the log-related query parameters.
+func parseLogOptions(r *http.Request) (*models.LogOptions, field.ErrorList) {
+	var valErrs field.ErrorList
+	query := r.URL.Query()
+
+	opts := &models.LogOptions{
+		Container: query.Get(logsContainerQueryParam),
+	}
+
+	if raw := query.Get(logsTailQueryParam); raw != "" {
+		tail, err := strconv.ParseInt(raw, 10, 64)
+		if err != nil || tail <= 0 {
+			valErrs = append(valErrs, field.Invalid(field.NewPath(logsTailQueryParam), raw, "must be a positive integer"))
+		} else {
+			opts.TailLines = tail
+		}
+	}
+
+	if raw := query.Get(logsPreviousQueryParam); raw != "" {
+		previous, err := strconv.ParseBool(raw)
+		if err != nil {
+			valErrs = append(valErrs, field.Invalid(field.NewPath(logsPreviousQueryParam), raw, "must be a boolean"))
+		} else {
+			opts.Previous = previous
+		}
+	}
+
+	return opts, valErrs
+}
--- a/workspaces/backend/api/workspace_logs_handler.go
+++ b/workspaces/backend/api/workspace_logs_handler.go
@@ -18,10 +18,13 @@ package api
 
 import (
 	"errors"
+	"io"
 	"net/http"
 	"strconv"
+	"time"
 
 	"github.com/julienschmidt/httprouter"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/kubeflow/notebooks/workspaces/backend/api/constants"
@@ -33,33 +36,32 @@ import (
 
 const (
 	logsContainerQueryParam = "container"
-	logsTailQueryParam      = "tail"
+	logsTailLinesQueryParam = "tailLines"
 	logsPreviousQueryParam  = "previous"
+	logSinceTimeQueryParam  = "sinceTime"
 )
-
-// WorkspaceLogsEnvelope is the response envelope for workspace container logs.
-type WorkspaceLogsEnvelope Envelope[models.WorkspaceLogs]
 
 // GetWorkspaceLogsHandler returns a point-in-time snapshot of container logs for a workspace pod.
 //
 //	@Summary		Get workspace container logs (batch)
-//	@Description	Returns a point-in-time snapshot of container logs for the workspace pod as a JSON array of log lines.
+//	@Description	Returns a point-in-time snapshot of container logs for the workspace pod as a raw text/plain stream proxied directly from the Kubernetes pod logs API.
 //	@Tags			workspaces
 //	@ID				getWorkspaceLogsBatch
-//	@Produce		json
-//	@Param			namespace	path		string					true	"Namespace of the workspace"	extensions(x-example=kubeflow-user-example-com)
-//	@Param			name		path		string					true	"Name of the workspace"			extensions(x-example=my-workspace)
-//	@Param			container	query		string					false	"Target container name. Defaults to the first (primary) container."
-//	@Param			tail		query		integer					false	"Number of lines from the end of the log to return. Defaults to 1000."
-//	@Param			previous	query		boolean					false	"If true, returns logs from the previous terminated container instance."
-//	@Success		200			{object}	WorkspaceLogsEnvelope	"Successful operation."
-//	@Failure		400			{object}	ErrorEnvelope			"Bad Request. Container not found in workspace pod."
-//	@Failure		401			{object}	ErrorEnvelope			"Unauthorized."
-//	@Failure		403			{object}	ErrorEnvelope			"Forbidden."
-//	@Failure		404			{object}	ErrorEnvelope			"Workspace not found."
-//	@Failure		409			{object}	ErrorEnvelope			"Conflict. Workspace pod is not running."
-//	@Failure		422			{object}	ErrorEnvelope			"Unprocessable Entity. Validation error."
-//	@Failure		500			{object}	ErrorEnvelope			"Internal server error."
+//	@Produce		plain
+//	@Param			namespace	path		string			true	"Namespace of the workspace"	extensions(x-example=kubeflow-user-example-com)
+//	@Param			name		path		string			true	"Name of the workspace"			extensions(x-example=my-workspace)
+//	@Param			container	query		string			false	"Target container name. Defaults to the first (primary) container."
+//	@Param			tailLines	query		integer			false	"Number of lines from the end of the log to return. Defaults to 1000."
+//	@Param			sinceTime	query		string			false	"Only return logs after this RFC3339 timestamp (e.g. 2026-07-15T10:30:00Z)."
+//	@Param			previous	query		boolean			false	"If true, returns logs from the previous terminated container instance."
+//	@Success		200			{string}	string			"Raw container log stream (text/plain)."
+//	@Failure		400			{object}	ErrorEnvelope	"Bad Request. Container not found in workspace pod."
+//	@Failure		401			{object}	ErrorEnvelope	"Unauthorized."
+//	@Failure		403			{object}	ErrorEnvelope	"Forbidden."
+//	@Failure		404			{object}	ErrorEnvelope	"Workspace not found."
+//	@Failure		409			{object}	ErrorEnvelope	"Conflict. Workspace pod is not running."
+//	@Failure		422			{object}	ErrorEnvelope	"Unprocessable Entity. Validation error."
+//	@Failure		500			{object}	ErrorEnvelope	"Internal server error."
 //	@Router			/workspaces/{namespace}/{name}/podtemplate/logs/batch [get]
 func (a *App) GetWorkspaceLogsHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	namespace := ps.ByName(constants.NamespacePathParam)
@@ -90,7 +92,7 @@ func (a *App) GetWorkspaceLogsHandler(w http.ResponseWriter, r *http.Request, ps
 	}
 	// ============================================================
 
-	logs, err := a.repositories.Logs.GetWorkspaceLogs(r.Context(), namespace, workspaceName, opts)
+	stream, err := a.repositories.Logs.OpenLogStream(r.Context(), namespace, workspaceName, opts)
 	if err != nil {
 		switch {
 		case errors.Is(err, repository.ErrWorkspaceNotFound):
@@ -108,9 +110,16 @@ func (a *App) GetWorkspaceLogsHandler(w http.ResponseWriter, r *http.Request, ps
 		}
 		return
 	}
+	defer func() { _ = stream.Close() }()
 
-	responseEnvelope := &WorkspaceLogsEnvelope{Data: logs}
-	a.dataResponse(w, r, responseEnvelope)
+	// Success responses are always a raw text/plain stream.
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(http.StatusOK)
+
+	// Proxy the log stream directly to the client.
+	if _, err := io.Copy(w, stream); err != nil {
+		a.logger.Error("error while streaming workspace logs", "error", err, "namespace", namespace, "workspace", workspaceName)
+	}
 }
 
 // parseLogOptions parses and validates the log-related query parameters.
@@ -122,10 +131,10 @@ func parseLogOptions(r *http.Request) (*models.LogOptions, field.ErrorList) {
 		Container: query.Get(logsContainerQueryParam),
 	}
 
-	if raw := query.Get(logsTailQueryParam); raw != "" {
+	if raw := query.Get(logsTailLinesQueryParam); raw != "" {
 		tail, err := strconv.ParseInt(raw, 10, 64)
 		if err != nil || tail <= 0 {
-			valErrs = append(valErrs, field.Invalid(field.NewPath(logsTailQueryParam), raw, "must be a positive integer"))
+			valErrs = append(valErrs, field.Invalid(field.NewPath(logsTailLinesQueryParam), raw, "must be a positive integer"))
 		} else {
 			opts.TailLines = tail
 		}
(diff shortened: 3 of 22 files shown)


christian-heusel:
@YoyinZyc FYI I have added a compression middleware to the backend in parallel which should help quite a bit for the log endpoint 🤗  https://github.com/kubeflow/notebooks/pull/1269

workspaces/backend/api/response_errors.go
```diff
@@ -185,6 +185,18 @@ func (a *App) notFoundResponse(w http.ResponseWriter, r *http.Request) {
...
+	httpError := &HTTPError{
+		StatusCode: http.StatusNotFound,
+		ErrorResponse: ErrorResponse{
+			Code:    strconv.Itoa(http.StatusNotFound),
+			Message: err.Error(),
+		},
+	}
+	a.errorResponse(w, r, httpError)
+}
+
```
andyatmiami: Open to discussion - but I'm not sure this function is really warranted.. all existing endpoints use the existing `a.notFoundResponse(w, r)` and _I think_ that works well enough as is..

Would prefer to keep with convention here - but let me know if I am overlooking something and/or why you think we should add this...

workspaces/backend/api/workspace_logs_handler.go
```diff
@@ -0,0 +1,163 @@
...
+	if _, ok := a.requireAuth(w, r, authPolicies); !ok {
+		return
+	}
+	// ============================================================
+
+	stream, err := a.repositories.Logs.OpenLogStream(r.Context(), namespace, workspaceName, opts)
+	if err != nil {
+		switch {
+		case errors.Is(err, repository.ErrWorkspaceNotFound):
+			a.notFoundResponseWithMessage(w, r, err)
```
andyatmiami: Related to: https://github.com/kubeflow/notebooks/pull/1267/changes#r3695881990

I think we should just stick with `a.notFoundResponse(w, r)` to "keep it simple"

workspaces/backend/internal/repositories/logs/repo.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/config"
+
+	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/workspaces/podtemplate/logs"
+)
+
+var (
+	ErrWorkspaceNotFound    = fmt.Errorf("workspace not found")
+	ErrPodNotRunning        = fmt.Errorf("workspace pod is not running")
+	ErrContainerNotFound    = fmt.Errorf("container not found in pod")
```
andyatmiami:
```suggestion
	ErrContainerNotFound    = fmt.Errorf("container not found in workspace pod")
```

workspaces/backend/api/workspace_logs_handler.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/helper"
+	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/workspaces/podtemplate/logs"
+	repository "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/logs"
+)
+
+const (
+	logsContainerQueryParam = "container"
+	logsTailLinesQueryParam = "tailLines"
+	logsPreviousQueryParam  = "previous"
+	logSinceTimeQueryParam  = "sinceTime"
```
andyatmiami:
```suggestion
	logsSinceTimeQueryParam  = "sinceTime"
```

workspaces/backend/api/workspace_podtemplate_logs_handler_test.go
```diff
@@ -0,0 +1,199 @@
...
+			By("executing GetWorkspaceLogsHandler")
+			rr := httptest.NewRecorder()
+			a.GetWorkspaceLogsHandler(rr, req, ps)
+			rs := rr.Result()
+			defer rs.Body.Close()
+
+			By("verifying status is 409 Conflict")
+			Expect(rs.StatusCode).To(Equal(http.StatusConflict))
+		})
+	})
```
andyatmiami: The 409 test ("workspace pod is not running") only checks the status code:

 ` Expect(rs.StatusCode).To(Equal(http.StatusConflict))`

But the 404 test in this same file goes further — it parses the ErrorEnvelope and asserts the error message matches the sentinel. The 409 test should  do the same for consistency within the file as well as generally being much more reliable/robust/accurate.

workspaces/backend/api/workspace_podtemplate_logs_handler.go
andyatmiami, line 1: Lets call this `workspace_podtemplate_logs_handler.go` to be consistent with other `*handler` files (that have a `/podtemplate` segment in their URL

workspaces/backend/api/workspace_podtemplate_logs_handler_test.go
andyatmiami, line 1: Lets call this `workspace_podtemplate_logs_handler_test.go` to be consistent with other `*handler` files (that have a `/podtemplate` segment in their URL

workspaces/backend/api/constants/paths.go
```diff
@@ -30,6 +30,7 @@ const (
 	WorkspaceActionsPath            = WorkspacesByNamePath + "/actions"
 	PauseWorkspacePath              = WorkspaceActionsPath + "/pause"
 	WorkspacePodTemplateDetailsPath = WorkspacesByNamePath + "/podtemplate/details"
+	WorkspacePodLogsBatchPath       = WorkspacesByNamePath + "/podtemplate/logs/batch"
```
andyatmiami:
```suggestion
	WorkspacePodTemplatePodLogsBatchPath       = WorkspacesByNamePath + "/podtemplate/logs/batch"
```

workspaces/backend/api/workspace_logs_handler.go
```diff
@@ -0,0 +1,163 @@
...
+//	@Success		200			{string}	string			"Raw container log stream (text/plain)."
+//	@Failure		400			{object}	ErrorEnvelope	"Bad Request. Container not found in workspace pod."
+//	@Failure		401			{object}	ErrorEnvelope	"Unauthorized."
+//	@Failure		403			{object}	ErrorEnvelope	"Forbidden."
+//	@Failure		404			{object}	ErrorEnvelope	"Workspace not found."
+//	@Failure		409			{object}	ErrorEnvelope	"Conflict. Workspace pod is not running."
+//	@Failure		422			{object}	ErrorEnvelope	"Unprocessable Entity. Validation error."
+//	@Failure		500			{object}	ErrorEnvelope	"Internal server error."
+//	@Router			/workspaces/{namespace}/{name}/podtemplate/logs/batch [get]
+func (a *App) GetWorkspaceLogsHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
andyatmiami:
```suggestion
func (a *App) GetWorkspacePodTemplateLogsHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```

workspaces/backend/api/workspace_logs_handler.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+}
+
+// parseLogOptions parses and validates the log-related query parameters.
+func parseLogOptions(r *http.Request) (*models.LogOptions, field.ErrorList) {
+	var valErrs field.ErrorList
+	query := r.URL.Query()
+
+	opts := &models.LogOptions{
+		Container: query.Get(logsContainerQueryParam),
```
andyatmiami: I wonder if we should consider adding `kubectl.kubernetes.io/default-container` to the Workspace pod we reconcile in the controller (?). That being a controller change - its certainly out of scope for **THIS PR** - but I'm curious if we can/should rely on this implicit defaulting here...

We hard-code the name `main` - perhaps we should default to that to ensure reasonable default behavior when container name not provided?  Unless we are sure this default wouldn't ever result in sidecars being chosen...

workspaces/backend/api/workspace_podtemplate_logs_handler.go
```diff
@@ -0,0 +1,163 @@
...
+		default:
+			a.serverErrorResponse(w, r, err)
+		}
+		return
+	}
+	defer func() { _ = stream.Close() }()
+
+	// Success responses are always a raw text/plain stream.
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(http.StatusOK)
```
andyatmiami, line 106: ℹ️ (no change requested - just calling this out for transparency)

If the subsequent io.Copy fails mid-stream, you can't change the status code — so  logging the error (as you're doing) is the correct fallback. The client will see a truncated response, which is the expected behavior for a streaming  endpoint that breaks mid-flight.

workspaces/backend/internal/repositories/podlogs/repo.go
```diff
@@ -0,0 +1,191 @@
...
+	ErrPodNotRunning        = fmt.Errorf("workspace pod is not running")
+	ErrContainerNotFound    = fmt.Errorf("container not found in pod")
+	ErrContainerNotRunning  = fmt.Errorf("container has not started yet")
+	ErrPreviousLogsNotFound = fmt.Errorf("no logs found for the previous container instance")
+)
+
+var (
+	// safeLimitBytes is the maximum number of bytes the Kubernetes API will return
+	// for a single log request, bounding the size of the proxied stream.
+	safeLimitBytes = int64(100 * 1024 * 1024) // 100 MB
```
andyatmiami, line 53: ℹ️ No change requested - just calling this out for future consideration..

Probably / maybe want to expose this as an env var and/or command flag in future.. the "right" value for this likely depends on a combination of amount of memory allocated to the controller as well as expected concurrent load..

But those are all straightforward changes (once behavior agreed upon) that can cleanly be applied on top of this work.

workspaces/backend/internal/repositories/logs/repo.go
```diff
@@ -0,0 +1,191 @@
...
+		Timestamps: true,
+		SinceTime:  opts.SinceTime,
+	})
+
+	stream, err := req.Stream(ctx)
+	if err != nil {
+		// When previous=true but the container has never restarted, the Kubernetes
+		// API returns a 400 with a "previous terminated container ... not found"
+		// message. Surface this as a semantic error instead of a generic 500.
+		if opts.Previous && apierrors.IsBadRequest(err) && strings.Contains(err.Error(), "previous terminated container") {
```
andyatmiami: ℹ️ 

substring parsing on 3rd party error messages always _makes me nervous_ - but probably fine in this instance (worst that can happen if that underlying error string were to ever change would be to misreport a `500`)

andyatmiami: however, it might be worth a non-exported helper function here to handle this scenario...

that gives us a more natural place to "document this" (on the function) - and then makes **this actual inline code** easier to read/ration about as its not a 3-conditional eval

🤔 

Let me know what you think - not requiring a change here - so we can discuss (and you reject!)

workspaces/backend/internal/repositories/logs/repo.go
```diff
@@ -0,0 +1,191 @@
...
+			return nil, ErrPreviousLogsNotFound
+		}
+		return nil, fmt.Errorf("failed to open log stream for pod %s, container %s: %w", podName, containerName, err)
```
andyatmiami: seems like we can at least define the "template string" as a `const` at the top of this file - so its easier to see/locate all our error strings together..

workspaces/backend/internal/repositories/logs/repo.go
```diff
@@ -0,0 +1,191 @@
...
+		// A requested container must exist among the pod's containers.
+		if !containerExists(containerName, podStatus.Containers, podStatus.InitContainers) {
+			return "", "", ErrContainerNotFound
+		}
+	} else {
+		// None requested: default to the primary (first regular) container.
+		if len(podStatus.Containers) == 0 {
+			return "", "", ErrContainerNotRunning
+		}
+		containerName = podStatus.Containers[0].Name
```
andyatmiami: 🤔 I wonder if we capitalize on the fact we hardcode `main` as the container name for our primary container - and in this branch "force" that as the default...

i guess defensively we'd still want to ensure there **IS** a `main` container to be robust against _the unknown_

workspaces/backend/internal/repositories/podlogs/repo.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	return podName, containerName, nil
+}
+
+// ensureContainerStarted checks the live Pod status and returns ErrContainerNotRunning
+// if the target container is still in the Waiting state (i.e. it has not started yet
+// and therefore has no logs available for the current instance).
+func (r *LogsRepository) ensureContainerStarted(ctx context.Context, namespace, podName, containerName string) error {
+	pod, err := r.clientset.CoreV1().Pods(namespace).Get(ctx, podName, metav1.GetOptions{})
```
andyatmiami, line 180: This is probably implied from the function comment:
- `the live Pod status`

... but might be nice to also **explicitly** call out here we are opting to use the `clientset` (vs. `client`) as we are (presumably) trying to make a real-time decision on state and want to avoid a potentially stale cached version

workspaces/backend/internal/repositories/logs/repo.go
```diff
@@ -0,0 +1,191 @@
...
+// if the target container is still in the Waiting state (i.e. it has not started yet
+// and therefore has no logs available for the current instance).
+func (r *LogsRepository) ensureContainerStarted(ctx context.Context, namespace, podName, containerName string) error {
+	pod, err := r.clientset.CoreV1().Pods(namespace).Get(ctx, podName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// The Workspace status references a pod that no longer exists.
+			return ErrPodNotRunning
+		}
+		return fmt.Errorf("failed to get pod %s: %w", podName, err)
```
andyatmiami: seems like we can at least define the "template string" as a `const` at the top of this file - so its easier to see/locate all our error strings together..

workspaces/backend/internal/repositories/logs/repo.go
```diff
@@ -0,0 +1,191 @@
...
+			if cs.Name != containerName {
+				continue
+			}
+			// A container that is still Waiting has never started and has no logs yet.
+			if cs.State.Waiting != nil {
+				return ErrContainerNotRunning
+			}
+			return nil
+		}
+	}
```
andyatmiami: While admittedly not a realistic "performance concern" - I personally like the "more boring" two-loop style iteration present here:
- https://github.com/kubeflow/notebooks/pull/1267/changes#diff-1c06a4d26f6bb77a636e82b1459e5fcb14434fe0b77f183ae33b435977cc1f93R105-R114

its a simple linear scan with no extra allocations...

but at minimum - I think we should be consistent in iteration logic implementation

andyatmiami (changes requested):
@YoyinZyc - Sorry I've been _distracted by life_ - and took me awhile to be able to dig into this PR.

Overall very excited about the implementation here - a lot of "nitpicks" around various things that is completely natural this being your first pR..

But the core implementation is **really well done**

Let me know if you disagree with any of my comments and want to talk about them - always happy to discuss!

workspaces/backend/internal/models/workspaces/podtemplate/logs/types.go
```diff
@@ -0,0 +1,34 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type LogOptions struct {
+	// The name of the container to retrieve logs from.
+	// If omitted, defaults to the first container (the workspace's primary container).
+	Container string
+	// The number of lines to retrieve from the end of the logs.
+	// By default, the value is 1000.
+	TailLines int64
```
christian-heusel, line 29: This does not currently seem to be respected:
```
$ curl -sk -H "Kubeflow-Userid: admin" https://localhost:8443/workspaces/api/v1/workspaces/default/jupyterlab-workspace/podtemplate/logs/batch\?tail=10 | wc -l 
42
```

christian-heusel: Ah nevermind, this was just due to me using the parameter naming:
```
$ curl -sk -H "Kubeflow-Userid: admin" https://localhost:8443/workspaces/api/v1/workspaces/default/jupyterlab-workspace/podtemplate/logs/batch\?tailLines=10 | wc -l 
10
```

Maybe we should guard against wrongly spelt parameter? 🤔 Anyways, that's most likely out of scope for this PR 😅 

Also see https://github.com/kubeflow/notebooks/pull/1267/changes#r3708558872

workspaces/backend/api/workspace_logs_handler.go
christian-heusel: I think those should potentially live in `api/constants/query_params.go`:
https://github.com/kubeflow/notebooks/blob/97159cbd6d1e4a671c5c078d83fa9bd678741813/workspaces/backend/api/constants/query_params.go#L17-L20

workspaces/backend/api/workspace_podtemplate_logs_handler.go
```diff
@@ -0,0 +1,163 @@
...
+		if err != nil {
+			valErrs = append(valErrs, field.Invalid(field.NewPath(logSinceTimeQueryParam), raw, "must be a valid RFC3339 timestamp"))
+		} else {
+			sinceTime := metav1.NewTime(t)
+			opts.SinceTime = &sinceTime
+		}
+	}
+
+	return opts, valErrs
+}
```
christian-heusel, line 158: I think to more closely follow the codebase conventions we could potentially introduce 3 new helper functions and add them to `internal/helper/validation.go`:
```go
func ValidateFieldIsPositiveInt64(path *field.Path, value string) (int64, field.ErrorList)
func ValidateFieldIsBool(path *field.Path, value string) (bool, field.ErrorList)
func ValidateFieldIsRFC3339Time(path *field.Path, value string) (metav1.Time, field.ErrorList)
```

Also we could validate the container name:
```go
if raw := query.Get(logsContainerQueryParam); raw != "" { 
  valErrs = append(valErrs, helper.ValidateFieldIsDNS1123Label(field.NewPath(logsContainerQueryParam), raw)...)
  opts.Container = raw
}
```

workspaces/backend/api/workspace_podtemplate_logs_handler.go
```diff
@@ -0,0 +1,163 @@
...
+
+	stream, err := a.repositories.Logs.OpenLogStream(r.Context(), namespace, workspaceName, opts)
+	if err != nil {
+		switch {
+		case errors.Is(err, repository.ErrWorkspaceNotFound):
+			a.notFoundResponseWithMessage(w, r, err)
+		case errors.Is(err, repository.ErrPreviousLogsNotFound):
+			a.conflictResponse(w, r, err, nil)
+		case errors.Is(err, repository.ErrPodNotRunning):
+			a.conflictResponse(w, r, err, nil)
```
christian-heusel: Is a `conflictResponse` really the right thing to return here? 🤔 
I'm asking because this is also what is returned in case a workspace is paused:
```
$ curl -k -H "Kubeflow-Userid: admin" https://localhost:8443/workspaces/api/v1/workspaces/default/jupyterlab-workspace/podtemplate/logs/batch
{"error":{"code":"409","message":"workspace pod is not running","cause":{}}}
```

workspaces/backend/internal/repositories/logs/repo.go
christian-heusel: I looked a bit into alternatives for it and since we already fetch the pod we can also do something like this (🤖 -generated; just want to showcase the idea):
```diff
diff --git a/workspaces/backend/internal/repositories/logs/repo.go b/workspaces/backend/internal/repositories/logs/repo.go
index f8a7917b..b37afd20 100644
--- a/workspaces/backend/internal/repositories/logs/repo.go
+++ b/workspaces/backend/internal/repositories/logs/repo.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"strings"
 
 	kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"
 	corev1 "k8s.io/api/core/v1"
@@ -87,12 +86,6 @@ func (r *LogsRepository) OpenLogStream(ctx context.Context, namespace, workspace
 
 	stream, err := req.Stream(ctx)
 	if err != nil {
-		// When previous=true but the container has never restarted, the Kubernetes
-		// API returns a 400 with a "previous terminated container ... not found"
-		// message. Surface this as a semantic error instead of a generic 500.
-		if opts.Previous && apierrors.IsBadRequest(err) && strings.Contains(err.Error(), "previous terminated container") {
-			return nil, ErrPreviousLogsNotFound
-		}
 		return nil, fmt.Errorf("failed to open log stream for pod %s, container %s: %w", podName, containerName, err)
 	}
 	return stream, nil
@@ -145,47 +138,52 @@ func (r *LogsRepository) resolvePodAndContainer(ctx context.Context, namespace,
 		containerName = podStatus.Containers[0].Name
 	}
 
-	// When requesting current (not previous) logs, ensure the target container has
-	// actually started by inspecting the live Pod status. A container still in the
-	// Waiting state (e.g. PodInitializing, ContainerCreating, ImagePullBackOff) has
-	// no current log stream yet, and the Kubernetes API would return an opaque error;
-	// surface it as a semantic 409 instead. Previous logs are exempt, since a
-	// terminated instance can have logs even while the current instance is Waiting.
-	if !opts.Previous {
-		if err := r.ensureContainerStarted(ctx, namespace, podName, containerName); err != nil {
-			return "", "", err
+	// Inspect the live Pod status to determine whether the requested log stream can
+	// actually exist. Deciding this up front from the typed status lets us return a
+	// semantic error instead of interpreting the opaque error the Kubernetes API
+	// would otherwise return.
+	cs, err := r.findContainerStatus(ctx, namespace, podName, containerName)
+	if err != nil {
+		return "", "", err
+	}
+	if opts.Previous {
+		// A previous instance exists only if the container has terminated at least
+		// once. This holds even while the current instance is Waiting (e.g. a
+		// container in CrashLoopBackOff still has logs from its last run).
+		if cs.LastTerminationState.Terminated == nil {
+			return "", "", ErrPreviousLogsNotFound
 		}
+	} else if cs.State.Waiting != nil {
+		// A container still in the Waiting state (e.g. PodInitializing,
+		// ContainerCreating, ImagePullBackOff) has never started, so it has no log
+		// stream for the current instance yet.
+		return "", "", ErrContainerNotRunning
 	}
 
 	return podName, containerName, nil
 }
 
-// ensureContainerStarted checks the live Pod status and returns ErrContainerNotRunning
-// if the target container is still in the Waiting state (i.e. it has not started yet
-// and therefore has no logs available for the current instance).
-func (r *LogsRepository) ensureContainerStarted(ctx context.Context, namespace, podName, containerName string) error {
+// findContainerStatus returns the live status of the named container, searching both
+// the regular and init container statuses of the pod.
+func (r *LogsRepository) findContainerStatus(ctx context.Context, namespace, podName, containerName string) (*corev1.ContainerStatus, error) {
 	pod, err := r.clientset.CoreV1().Pods(namespace).Get(ctx, podName, metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			// The Workspace status references a pod that no longer exists.
-			return ErrPodNotRunning
+			return nil, ErrPodNotRunning
 		}
-		return fmt.Errorf("failed to get pod %s: %w", podName, err)
+		return nil, fmt.Errorf("failed to get pod %s: %w", podName, err)
 	}
 
-	// Search both regular and init container statuses for the target container.
 	for _, group := range [][]corev1.ContainerStatus{pod.Status.ContainerStatuses, pod.Status.InitContainerStatuses} {
-		for _, cs := range group {
-			if cs.Name != containerName {
-				continue
-			}
-			// A container that is still Waiting has never started and has no logs yet.
-			if cs.State.Waiting != nil {
-				return ErrContainerNotRunning
+		for i := range group {
+			if group[i].Name == containerName {
+				return &group[i], nil
 			}
-			return nil
 		}
 	}
 
-	return ErrContainerNotRunning
+	// The container is declared in the pod spec but has no status yet, so it has not
+	// started and has no logs.
+	return nil, ErrContainerNotRunning
 }
```

christian-heusel (commented):
Great work @YoyinZyc this is already quite awesome 🔥

I looked through the code a bit and left a review below! Feel free to apply / challenge / ignore it as needed 🤗

workspaces/backend/internal/repositories/podlogs/repo.go
YoyinZyc: Done

workspaces/backend/internal/repositories/logs/repo.go
YoyinZyc: Done

workspaces/backend/api/workspace_podtemplate_logs_handler_test.go
YoyinZyc: Done

workspaces/backend/api/workspace_logs_handler.go
YoyinZyc: Moved the params to constants/query_params.go

workspaces/backend/api/workspace_podtemplate_logs_handler_test.go
YoyinZyc: Added the err message assertion for all test cases.

workspaces/backend/api/workspace_podtemplate_logs_handler.go
YoyinZyc: updated

workspaces/backend/internal/repositories/logs/repo.go
YoyinZyc: yeah I updated to check the `LastTerminationState.Terminated` if the previous is set.

workspaces/backend/internal/repositories/logs/repo.go
YoyinZyc: using `main` makes more sense as it is explicitly declared in the controller. Updated.

workspaces/backend/internal/repositories/logs/repo.go
YoyinZyc: done

workspaces/backend/api/workspace_logs_handler.go
YoyinZyc: Comparing with using `kubectl.kubernetes.io/default-container`, I think explicitly using `main` makes more sense. changed to use `main` as the default container.

workspaces/backend/internal/repositories/logs/repo.go
YoyinZyc: changed to check `LastTerminationState.Terminated` to decide if we should raise error `ErrPreviousLogsNotFound`. We will no long have the concern on the third party error string.

workspaces/backend/api/constants/paths.go
YoyinZyc: done

workspaces/backend/internal/models/workspaces/podtemplate/logs/types.go
YoyinZyc: I think it is not common to reject unknown param? Fixed  https://github.com/kubeflow/notebooks/pull/1267/changes#r3708558872

workspaces/backend/api/workspace_logs_handler.go
YoyinZyc: Moved

workspaces/backend/api/workspace_podtemplate_logs_handler.go
YoyinZyc: I think it should be conflict as it falls in to the condition that "Dependent resource deletion"? Or do you think we should return other code? It is also something proposed in the design 
image

workspaces/backend/api/workspace_podtemplate_logs_handler.go
YoyinZyc: updated

workspaces/backend/internal/repositories/logs/repo.go
YoyinZyc: changed to two loops with helper inner function

workspaces/backend/internal/repositories/logs/repo.go
YoyinZyc: done

workspaces/backend/api/workspace_logs_handler.go
YoyinZyc: Done

workspaces/backend/api/response_errors.go
YoyinZyc: I added the new method is because I found other errors return the concrete underlying message like 409 or 413. But I am ok to reuse the existing helper as for now the "not found" basically indicates the "workspace not found"

YoyinZyc:
@andyatmiami @christian-heusel thank you both for reviewing the pr. I hope I didn't miss your comment and resolved most of your concerns. Please take another look. Thanks!

workspaces/backend/internal/repositories/logs/repo.go
```diff
@@ -0,0 +1,220 @@
...
+		LimitBytes: &safeLimitBytes,
+		Previous:   opts.Previous,
+		Follow:     false,
+		Timestamps: true,
+		SinceTime:  opts.SinceTime,
+	})
+
+	stream, err := req.Stream(ctx)
+	if err != nil {
+		return nil, ErrOpenLogStream
```
andyatmiami: The `ErrOpenLogStream` sentinel discards the original error — and is not checked by the handler (falls to `default → 500`), so the only effect is losing diagnostic context in the server-side log.

Christian's suggested refactoring in [this comment](https://github.com/kubeflow/notebooks/pull/1267#discussion_r3708675866) already addresses this with the codebase-consistent pattern:

```go
return nil, fmt.Errorf("failed to open log stream for pod %s, container %s: %w", podName, containerName, err)
```

This matches the existing precedent in `workspaces/repo.go` (e.g., `"failed to patch workspace: %w"`) and eliminates the unused sentinels. It looks like this part of Christian's feedback wasn't applied in `66b0e5c` — could you pick it up?

workspaces/backend/internal/repositories/logs/repo.go
```diff
@@ -0,0 +1,220 @@
...
+// The checker utilizes the clientset to fetch the live Pod status,
+// which is more accurate than the Workspace's cached status.
+func (r *LogsRepository) ensureLogsAvailable(ctx context.Context, namespace, podName, containerName string, previous bool) error {
+	pod, err := r.clientset.CoreV1().Pods(namespace).Get(ctx, podName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// The Workspace status references a pod that no longer exists.
+			return ErrPodNotRunning
+		}
+		return ErrGetPod
```
andyatmiami: The `ErrGetPod` sentinel discards the original error — and is not checked by the handler (falls to `default → 500`), so the only effect is losing diagnostic context in the server-side log.

Christian's suggested refactoring in [this comment](https://github.com/kubeflow/notebooks/pull/1267#discussion_r3708675866) already addresses this with the codebase-consistent pattern:

```go
return nil, fmt.Errorf("failed to get pod %s: %w", podName, err)
```

This matches the existing precedent in `workspaces/repo.go` (e.g., `"failed to patch workspace: %w"`) and eliminates the unused sentinels. It looks like this part of Christian's feedback wasn't applied in `66b0e5c` — could you pick it up?

workspaces/backend/api/workspace_podtemplate_logs_handler.go
andyatmiami: I'll raise my hand here to say I think I got this wrong in the proposal.. I should have consulted what `k8s` API does in these scenarios - and in all cases where we are recommending `409` for this API... `k8s` consistently returns `400`

Image

So I think we should (naturally!) follow `k8s` lead here - and I apologize I didn't consider this initially in the design.

Furthermore - we have no instances (currently) in the codebase for returning `409` on `GET` calls... while HTTP Spec is (frustratingly) vague and doesn't explicit prohibit `409` on "read" calls - its not a space I think we should break new ground on upon further reflection.

workspaces/backend/api/workspace_podtemplate_logs_handler.go
```diff
@@ -0,0 +1,155 @@
...
+// parseLogOptions parses and validates the log-related query parameters.
+func parseLogOptions(r *http.Request) (*models.LogOptions, field.ErrorList) {
+	var valErrs field.ErrorList
+	query := r.URL.Query()
+
+	opts := &models.LogOptions{}
+
+	if raw := query.Get(constants.ContainerQueryParam); raw != "" {
+		valErrs = append(valErrs, helper.ValidateKubernetesContainersName(field.NewPath(constants.ContainerQueryParam), raw)...)
+		opts.Container = raw
```
andyatmiami: Super minor nit - but in all other query param checks... we only set a value if an error is **NOT** returned..

This is the one inconsistency where we set the value anyway (vs. have this in an `else` block)

I see no need for this inconsistency - but let me know if I am overlooking something!

workspaces/backend/api/workspace_podtemplate_logs_handler.go
andyatmiami: (note: i will also retroactively update proposal to call this out!)

workspaces/backend/internal/repositories/podlogs/repo.go
andyatmiami, line 1: feel free to push back here - but after consulting with some folks.. `podlogs` might be a better name for this repository.
- and likewise, `PodLogsRepository` + `NewPodLogsRepository`

---

ℹ️ Prior conventions have us including repository logic for "nested" paths in the "owned" CRD - i.e. for APIs nested under `Workspaces` - we would include that logic in `workspaces/repo.go`.  **HOWEVER** - given the strict scoping/context of this logic - a separate repo does still feel like the right thing.. particularly as we are passing a `clientset` that would "pollute" the contract in `workspaces/repo.go`. Just wanted to share this insight as it came up during discussion and I find it relevant/helpful.

andyatmiami (changes requested):
@YoyinZyc - apologies for the back-n-forth - but subsequent review had me notice another couple small things - but hoping you find them non-controversial and easy enough to implement 🙇 

I think we should be **ready to roll** after this last round of feedback/discussion.  Reach out should you take issue with any of the comments.

workspaces/backend/api/workspace_podtemplate_logs_handler.go
YoyinZyc: Done

workspaces/backend/internal/repositories/podlogs/repo.go
YoyinZyc: I don't have strong option on the naming, `podlogs` does sound more clear. fixed

workspaces/backend/internal/repositories/logs/repo.go
YoyinZyc: fixed

workspaces/backend/internal/repositories/logs/repo.go
YoyinZyc: I think I mistakenly removed the concrete error when fixing https://github.com/kubeflow/notebooks/pull/1267#discussion_r3695958626, changed it to use a format string template.

workspaces/backend/api/workspace_podtemplate_logs_handler.go
YoyinZyc: ok, changed 409 to 400.

YoyinZyc:
@andyatmiami I applied all the fixes. Let me know if I am missing anything. Thanks for again for reviewing it so quickly.

andyatmiami (approved):
/lgtm
/approve

---

appreciate your attentiveness to getting this PR over the finish line @YoyinZyc .  Really impressive work for a first contribution 💯 

verification notes here:
- [PR-1267-TEST-RESULTS.md](https://github.com/user-attachments/files/30850385/PR-1267-TEST-RESULTS.md)
    - bot driven, human reviewed/approved

christian-heusel (changes requested):
@YoyinZyc it seems like one of your commits is missing the dco certification, could you add that?

christian-heusel (approved):

PR #1270, opened 2026-07-22
fix: inject kubeflow-userid header for Swagger UI "Try it out"

christian-heusel:
The backend's request authenticator requires the configured user id
header (default kubeflow-userid) on every request. The Swagger UI has no
way to prompt for it, so every "Try it out" call was failing with
401. Add a Swagger UI config to set the header to "admin", matching the
user bound to the admin ClusterRole in the Tilt dev cluster.

image

### DIFF
--- a/workspaces/backend/api/swagger_handler.go
+++ b/workspaces/backend/api/swagger_handler.go
@@ -17,21 +17,104 @@ limitations under the License.
 package api
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 
 	"github.com/julienschmidt/httprouter"
 	httpSwagger "github.com/swaggo/http-swagger/v2"
+	"github.com/swaggo/swag"
 
+	"github.com/kubeflow/notebooks/workspaces/backend/api/constants"
 	"github.com/kubeflow/notebooks/workspaces/backend/openapi"
 )
 
+// swaggerDocPath is the internal route path (rooted at constants.PathPrefix) at which the Go
+// handler sees requests for the OpenAPI definition. This is distinct from the external browser
+// path used by httpSwagger.URL() (see openapi.SwaggerInfo.BasePath below).
+var swaggerDocPath = constants.PathPrefix + "/swagger/doc.json"
+
+const (
+	// userIdSecurityScheme is the name of the security scheme which we inject into the OpenAPI
+	// definition served by the Swagger UI, so users can set the user id header with the
+	// "Authorize" button.
+	//
+	// NOTE: the name of the user id header is configurable (see EnvConfig.UserIdHeader), so this
+	//       scheme can only be resolved at runtime, and is not part of `openapi/swagger.json`.
+	userIdSecurityScheme = "UserIdHeader"
+
+	// defaultSwaggerUserId is the user id which the Swagger UI is pre-authorized with, so
+	// "Try it out" works without any manual steps.
+	//
+	// NOTE: this matches the user which is bound to the admin ClusterRole in the Tilt dev cluster.
+	defaultSwaggerUserId = "admin"
+)
+
 func (a *App) GetSwaggerHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	// intercept the OpenAPI definition, so we can inject the user id header security scheme
+	if r.URL.Path == swaggerDocPath {
+		a.serveSwaggerDoc(w, r)
+		return
+	}
+
 	httpSwagger.Handler(
 		// this url is the one that the swagger frontend uses to fetch the OpenAPI definition
 		httpSwagger.URL(fmt.Sprintf("%s/swagger/doc.json", openapi.SwaggerInfo.BasePath)),
 		httpSwagger.DeepLinking(true),
 		httpSwagger.DocExpansion("list"),
 		httpSwagger.DomID("swagger-ui"),
+		httpSwagger.PersistAuthorization(true),
+		// pre-authorize the UI with a default user id, so "Try it out" works out of the box
+		//
+		// NOTE: this must run from the `onComplete` callback, which fires once the definition has
+		//       been fetched and rendered, rather than from `AfterScript`, which runs right after
+		//       `SwaggerUIBundle()` returns, before the security schemes have loaded (in which
+		//       case `preauthorizeApiKey` silently does nothing).
+		httpSwagger.UIConfig(map[string]string{
+			"onComplete": fmt.Sprintf(`() => { window.ui.preauthorizeApiKey(%q, %q) }`, userIdSecurityScheme, defaultSwaggerUserId),
+		}),
 	).ServeHTTP(w, r)
+}
+
+// serveSwaggerDoc serves the OpenAPI definition with the user id header security scheme injected.
+func (a *App) serveSwaggerDoc(w http.ResponseWriter, r *http.Request) {
+	doc, err := swag.ReadDoc(openapi.SwaggerInfo.InstanceName())
+	if err != nil {
+		a.serverErrorResponse(w, r, fmt.Errorf("failed to read OpenAPI definition: %w", err))
+		return
+	}
+
+	var spec map[string]any
+	if err := json.Unmarshal([]byte(doc), &spec); err != nil {
+		a.serverErrorResponse(w, r, fmt.Errorf("failed to unmarshal OpenAPI definition: %w", err))
+		return
+	}
+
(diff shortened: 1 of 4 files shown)


workspaces/backend/cmd/main.go
```diff
@@ -47,6 +47,17 @@ import (
 //	@consumes	application/json
 //	@produces	application/json
 
+//	@securityDefinitions.apikey	KubeflowUserId
+//	@in							header
+//	@name						kubeflow-userid
```
thesuperzapper: We need to discuss if we want to hard-code this in the public swagger yaml, or only inject it in the RUNNING backend swagger (which is used by the "try it out"), because the backend parameterizes this with:

https://github.com/kubeflow/notebooks/blob/d51c5531b8130a8a7cbd69f071fdb5b5a7e370e6/workspaces/backend/cmd/main.go#L70-L75

Also, when notebooks v2 is deployed in prod, you usually inject auth with oauth2-proxy cookies.

But if we are simply trying to make dev easier, we probably should just inject this section of the swagger doc at runtime in `workspaces/backend/api/swagger_handler.go`.

christian-heusel: Addressed in the latest revision 🤗

workspaces/backend/api/swagger_handler.go
```diff
@@ -17,21 +17,99 @@ limitations under the License.
...
+	// defaultSwaggerUserId is the user id which the Swagger UI is pre-authorized with, so
+	// "Try it out" works without any manual steps.
+	//
+	// NOTE: this matches the user which is bound to the admin ClusterRole in the Tilt dev cluster.
+	defaultSwaggerUserId = "admin"
+)
+
 func (a *App) GetSwaggerHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	// intercept the OpenAPI definition, so we can inject the user id header security scheme
+	if strings.HasSuffix(r.URL.Path, "/swagger/doc.json") {
```
andyatmiami: **Suggestion**: Replace `strings.HasSuffix` with an exact path match

The suffix check on the URL path is looser than it needs to be:

`if strings.HasSuffix(r.URL.Path, "/swagger/doc.json") {`

Since `httprouter` doesn't allow registering a specific route alongside a *any wildcard (it panics), this in-handler dispatch is the right approach — but we can tighten it by comparing against the exact internal route path.

There are two different paths at play here:
- Internal route path (`/api/v1/swagger/doc.json`) — what the Go handler sees in `r.URL.Path`, rooted at `constants.PathPrefix`
- External browser path (e.g. `/workspaces/api/v1/swagger/doc.json`) — what `openapi.SwaggerInfo.BasePath` resolves to, used by the Swagger UI to fetch the spec from the browser

The intercept check needs to match the internal path, while `httpSwagger.URL()` needs the external one. Suggested change:

```var swaggerDocPath = constants.PathPrefix + "/swagger/doc.json"```

Then in the handler:

```
if r.URL.Path == swaggerDocPath {
    a.serveSwaggerDoc(w, r)
    return
}
```

This also removes the "strings" import.

⚠️  `httpSwagger.URL()` **must** continue using `openapi.SwaggerInfo.BasePath` (the external path) since that's what the browser uses to fetch the spec — so these two values can't be collapsed into a single variable.

File: workspaces/backend/api/swagger_handler.go [hunk 3]
```diff
@@ -17,21 +17,99 @@ limitations under the License.
...
+		//       `SwaggerUIBundle()` returns, before the security schemes have loaded (in which
+		//       case `preauthorizeApiKey` silently does nothing).
+		httpSwagger.UIConfig(map[string]string{
+			"onComplete": fmt.Sprintf(`() => { window.ui.preauthorizeApiKey(%q, %q) }`, userIdSecurityScheme, defaultSwaggerUserId),
+		}),
 	).ServeHTTP(w, r)
 }
+
+// serveSwaggerDoc serves the OpenAPI definition with the user id header security scheme injected.
+func (a *App) serveSwaggerDoc(w http.ResponseWriter, r *http.Request) {
```
andyatmiami, line 80: **Suggestion**: _Cache the patched OpenAPI spec instead of rebuilding per request_

The `security` scheme injection in `serveSwaggerDoc` only depends on `a.Config.UserIdHeader`, which is set once at startup and never changes. Currently the full JSON round-trip (_read → unmarshal → mutate → marshal_) runs on every request to `/swagger/doc.json`, but the result is always identical.

Consider caching the result on first successful request using double-checked locking:

```
var (
      swaggerDocMu   sync.Mutex
      swaggerDocJSON []byte
)

func getSwaggerDoc(userIdHeader string) ([]byte, error) {
      if doc := swaggerDocJSON; doc != nil {
              return doc, nil
      }

      swaggerDocMu.Lock()
      defer swaggerDocMu.Unlock()

      // double-check after acquiring lock
      if swaggerDocJSON != nil {
              return swaggerDocJSON, nil
      }

      // read, unmarshal, mutate, marshal — same as today
      doc, err := swag.ReadDoc(openapi.SwaggerInfo.InstanceName())
      if err != nil {
              return nil, fmt.Errorf("failed to read OpenAPI definition: %w", err)
      }
      // ... unmarshal, inject securityDefinitions/security, marshal ...

      swaggerDocJSON = out
      return out, nil
}
```

Then `serveSwaggerDoc` becomes:

```
func (a *App) serveSwaggerDoc(w http.ResponseWriter, r *http.Request) {
      doc, err := a.getSwaggerDoc()
      if err != nil {
              a.serverErrorResponse(w, r, err)
              return
      }

      w.Header().Set("Content-Type", "application/json; charset=utf-8")
      if _, err := w.Write(doc); err != nil {
              a.logger.Error("failed to write OpenAPI definition", "error", err)
      }
}
```

This gives us:
- Fast path is lock-free: once cached, concurrent requests just read the []byte — no mutex contention
- Exactly-once computation: concurrent first requests serialize through the mutex, only one builds the spec
- Retryable on error: errors aren't cached, so the next request retries the build rather than being permanently stuck

---
**Addendum**: _Why not manipulate the spec programmatically via `openapi.SwaggerInfo`?_

I initially wondered if we could avoid the JSON round-trip entirely by modifying the spec as a Go object through `openapi.SwaggerInfo` (exported from the generated `docs.go`). Unfortunately, `swag.Spec` is a very thin struct — it only exposes basic metadata fields (`Version`, `Host`, `BasePath`, `Schemes`, `Title`, `Description`). There's no `SecurityDefinitions` or `Security` field. The actual OpenAPI content
lives in `SwaggerTemplate` as a raw JSON text template, not a structured Go object.

So the JSON round-trip (unmarshal → mutate → marshal) is the cleanest available approach given the library's limitations. The alternative — string-manipulating the JSON template directly — would be more fragile, not less.

andyatmiami (commented):
Some things to stew on - neither is blocking merge of review.. so let me know if you want to adopt and/or if you think its a bad idea.

given the fact this Swagger Ui is really only exposed in tilt - none of this is really necessary - but i'll share it anyway! 😈

workspaces/backend/api/swagger_handler.go
christian-heusel: Thanks for the suggestion but as briefly discussed in the huddle I'll keep things as is:

1. This endpoint is not enabled in a production setting, only called rarely and hence adding caching here would be a premature optimization that makes the code harder to maintain without much benefit.
2. In the future there might actually be a value in being able to change this at runtime: Think about a future where we don't just have `admin` & `user` but i.e. would be able to dynamically inject the user here that the user identifies as in dex / oauth2-proxy

The potential ways of implementing this are an interesting excise to think through though, I think _if_ I'd do something about making this more static/efficient I'd prefer handling this not at the application layer but rather by supplying a few cache control headers or something to have the client deal with this instead😄

workspaces/backend/api/swagger_handler.go
christian-heusel: Implemented in the latest iteration, thanks for the suggestion and making this less brittle ✅

andyatmiami (approved):
/lgtm
/approve

---

thanks @christian-heusel for address this gaping hole I left originally (unintentionally) when doing some Tilt / auth work 💯 

tested and verified we get the custom button we would expect to get - and comfortable with your reasoning for deferring the other suggestion i left on initial review.

also, for transparency - i had asked christian to remove a couple empty lines from sample workspace files _in this PR_ as i had just noticed them and they are no-ops/not worth raising a separate PR for that alone.

Issue #867, opened 2026-01-30
[TASK] Implement Activity Rules (Culling) in Controller


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

🛑 **Depends on:** #1122, #840, #1123 (all CRD schema tasks must be merged first)

This task implements the activity rules (culling) feature in the workspaces/controller component, enabling
automatic detection of idle workspaces and their subsequent pausing to optimize cluster resources. This is
part of the larger culling feature described in https://github.com/kubeflow/notebooks/issues/741.

The controller must be extended to:
1. CRD schema changes are delivered by #1122 (activityProbe), #840 (activityRules), and #1123 (Workspace status fields).
   This task depends on all three being merged first.
2. Implement activity probing via Jupyter API polling and pod exec scripts
3. Evaluate activity rules to determine when workspaces should be paused
4. Integrate with reconciliation to schedule probes and enforce culling decisions

This task focuses exclusively on the controller component. Separate tasks will address backend API and
frontend changes.

Reference Materials

- Design Document: https://gist.githubusercontent.com/harshad16/57813a9caaa56e01778efdac4e49f764/raw/0185cf
88d3dda67ad458d503bbbf3690376bf5dc/activityRules.md

---
Implementation Details

1. CRD Type Changes (api/v1beta1/)

Workspace Status Updates (workspace_types.go):
- Add lastRunningTime (metav1.Time): When workspace entered Running state
- Enhance activity status:
  - lastActivity (metav1.Time): Most recent detected user activity
  - lastUpdate (metav1.Time): Last successful probe completion
  - lastProbe: Probe metadata (startTime, endTime, result, message)
  - rules.pauseWorkspace.eligibleAfter (metav1.Time): Calculated culling timestamp

2. Webhook Validation (internal/webhook/)

WorkspaceKind Webhook (workspacekind_webhook.go):
- Validate minProbeInterval > 0
- Validate pod exec scripts start with shebang (#!)
- Validate secondsSinceActive > 15 seconds
- Ensure only one catch-all rule (empty match) per effect type, positioned last
- Require activityProbe configuration when pause rules are defined

3. Activity Probe Implementations (internal/controller/ or internal/helper/)

Jupyter Probe:
- Poll /api/status endpoint on the workspace pod
- Extract last_activity timestamp from JSON response
- Handle connection errors gracefully (failed probes don't trigger culling)

[...]

PR #1279, opened 2026-07-24
feat: implement activity rules in controller

siyuanfoundation:
closes: #867

This PR implements the following:

1. Implement activity probing via Jupyter API polling and pod exec scripts
1. Evaluate activity rules to determine when workspaces should be paused
1. Integrate with reconciliation to schedule probes and enforce culling decisions

Tested with tilt using JupyterLab WorkspaceKind with 
```
  activityRules:
    - config:
        secondsSinceActive: 300
        minRunningSeconds: 100
      match:
        matchPodConfig:
          selector:
            matchLabels:
              cpu: "1000m"
      effect:
        pauseWorkspace: true
    - config:
        secondsSinceActive: 86400
      match: {}
      effect:
        pauseWorkspace: true
```
1. verified `eligibleAfter` is properly populated, and the small-cpu workspace is paused after being idle exceeding the `secondsSinceActive`, tiny-cpu workspace is not paused due to long `secondsSinceActive`
image

2. small-cpu workspace is restarted and back running.
3. small-cpu workspace is not paused if workspace stays active, and when activity is resumed in time before the next probe.


christian-heusel:
/retitle feat: implement activity rules in controller

workspaces/controller/cmd/main.go
```diff
@@ -172,8 +186,15 @@ func main() {
...
 		Config: cfg,
+		PodExecutor: &helper.RemoteCommandExecutor{
+			Clientset:  clientset,
+			RestConfig: restConfig,
+		},
 	}).SetupWithManager(mgr, &controller.Options{
 		RateLimiter: helper.BuildRateLimiter(),
+		// allow multiple Workspaces to be reconciled (and probed) in parallel so that a
+		// slow activity probe does not block other Workspaces' reconciliation
+		MaxConcurrentReconciles: maxConcurrentReconciles,
```
christian-heusel, line 197: I looked into this a bit because I feared that allowing concurrent reconciles might introduce a race, however I convinced myself that this is not an issue since both the ws and i.e. the sts will collapse to the same key (`namespace/X`) in the workqueue and there is only one key being worked on at the time.

Is this assesment correct?

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -135,6 +145,10 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, nil
 	}
 
+	// snapshot the Workspace as fetched, so the culling logic can issue a minimal
```
christian-heusel:
```suggestion
	// snapshot the Workspace as fetched, so the activityProbe logic can issue a minimal
```

workspaces/controller/internal/controller/workspace_activity.go
christian-heusel, line 1: Given that the feature was renamed to activityProbe this should most likely also be reflected in the filename and throughout it (i.e. in `reconcileActivityCulling`,  etc.)

workspaces/controller/internal/helper/activity.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// enforce the minimum running duration guard.
+	if minRunningSeconds > 0 {
+		if lastRunningTime <= 0 {
+			return false, eligibleAfter
+		}
+		runningDurationMs := now - lastRunningTime
+		if runningDurationMs < int64(minRunningSeconds)*millisPerSecond {
+			return false, eligibleAfter
+		}
```
christian-heusel: Maybe we could do something like this and get rid of the `millisPerSecond`?
```go
runningDuration := time.Duration(now-lastRunningTime) * time.Millisecond
if runningDuration < time.Duration(minRunningSeconds)*time.Second {
      return false, eligibleAfter
}
```

workspaces/controller/internal/helper/activity.go
```diff
@@ -0,0 +1,204 @@
...
+// CalculateEligibleAfter returns the time (UNIX epoch in milliseconds) after which the Workspace
+// would become eligible for the pauseWorkspace effect, given the last activity timestamp and the
+// rule's secondsSinceActive.
+//
+// If lastActivity is unknown (<= 0), it returns 0.
+func CalculateEligibleAfter(lastActivity int64, secondsSinceActive int32) int64 {
+	if lastActivity <= 0 {
+		return 0
+	}
+	return lastActivity + int64(secondsSinceActive)*millisPerSecond
```
christian-heusel:
```suggestion
	return lastActivity + (time.Duration(secondsSinceActive) * time.Second).Milliseconds()
```

workspaces/controller/internal/helper/activity.go
```diff
@@ -0,0 +1,204 @@
...
+// Only a rule with a non-nil effect (as determined by getEffect) terminates evaluation; rules
+// where getEffect returns nil are skipped (fallthrough) for this effect type.
+//
+// namespaceLabels are the labels of the Workspace's namespace, and podConfigLabels are the
+// labels of the podConfig option the Workspace is currently using.
+func EvaluateActivityRule[T any](
+	rules []kubefloworgv1beta1.ActivityRule,
+	namespaceLabels, podConfigLabels map[string]string,
+	getEffect func(kubefloworgv1beta1.ActivityRuleEffect) *T,
+) ActivityRuleDecision[T] {
```
christian-heusel: I'm curious, why did you decide to make `EvaluateActivityRule()` generic? 🤔 AFAIU it can currently only ever be a `bool`?

workspaces/controller/internal/helper/probe.go
```diff
@@ -0,0 +1,407 @@
...
+	Result kubefloworgv1beta1.WorkspaceProbeResult
+
+	// Message is a human-readable message about the probe result.
+	Message string
+
+	// LastActivity is the activity timestamp determined by the probe (UNIX epoch in
+	// milliseconds). It is nil when the probe did not succeed or when the activity timestamp
+	// should not be updated.
+	LastActivity *int64
+}
```
christian-heusel, line 96: Any reason you chose the time-related fields in `ProbeResult` to be of type `int64`? I would have imagined that since we use this type internally as a helper its easier to use the native go time type and then convert via [`UnixMilli(int64)`](https://pkg.go.dev/time#UnixMilli) / [`time.UnixMilli()`](https://pkg.go.dev/time#Time.UnixMilli) as needed 🤔

workspaces/controller/internal/helper/probe.go
```diff
@@ -0,0 +1,407 @@
...
+	// then print the resulting JSON file to stdout so the controller can read it back.
+	// Using a single shell invocation avoids needing multiple exec round-trips.
+	scriptPath := outputPath + ".script"
+	shellCommand := fmt.Sprintf(
+		"cat > %[1]s && chmod +x %[1]s && OUTPUT_JSON_PATH=%[2]s %[1]s; rc=$?; rm -f %[1]s; "+
+			"if [ $rc -ne 0 ]; then exit $rc; fi; "+
+			"if [ -f %[2]s ]; then cat %[2]s; rm -f %[2]s; fi",
+		scriptPath, outputPath,
+	)
+	command := []string{"/bin/sh", "-c", shellCommand}
```
christian-heusel, line 316: Even though `scriptPath` can not contain spaces (`[/.a-z0-9-]`) we should apply proper variable quoting in this section, since paths generally can have spaces.

christian-heusel:
```
$ shellcheck test.sh       

In test.sh line 2:
cat > $1 && chmod +x $1 && OUTPUT_JSON_PATH=$2 $1; rc=$?; rm -f $1;
      ^-- SC2086 (info): Double quote to prevent globbing and word splitting.
                     ^-- SC2086 (info): Double quote to prevent globbing and word splitting.
                                                                ^-- SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
cat > "$1" && chmod +x "$1" && OUTPUT_JSON_PATH=$2 $1; rc=$?; rm -f "$1";


In test.sh line 4:
if [ -f $2 ]; then cat $2; rm -f $2; fi
        ^-- SC2086 (info): Double quote to prevent globbing and word splitting.
                       ^-- SC2086 (info): Double quote to prevent globbing and word splitting.
                                 ^-- SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
if [ -f "$2" ]; then cat "$2"; rm -f "$2"; fi

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

christian-heusel (commented):
Hey @siyuanfoundation, this is really awesome stuff, thanks for pushing this forward 🔥 

I don't have any big comments, but noted down a few things that struck me as odd or small improvements, feel free to apply as sensible to you.  Of course for the final review the WG leads and potentially other controller experts can chime in! 🤗 

Generally I found that we still mix the naming from culling and activityProbe, I think that should be standardized before merge 😄

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -99,6 +99,14 @@ type WorkspaceReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 	Config *config.EnvConfig
+
+	// PodExecutor executes activity probe scripts inside Workspace Pods.
+	// If nil, podExec probes are skipped (used when exec is not configured, e.g. in some tests).
```
andyatmiami: This comment _seems_ misleading - as if I am reading this code corrected... we return a `Failure` result:
- https://github.com/kubeflow/notebooks/pull/1279/changes#diff-b87d913cf3d02827ba01d1b701b76b068f60741f6082cdf72a78830ad462209aR208-R215

Perhaps that is technically "skipping" - but maybe a little more specificity here would be good

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -109,7 +117,9 @@ type WorkspaceReconciler struct {
 // +kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups=core,resources=events,verbs=get;list;watch
 // +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch
```
andyatmiami, line 126: Maybe for @thesuperzapper to weigh in here...

It makes sense we'd use the controller-runtime's cached client here.. but we are also now introducing a shared informer that list+watches all namespaces cluster-wide..

Are we concerned about cardinality here?

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -435,6 +449,16 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 	workspace.Status = workspaceStatus
 
+	// reconcile the activity probe and culling rules
```
andyatmiami:
```suggestion
	// reconcile the activity probe and activity rules
```

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -1145,6 +1211,19 @@ func (r *WorkspaceReconciler) generateWorkspaceStatus(ctx context.Context, log l
...
+	//  - only advanced on a transition INTO the Running state, so it reflects the start of
+	//    the current continuous Running period
+	//  - reset status.Activity on transition into Running so stale activity from a previous
+	//    run does not cause an immediate pause when a culled Workspace is restarted
+	if workspaceState == kubefloworgv1beta1.WorkspaceStateRunning {
+		if workspace.Status.State != kubefloworgv1beta1.WorkspaceStateRunning || status.LastRunningTime == 0 {
+			status.LastRunningTime = metav1.Now().UnixMilli()
+			status.Activity = kubefloworgv1beta1.WorkspaceActivity{}
+		}
+	}
```
andyatmiami: Doesn't seem like we need nested-ifs here (and perhaps a helper function to implement the conditional would be beneficial and a more nature place to add these helpful inline comments..

workspaces/controller/internal/controller/workspace_activity.go
andyatmiami: `workspace_activity.go` probably a good name... and yes.. `culling` is a specific _feature_ that can be configured with Activity Probe + Activity Rules - we we'd want files and functions ideally to **not** reference `culling` in the name..

user defines activity probe(s) and activity rule(s) - and an _outcome_ of those configurations could be that a workspace is culled.. but in future - if/as we add other "outcomes" - we wouldn't want `Culling` in any named files/functions as that would get confusing..

workspaces/controller/internal/controller/workspace_culling.go
```diff
@@ -0,0 +1,353 @@
...
+				StartTime: now,
+				EndTime:   now,
+				Result:    kubefloworgv1beta1.WorkspaceProbeResultFailure,
+				Message:   fmt.Sprintf("Jupyter probe failed: port %q not found in imageConfig", activityProbe.Jupyter.PortId),
+			}
+		}
+		prober := r.HTTPProber
+		if prober == nil {
+			prober = &helper.DefaultHTTPProber{Client: &http.Client{Timeout: defaultJupyterProbeTimeout}}
+		}
```
andyatmiami: In production, `HTTPProber` is never set (not wired in `cmd/main.go`), so every Jupyter probe creates a new `http.Client` with its own transport and connection pool. This prevents connection reuse across probes to the same Pod IP and creates idle connection churn at scale. `PodExecutor` is already initialized once at startup — `HTTPProber` should maybe follow the same pattern?

I wasn't sure if this was an oversight - or if there was a technical reason for this decision...

workspaces/controller/internal/controller/workspace_controller_test.go
```diff
@@ -226,6 +227,160 @@ var _ = Describe("Workspace Controller", func() {
...
+			}
+
+			By("reconciling until spec.paused=true is persisted to the API server")
+			Eventually(func(g Gomega) {
+				updatedWS := &kubefloworgv1beta1.Workspace{}
+				if err := k8sClient.Get(ctx, req.NamespacedName, updatedWS); err == nil && ptr.Deref(updatedWS.Spec.Paused, false) {
+					return
+				}
+
+				_, err := reconciler.Reconcile(ctx, req)
```
andyatmiami: Is there any chance of this `Reconcile` conflicting with the `Reconcile` that is kicked off in the `BeforeSuite` ?
- https://github.com/kubeflow/notebooks/pull/1279/changes#diff-97af0f58b55722009178ea41b67fba506680a124cbf3bd6426e971ea8573e096R125-R129

Such that this could result in test flakiness (naturally at the more _inopportune of times_ 😎 ) ?

workspaces/controller/internal/controller/workspace_culling.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// only pause when the matched rule opts in, the Workspace is eligible, AND the decision is
+	// backed by a fresh probe in this reconcile (allowPause).
+	//  - a matched rule with pauseWorkspace: false exempts the Workspace from culling
+	//  - a failing probe leaves lastActivity unchanged, so eligibility is based on the
+	//    last known-good activity (never culled purely because a probe failed)
+	//  - when allowPause is false the activity data may be stale (no probe ran this reconcile),
+	//    so we must NOT pause even if the stale data says the Workspace is eligible
+	if decision.Value && eligible && allowPause {
+		workspace.Spec.Paused = ptr.To(true)
```
andyatmiami: Claude flagged this - I don't know if I ever would have seen it 🙃 

This callstack shows why the `Spec.Paused` mutation inside `evaluatePauseDecision` is dead:

  ```
Reconcile()
  │
  ├─ 1. originalWorkspace = workspace.DeepCopy()          // line 150: snapshot BEFORE any mutations
  │
  ├─ 2. reconcileActivityCulling(workspace, ...)           // line 456
  │     │
  │     ├─ runProbe() → probeResult
  │     ├─ updateActivityStatusFromProbe(workspace, ...)   // mutates workspace.Status.Activity
  │     │
  │     └─ evaluatePauseDecision(workspace, ..., allowPause=true)  // line 122
  │           │
  │           └─ workspace.Spec.Paused = ptr.To(true)      // ⚠️  line 318: SETS Spec.Paused on in-memory object
  │              return true                                // returns paused=true
  │
  │     return cullResult, paused=true, nil
  │
  ├─ 3. Status().Update(ctx, workspace)                    // line 464: sends status to API server
  │     │
  │     └─ ⚡ API server response OVERWRITES the entire     // the response has Spec.Paused = false
  │        in-memory `workspace` object                     // because Status().Update only persists status
  │                                                         //
  │        → workspace.Spec.Paused is now FALSE again       // the line 318 mutation is GONE
  │
  ├─ 4. if paused {                                        // line 481: yes, paused=true (the bool)
  │     │
  │     ├─ workspace.Spec.Paused = ptr.To(true)            // line 482: sets it AGAIN (this is the real one)
  │     │
  │     └─ Patch(ctx, workspace, MergeFrom(originalWorkspace))  // line 483: persists the diff
  │        │                                                     // original had Paused=false
  │        └─ sends only: {"spec":{"paused":true}}               // current has Paused=true → patch works
```

The mutation at line 318 inside `evaluatePauseDecision` has no effect because step 3 (`Status().Update`) replaces the in-memory workspace with the API  server's response, which still has Spec.Paused = false. The caller then re-sets it at line 482 before the actual Patch.

The suggestion is to remove the workspace.Spec.Paused = ptr.To(true) from evaluatePauseDecision (line 318) and just have it return the boolean — keep the actual spec mutation solely in the caller where it's persisted.

File: workspaces/controller/internal/controller/workspace_culling.go [hunk 15]
```diff
@@ -0,0 +1,353 @@
...
+//
+// allowPause gates whether an eligible Workspace is actually paused:
+//   - when true (called immediately after a fresh probe), an eligible Workspace is paused.
+//   - when false (called when a probe is NOT due), the `eligibleAfter` status is still refreshed
+//     for the UI, but the Workspace is never paused. This is critical: pausing MUST be backed by
+//     a fresh probe in the same reconcile, otherwise a Workspace could be culled based on a stale
+//     `lastActivity` even though the user resumed activity since the last probe.
+//
+// namespaceLabels and podConfigLabels are the labels used for rule matching.
+func (r *WorkspaceReconciler) evaluatePauseDecision(
```
andyatmiami: `evaluatePauseDecision` never uses its receiver `r`..

should this be a static method?

workspaces/controller/internal/controller/workspace_culling.go
```diff
@@ -0,0 +1,353 @@
...
+	workspace.Status.Activity.LastProbe = &kubefloworgv1beta1.WorkspaceActivityLastProbe{
+		StartTime: result.StartTime,
+		EndTime:   result.EndTime,
+		Result:    result.Result,
+		Message:   result.Message,
+	}
+
+	if result.Succeeded() {
+		if result.LastActivity != nil {
+			workspace.Status.Activity.LastActivity = *result.LastActivity
```
andyatmiami: The probe result unconditionally overwrites `LastActivity` without checking whether the new value is >= the current value. A misbehaving or compromised probe script could report a future timestamp (preventing culling indefinitely) or a past timestamp (triggering premature culling). This is aligned with the CRD contract (the probe is authoritative), but I want the implications to be explicitly acknowledged:
   - **Future timestamp**: workspace is never eligible for pause until wall-clock catches up.
   - **Past timestamp**: workspace appears idle longer than it actually has been, potentially triggering an early pause.
   
@thesuperzapper / @siyuanfoundation : just wanna double check and onfirm this is acceptable behavior?

workspaces/controller/internal/helper/activity.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	return result
+}
+
+// CalculateEligibleAfter returns the time (UNIX epoch in milliseconds) after which the Workspace
+// would become eligible for the pauseWorkspace effect, given the last activity timestamp and the
+// rule's secondsSinceActive.
+//
+// If lastActivity is unknown (<= 0), it returns 0.
+func CalculateEligibleAfter(lastActivity int64, secondsSinceActive int32) int64 {
```
andyatmiami, line 191: When a probe succeeds but reports no activity timestamp (e.g., `podExec` script returns `has_activity: false` without `last_activity`), `LastActivity` stays at `0`. On subsequent reconciles, `evaluatePauseDecision` computes `eligibleAfter = 0` (via `CalculateEligibleAfter`) and writes it to the status.

 The UI sees `EligibleAfter: 0`, which looks like "eligible immediately." But the controller will **never** pause in this state because `IsEligibleForPause` returns `false` when `lastActivity <= 0` ("never cull based on unknown activity time"). The status field and the actual controller behavior disagree.

Should `PauseWorkspace` be set to `nil` (rather than `EligibleAfter: 0`) when `lastActivity` is unknown? Or should the frontend interpret `EligibleAfter: 0` as "not applicable"? Either way, the contract should be documented.

@thesuperzapper - probably good for you to weigh in too here on expected behavior

workspaces/controller/internal/helper/probe.go
```diff
@@ -0,0 +1,407 @@
...
+	if timeout > 0 {
+		execCtx, cancel = context.WithTimeout(ctx, timeout)
+		defer cancel()
+	}
+
+	// Write the script to the output-path directory, execute it with OUTPUT_JSON_PATH set,
+	// then print the resulting JSON file to stdout so the controller can read it back.
+	// Using a single shell invocation avoids needing multiple exec round-trips.
+	scriptPath := outputPath + ".script"
+	shellCommand := fmt.Sprintf(
```
andyatmiami, line 310: The shell command runs the probe script and then `cat`s the `OUTPUT_JSON_PATH` file — both write to the same `stdout` buffer. If the probe script prints anything to stdout (debug output, library banners, `echo` statements), `parsePodExecOutput` will attempt to `json.Unmarshal` the combined output and fail with "invalid JSON file". This is a subtle, hard-to-debug failure mode: script authors would reasonably expect their stdout to be ignored since the CRD contract says to write results to `OUTPUT_JSON_PATH`.

workspaces/controller/internal/helper/probe.go
```diff
@@ -0,0 +1,407 @@
...
+		"cat > %[1]s && chmod +x %[1]s && OUTPUT_JSON_PATH=%[2]s %[1]s; rc=$?; rm -f %[1]s; "+
+			"if [ $rc -ne 0 ]; then exit $rc; fi; "+
+			"if [ -f %[2]s ]; then cat %[2]s; rm -f %[2]s; fi",
+		scriptPath, outputPath,
+	)
+	command := []string{"/bin/sh", "-c", shellCommand}
+
+	var stdout, stderr bytes.Buffer
```
andyatmiami: `RunJupyterProbe` correctly bounds its response with `io.LimitReader(resp.Body, maxProbeResponseBytes)` (1 MiB, line 238), but this uses an unbounded `bytes.Buffer`. 

A misbehaving probe script writing large output to `OUTPUT_JSON_PATH` (then `cat`'d to stdout) could cause memory pressure on the controller.

workspaces/controller/internal/controller/workspace_culling.go
```diff
@@ -0,0 +1,353 @@
...
+	if err := r.Get(ctx, client.ObjectKey{Name: namespaceName}, ns); err != nil {
+		if apierrors.IsNotFound(err) {
+			return map[string]string{}, nil
+		}
+		return nil, err
+	}
+	if ns.Labels == nil {
+		return map[string]string{}, nil
+	}
+	return ns.Labels, nil
```
andyatmiami: The `Namespace` object is from the controller-runtime cache. While current versions deep-copy on `Get`, returning `maps.Clone(ns.Labels)` would make the safety guarantee explicit and future-proof.

workspaces/controller/internal/helper/activity.go
```diff
@@ -0,0 +1,204 @@
...
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
+
+// selectorMatches converts a metav1.LabelSelector and evaluates it against the given labels.
+func selectorMatches(selector *metav1.LabelSelector, lbls map[string]string) (bool, error) {
+	sel, err := metav1.LabelSelectorAsSelector(selector)
```
andyatmiami, line 165: `EvaluateActivityRule` calls `metav1.LabelSelectorAsSelector()` per rule per reconcile, parsing and compiling the selector each time. For a small number of rules this is fine, but with thousands of workspaces all referencing the same WorkspaceKind, the same selectors are recompiled on every reconcile. Not a blocker — selector compilation is fast — but at extreme scale (10K+ workspaces, frequent probes), caching compiled selectors per WorkspaceKind would eliminate redundant work.

workspaces/controller/internal/controller/workspace_activity.go
```diff
@@ -0,0 +1,353 @@
...
+		return ctrl.Result{}, false, nil
+	}
+
+	now := metav1.Now().UnixMilli()
+
+	minProbeInterval := time.Duration(ptr.Deref(activityProbe.MinProbeIntervalSeconds, kubefloworgv1beta1.DefaultMinProbeIntervalSeconds)) * time.Second
+	probeInterval := time.Duration(ptr.Deref(activityProbe.ProbeIntervalSeconds, kubefloworgv1beta1.DefaultProbeIntervalSeconds)) * time.Second
+
+	// gather the labels used by activity rule matching.
+	namespaceLabels, err := r.getNamespaceLabels(ctx, workspace.Namespace)
```
andyatmiami, line 98: ℹ️ `getNamespaceLabels` is called on every reconcile regardless of whether any `activityRules` define a `matchNamespace`.

I just want to transparently call this out - but I don't think the additional complexity is warranted to try and defer this call - as then we'd seemingly be required to iterate rules 2x... negating any desired performance benefit.

workspaces/controller/internal/helper/activity.go
```diff
@@ -0,0 +1,204 @@
...
+
+		effect := getEffect(rule.Effect)
+		if effect == nil {
+			// this rule does not configure the requested effect, fall through.
+			continue
+		}
+
+		matched, err := activityRuleMatches(rule.Match, namespaceLabels, podConfigLabels)
+		if err != nil {
+			// treat an invalid selector as non-matching (validation should prevent this).
```
andyatmiami: Part of me would want at least a debug log (or something) here to alert us if we someone entered this logic branch.... but then threading a logger into this function seems overkill...

WDYT?

workspaces/controller/internal/helper/probe.go
```diff
@@ -0,0 +1,407 @@
...
+		VersionedParams(&corev1.PodExecOptions{
+			Container: container,
+			Command:   command,
+			Stdin:     stdin != nil,
+			Stdout:    stdout != nil,
+			Stderr:    stderr != nil,
+			TTY:       false,
+		}, scheme.ParameterCodec)
+
+	exec, err := remotecommand.NewSPDYExecutor(e.RestConfig, http.MethodPost, req.URL())
```
andyatmiami: ⚠️ This is another Claude review flag - I'd prefer to get your opinion here - and push back if/as necessary!

This project currently targets Kubernetes 1.31+ and uses client-go v0.31.0, where WebSocket-based exec is GA. This proposed implementation uses `remotecommand.NewSPDYExecutor` (the legacy SPDY protocol), which still works but is the older path. The Kubernetes ecosystem is moving toward WebSocket exec, and some managed Kubernetes offerings may eventually deprecate the SPDY exec endpoint. Consider using `NewFallbackExecutor` which tries WebSocket first and falls back to SPDY for broader compatibility

andyatmiami (changes requested):
@siyuanfoundation ... don't let number of comments here "get it twisted" - this is a very impressive contribution and its obvious the effort/attention you put into this - so thank you.

Aside from the comments I have explicitly called out in the review - I also wanetd to flag a general pattern..

There are 15+ error and result message strings scattered across `RunJupyterProbe`, `RunPodExecProbe`, `runProbe`, and `parsePodExecOutput`, many sharing repeated prefixes like `"Jupyter probe failed: "` and `"PodExec probe failed: "`. Extracting the prefixes and common messages as constants at the top of the file (alongside the existing `maxProbeResponseBytes`, `probeContainerName`, `jupyterStatusPath` constants) would:
    - Make the full message contract visible at a glance
    - Reduce prefix duplication
    - Make it easier to keep messaging consistent if new probe types are added


ℹ️ As always - feel free to question/push back on any of these comments...  I did a lot of staring at call stacks while reviewing - so its possible I am just "confused" on some of my feedback - keep that in mind 😇

workspaces/controller/cmd/main.go
siyuanfoundation: yes, it is common to use `MaxConcurrentReconciles` in controllers. Only different resources are processed simultaneously.

workspaces/controller/internal/helper/activity.go
siyuanfoundation: It is explained above
```
// The type parameter T allows supporting both boolean effects (like pauseWorkspace) and
// potential future non-boolean effects (like logRetentionDays or resourceLimits).
```

workspaces/controller/internal/controller/workspace_culling.go
siyuanfoundation: good catch! Missed that

workspaces/controller/internal/helper/probe.go
siyuanfoundation: makes sense

workspaces/controller/internal/controller/workspace_culling.go
siyuanfoundation: done

workspaces/controller/internal/controller/workspace_controller.go
siyuanfoundation: the number of namespaces are usually orders of magnitudes smaller than pods, and their sizes are very small. I don't think that should be a concern. This is far better than hitting the API Server for every workspace.

workspaces/controller/internal/controller/workspace_controller.go
siyuanfoundation: updated

workspaces/controller/internal/helper/probe.go
siyuanfoundation: good point. Updated

workspaces/controller/internal/controller/workspace_culling.go
siyuanfoundation: I think that's acceptable. I have updated the CRD doc to be more explicit about it.

workspaces/controller/internal/helper/activity.go
siyuanfoundation: modified to return the err to be consistent with `LabelSelectorAsSelector`

workspaces/controller/internal/helper/probe.go
siyuanfoundation: done

workspaces/controller/internal/helper/activity.go
siyuanfoundation: good point. Added cache

workspaces/controller/internal/controller/workspace_activity.go
siyuanfoundation: makes sense. renamed.

workspaces/controller/internal/helper/activity.go
siyuanfoundation: done

workspaces/controller/internal/helper/probe.go
siyuanfoundation: I was mainly thinking about consistency with the CRD and it is only really used to pass the values. I guess if think about adding metrics in the future, it makes sense to use Time here.

workspaces/controller/internal/controller/workspace_culling.go
siyuanfoundation: Good catch. moved to the startup.

workspaces/controller/internal/controller/workspace_culling.go
siyuanfoundation: done

workspaces/controller/internal/helper/probe.go
siyuanfoundation: added buffer limit

workspaces/controller/internal/controller/workspace_activity.go
siyuanfoundation: ack

workspaces/controller/internal/controller/workspace_controller_test.go
siyuanfoundation: ah, would have missed that. Should be good now.

workspaces/controller/internal/helper/activity.go
siyuanfoundation: my preference is `the frontend should interpret EligibleAfter: 0 as "not applicable"`. The reasons are
1. there are many other timestamps like `PauseTime` `LastActivity` that default to 0. So generally 0 is not a valid actionable timestamp.
2. in the current implementation `IsEligibleForPause=nil` means no rule is matched for Pause. Keeping `PauseWorkspace != nil` allows the UI to display that auto-pause is enabled/configured

andyatmiami: OK, that seems reasonable to me @siyuanfoundation - thanks for providing rationale there 💯 

👀 @thaorell @paulovmr 

(w.r.t. frontend _interpretations_ )

workspaces/controller/internal/helper/activity.go
andyatmiami: seems a reasonable safeguard that isn't overly complicated (even if we don't need to immediately rely on this abstraction.)

considering this discussion resolved - but anyone can feel free to re-open if they feel so inclined.

workspaces/controller/internal/controller/workspace_activity.go
```diff
@@ -0,0 +1,363 @@
...
+			return &helper.ProbeResult{
+				StartTime: now,
+				EndTime:   now,
+				Result:    kubefloworgv1beta1.WorkspaceProbeResultFailure,
+				Message:   fmt.Sprintf("%sport %q not found in imageConfig", helper.ProbeMessagePrefixJupyterFailed, activityProbe.Jupyter.PortId),
+			}
+		}
+		if r.HTTPProber == nil {
+			r.HTTPProber = &helper.DefaultHTTPProber{Client: &http.Client{Timeout: defaultJupyterProbeTimeout}}
+		}
```
andyatmiami, line 212: This is definitely an improvement over the _initialize per call_ pattern - but I feel like this code still has 2 minor issues.

1. this init is a real data race under MaxConcurrentReconciles: 10
    - admittedly benign in actual "impact" - we just create more than we need - and "orphans" eventually get GC'd

2. inconsistent initialization pattern as compared to `PodExecutor`


I wonder if we could/should init `HTTPProber` alongside `PodExecutor`?  It would then be consistent in its implementation style and truly be a "run once" paradigm 🤔

workspaces/controller/internal/helper/activity.go
```diff
@@ -0,0 +1,230 @@
...
+//
+// A Workspace is eligible when:
+//   - the current time is at or after eligibleAfter (lastActivity + secondsSinceActive), AND
+//   - the Workspace has been running for at least minRunningSeconds (based on lastRunningTime).
+//
+// lastActivity, lastRunningTime and now are all UNIX epoch in milliseconds.
+func IsEligibleForPause(lastActivity, lastRunningTime, now int64, secondsSinceActive, minRunningSeconds int32) (bool, int64) {
+	eligibleAfter := CalculateEligibleAfter(lastActivity, secondsSinceActive)
+
+	// never cull based on an unknown activity time.
```
andyatmiami:
```suggestion
	// never pause based on an unknown activity time.
```

workspaces/controller/internal/helper/probe.go
```diff
@@ -0,0 +1,450 @@
...
+	LastActivity *time.Time
+}
+
+// Succeeded reports whether the probe completed successfully.
+func (r *ProbeResult) Succeeded() bool {
+	return r.Result == kubefloworgv1beta1.WorkspaceProbeResultSuccess
+}
+
+// newFailureResult builds a failed ProbeResult with the given start time and message. The end
+// time is set to the current time, and LastActivity is left nil so failing probes never cull.
```
andyatmiami:
```suggestion
// time is set to the current time, and LastActivity is left nil so failing probes never trigger any activity rule effects.
```

workspaces/controller/internal/helper/probe.go
```diff
@@ -0,0 +1,450 @@
...
+}
+
+// RunJupyterProbe polls the Jupyter `/api/status` endpoint on the Workspace Pod and
+// extracts the `last_activity` timestamp.
+//
+// The serviceHost should be the DNS name (or IP) that resolves to the Workspace Pod, and
+// port is the container port to probe. basePath is the HTTP path prefix (if any) used by the
+// Jupyter server. A failed probe (connection error, non-2xx status, invalid body) returns a
+// ProbeResult with Result == Failure/Timeout and a nil LastActivity, so that failing probes
+// never trigger culling.
```
andyatmiami:
```suggestion
// never trigger activity rule effects.
```

workspaces/controller/internal/helper/probe.go
```diff
@@ -0,0 +1,450 @@
...
+}
+
+// RunPodExecProbe executes the given script inside the Workspace Pod via the Kubernetes exec
+// API and parses the JSON output written to a randomized OUTPUT_JSON_PATH.
+//
+// The script is executed via `/bin/sh -c` with the OUTPUT_JSON_PATH environment variable set
+// to a randomized path. The script itself provides the interpreter via its shebang. After the
+// script exits with code 0, the JSON output file is read back and parsed. A failed probe
+// (non-zero exit code, timeout, missing/invalid JSON) returns a ProbeResult with a nil
+// LastActivity so that failing probes never trigger culling.
```
andyatmiami:
```suggestion
// LastActivity so that failing probes never trigger activity rule effects.
```

workspaces/controller/test/e2e/e2e_test.go
```diff
@@ -405,4 +434,686 @@ var _ = Describe("controller", Ordered, func() {
...
+			patchAnnotation := func() error {
+				cmd := exec.Command("kubectl", "annotate", "workspace", staleWorkspaceName,
+					"-n", workspaceNamespace, fmt.Sprintf("test.reconcile=%d", time.Now().UnixNano()), "--overwrite")
+				_, err := utils.Run(cmd)
+				return err
+			}
+			Eventually(patchAnnotation, timeout, interval).Should(Succeed())
+
+			By("verifying that the Workspace is NOT paused at eligibleAfter because a fresh probe has not run yet")
+			checkNotCulledYet := func(g Gomega) {
```
andyatmiami:
```suggestion
			checkNotPausedYet := func(g Gomega) {
```

workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -354,7 +354,13 @@ type ActivityProbe struct {
...
 	// the desired interval in seconds between successful probes.
 	// - If a probe succeeds, the controller schedules the next probe after this duration (requeuing after probeInterval).
-	// - Determines the freshness of workspace activity status used for culling inactive workspaces.
+	// - Determines the freshness of workspace activity status used by activity rules.
+	// - ACTIVITY TIMING CAVEAT: a Workspace is only paused immediately after a fresh probe confirms it is still
+	//   inactive (a Workspace is never paused based on stale activity data, so an actively-used Workspace whose
+	//   user resumed activity between probes is not paused). Consequently, activity rules are only evaluated at probe time,
+	//   so a Workspace may keep running for up to ~probeIntervalSeconds after it first becomes eligible
+	//   (lastActivity + secondsSinceActive) before it is actually paused. Lower this value for tighter timing,
+	//   at the cost of more frequent probing.
```
andyatmiami:
```suggestion
	//  - Determines the freshness of workspace activity status used by activity rules.
	//  - ACTIVITY TIMING CAVEAT: a Workspace is only paused immediately after a fresh probe confirms it is still
	//    inactive (a Workspace is never paused based on stale activity data, so an actively-used Workspace whose
	//    user resumed activity between probes is not paused). Consequently, activity rules are only evaluated at probe time,
	//    so a Workspace may keep running for up to ~probeIntervalSeconds after it first becomes eligible
	//    (lastActivity + secondsSinceActive) before it is actually paused. Lower this value for tighter timing,
	//    at the cost of more frequent probing.
```

ℹ️ note that L356 should also have an additional whitespace before the `-`


i.e.

//  -  (two-space) bullets and //     (four-space) continuations

workspaces/controller/internal/helper/activity.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	return true, nil
+}
+
+var (
+	// selectorCache stores compiled labels.Selector instances to eliminate redundant
+	// metav1.LabelSelectorAsSelector parsing across high-frequency reconciles and large
+	// numbers of Workspaces referencing the same WorkspaceKind selectors.
+	selectorCache sync.Map // map[string]cachedSelector
```
andyatmiami, line 141: I don't think this is anything to act on - but this is _technically_ an unbounded map..

realistically I think `(number of WorkspaceKinds) × (number of distinct selectors per kind)` naturally limits this to a reasonable size.. but I just wanted to transparently call out the observation.

workspaces/controller/test/utils/common.go
andyatmiami, line 1: `sigs.k8s.io/yaml` is already a dependency. These functions could unmarshal into an unstructured.Unstructured (or even just map[string]interface{}),  set the fields programmatically, and marshal back.

(but this is a test file - and the code works and is unlikely to change... and it will fail appropriately if underlying assumptions are violated - so feel free to ignore)

workspaces/controller/internal/controller/workspace_activity.go
```diff
@@ -0,0 +1,363 @@
...
+				StartTime: now,
+				EndTime:   now,
+				Result:    kubefloworgv1beta1.WorkspaceProbeResultFailure,
+				Message:   helper.ProbeMessagePrefixPodExecFailed + "exec is not configured",
+			}
+		}
+		// NOTE: unlike the Jupyter branch, we pass the outer ctx and the timeout value
+		// separately. RunPodExecProbe derives its own timeout-bounded context internally so
+		// it can (a) report the timeout duration in the failure message and (b) distinguish a
+		// probe timeout (WorkspaceProbeResultTimeout) from outer-context cancellation.
```
andyatmiami: No request for change here - just want to bolster my understanding of the code...

Naively it seems to me we **COULD** construct `RunJupyterProbe` to behave/report the same as we do here... can you just help me understand why this branch _is different_

workspaces/controller/internal/controller/workspace_activity.go
```diff
@@ -0,0 +1,363 @@
...
+	// only pause when the matched rule opts in, the Workspace is eligible, AND the decision is
+	// backed by a fresh probe in this reconcile (allowPause).
+	//  - a matched rule with pauseWorkspace: false exempts the Workspace from pausing
+	//  - a failing probe leaves lastActivity unchanged, so eligibility is based on the
+	//    last known-good activity (never paused purely because a probe failed)
+	//  - when allowPause is false the activity data may be stale (no probe ran this reconcile),
+	//    so we must NOT pause even if the stale data says the Workspace is eligible
+	if decision.Value && eligible && allowPause {
+		return true, nil
+	}
```
andyatmiami: we already have a `decision.Value` conditional in the logic directly preceding this block...

I wonder if we embed this check inside there so the code reads more linearly?

andyatmiami: embedding this logic then helps keep the conditional `pauseWorkspace = nil` "clearout" + `return false` correlation consistent throughout the function... 

i.e. I don't think we'd need that `else if` then above

workspaces/controller/internal/controller/workspace_activity.go
```diff
@@ -0,0 +1,363 @@
...
+		if apierrors.IsNotFound(err) {
+			return map[string]string{}, nil
+		}
+		return nil, err
+	}
+	if ns.Labels == nil {
+		return map[string]string{}, nil
+	}
+	return maps.Clone(ns.Labels), nil
+}
```
andyatmiami, line 338: I **think** we could slightly simply this function to:

```
  func (r *WorkspaceReconciler) getNamespaceLabels(ctx context.Context, namespaceName string) (map[string]string, error) {
      ns := &corev1.Namespace{}
      if err := r.Get(ctx, client.ObjectKey{Name: namespaceName}, ns); err != nil {
              return map[string]string{}, nil
          }
          return nil, err
      }
      result := make(map[string]string, len(ns.Labels))
      maps.Copy(result, ns.Labels)
      return result, nil
  }
```

just less of the map signature present... 

(suggestion not requirement)

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -453,9 +477,51 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		workspace.Spec.Paused = new(true)
+		if err := r.Patch(ctx, workspace, client.MergeFrom(originalWorkspace)); err != nil {
+			if apierrors.IsConflict(err) {
+				log.V(2).Info("update conflict while pausing Workspace, will requeue")
+				return ctrl.Result{Requeue: true}, nil
+			}
+			log.Error(err, "unable to pause Workspace")
+			return ctrl.Result{}, err
+		}
+		log.V(1).Info("Workspace paused due to inactivity")
```
andyatmiami, line 597: just a general comment - we only seem to use `V(0)` or `V(2)` right now - but its not clear to me if we have a real "policy" here...

`V(1)` usage seems reasonable (to me) as this is probably more important information as compared to `V(2)` examples like "Statefulset created"

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -453,9 +477,51 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 }
 
+// mergeReconcileResult combines two reconcile results, preferring the sooner requeue.
+func mergeReconcileResult(a, b ctrl.Result) ctrl.Result {
+	if a.Requeue {
+		return a
+	}
+	if b.Requeue {
+		return b
+	}
```
andyatmiami: these conditions could also just be embedded in the below switch, yes? trivial point - take it or leave it!

workspaces/controller/internal/helper/activity.go
```diff
@@ -0,0 +1,230 @@
...
+		return cached.sel.Matches(labels.Set(lbls)), nil
+	}
+
+	sel, err := metav1.LabelSelectorAsSelector(selector)
+	selectorCache.Store(key, cachedSelector{sel: sel, err: err})
+	if err != nil {
+		return false, err
+	}
+	return sel.Matches(labels.Set(lbls)), nil
+}
```
andyatmiami, line 173: suggested (not required) refactor:

```
func selectorMatches(selector *metav1.LabelSelector, lbls map[string]string) (bool, error) {
      if selector == nil {
          return true, nil
      }

      key := metav1.FormatLabelSelector(selector)
      cached, ok := selectorCache.Load(key)

      var sel labels.Selector
      var err error
      if ok {
          c := cached.(cachedSelector)
          sel, err = c.sel, c.err
      } else {
          sel, err = metav1.LabelSelectorAsSelector(selector)
          selectorCache.Store(key, cachedSelector{sel: sel, err: err})
      }

      if err != nil {
          return false, err
      }
      return sel.Matches(labels.Set(lbls)), nil
  }
```

- i think it reads more naturally to not have 2 different `.Matches` functions getting invoked...

workspaces/controller/internal/helper/probe.go
```diff
@@ -0,0 +1,450 @@
...
+			return &endTime, nil
+		}
+		// explicitly inactive
+		if out.LastActivity != nil {
+			t, err := parseISO8601(*out.LastActivity)
+			if err != nil {
+				return nil, fmt.Errorf("invalid JSON file: %w", err)
+			}
+			return &t, nil
+		}
```
andyatmiami: This embedded logic block here seems to go against the [inline comment](https://github.com/kubeflow/notebooks/pull/1279/changes#diff-de69604659158d1c282f03d2d381d676abf7f3a652be6a08be158c641b7aeb03R366) as well as the design doc:

Image

Did you come across a reason/use case where we'd want to implement this as presented in the code?

andyatmiami (changes requested):
@siyuanfoundation 

did another deep drive through the code - flagged a few more things but I feel this is the last round of "intensive review" we are going to need.

local testing all checks out here - so next round of updates will go much quicker as essentially i'll just verify the delta and make sure my test cases continue to pass.

as always - please reach out if you want to discuss any of these items.. i tried to clearly call out "optional/suggested" comments from ones that are more _preferred_

workspaces/controller/internal/controller/workspace_activity.go
siyuanfoundation: monved init to main.go

workspaces/controller/internal/controller/workspace_activity.go
siyuanfoundation: refactored

workspaces/controller/internal/controller/workspace_activity.go
siyuanfoundation: refactored to make `RunJupyterProbe` consistent with `RunPodExecProbe`

workspaces/controller/internal/controller/workspace_controller.go
siyuanfoundation: maybe we should have a guideline around logs

workspaces/controller/internal/helper/probe.go
siyuanfoundation: The design doc is not very clear about what to choose when both `has_activity` and `last_activity` are present. This is what we have merged in the CRD: https://github.com/kubeflow/notebooks/blob/8c4881424be902164866a4d2076c46be4321c3e2/workspaces/controller/api/v1beta1/workspacekind_types.go#L388-L391

My thinking is:
1. if `has_activity=true`, it means there is activity now, so we ignore  `last_activity`
2. it also means all the other conditions should mean `has_activity=false`, whether it is explicit in the json file or not. 
3. when only `last_activity` is present, it is implied `has_activity=false`. So I think it is sensible to to report `has_activity=false` + `last_activity` and expect the same behavior. 

/cc @thesuperzapper for your opinions

workspaces/controller/api/v1beta1/workspace_types.go
```diff
@@ -226,6 +226,8 @@ type WorkspaceStatus struct {
 type WorkspaceActivity struct {
 	// the last time activity was observed on the Workspace (UNIX epoch in milliseconds)
 	//  - this is the value returned by the activity probe, not the time the probe was run
+	//  - the activity probe is AUTHORITATIVE: on success, the probe result unconditionally
```
christian-heusel: Maybe I'm missing something, but I wouldn't caps lock this 🤔 

```suggestion
	//  - the activity probe is authorative: on success, the probe result unconditionally
```

christian-heusel (approved):
Very nice work @siyuanfoundation 🚀

workspaces/controller/internal/helper/probe.go
andyatmiami: discussed this in slack and agreed to pursue an implementation whereby `last_activity` and `has_activity` are more or less **mutually exclusive** 
- users should be encouraged through documentation, etc **not to set both**
- if both are set  - `has_activity` is honored (and `last_activity` is ignored)

andyatmiami (approved):
/lgtm
/approve

---

a few months ago - I had **no idea** if/when we'd have time to fully implement the activity probe/rules feature as its very complex...  you joining the community and the quality work you have delivered @siyuanfoundation has been just amazing.

so happy to see this PR land which brings about a full end-to-end "story" around activity rules (which is a super set of the legacy "culling" behavior).

excellent work 🚀 

testing methodology:
- [PR-1279-TEST-RESULTS.md](https://github.com/user-attachments/files/31235560/PR-1279-TEST-RESULTS.md)
    - agent executed, human reviewed/approved

---

ℹ️ for transparency - i did find one (unlikely) edge case we will address in a follow up PR - but its not worth holding up all this other excellent work from getting merged.
    - if a workspace only/ever reports `has_activity: false` - `eligibleAfter` will never be computed, and as such - no activity rule effects will ever fire...

PR #1280, opened 2026-07-28
chore: prepare frontend for npm v12's stricter install-script defaults

christian-heusel:
npm v12 blocks preinstall/install/postinstall scripts from dependencies unless explicitly approved via `npm approve-scripts`.

Therefore we add a version-pinned `allowScripts` allowlist for the packages that genuinely need their install scripts, and explicitly deny the rest after verifying via clean install + full test/build suite that their scripts can be omitted here.

Also raise `engines.npm` to `^11.16.0`, the version where `allowScripts` support landed, and pin CI/Docker to that version explicitly since neither `actions/setup-node` nor the `node:24-slim` base image otherwise enforce it.

### DIFF
--- a/workspaces/frontend/package.json
+++ b/workspaces/frontend/package.json
@@ -16,7 +16,7 @@
     }
   },
   "engines": {
-    "npm": "^11.10.0",
+    "npm": "^11.16.0",
     "node": "^24"
   },
   "scripts": {
@@ -188,5 +188,14 @@
     "nyc": "^17.1.0",
     "serve": "^14.2.4",
     "ts-jest": "^29.4.9"
+  },
+  "allowScripts": {
+    "@swc/core@1.13.3": true,
+    "cypress@14.5.4": true,
+    "unrs-resolver@1.11.1": true,
+    "@parcel/watcher": false,
+    "core-js": false,
+    "core-js-pure": false,
+    "esbuild": false
   }
 }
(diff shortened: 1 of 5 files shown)


workspaces/frontend/package.json
```diff
@@ -188,5 +188,14 @@
     "nyc": "^17.1.0",
     "serve": "^14.2.4",
     "ts-jest": "^29.4.9"
+  },
+  "allowScripts": {
```
paulovmr, line 192: The `allowScripts` entries for allowed packages use exact version pins (`@swc/core@1.13.3`, `cypress@14.5.4`, `unrs-resolver@1.11.1`). When any of these dependencies are upgraded, the corresponding `allowScripts` entry must also be updated — otherwise `npm ci` will silently skip the install script for the new version, potentially leaving it without native binaries (e.g., @swc/core). This is npm's intended security behavior, but it could cause confusing runtime failures during routine upgrades. Consider adding a brief inline comment near `allowScripts` to remind future contributors of this coupling, or explore whether unversioned entries with `true` would be acceptable for your threat model.

christian-heusel: Good point, this is especially annoying when thinking about dependabot-based updates (which I think can't be tricked into updating both references) 🤔

So how should we go about this?
Considering the options, I think we should just blanket-allow on a per-package basis (and leave the version out of the loop here) and make a security review part of the package update review 😅 This way we still reduce the attack surface to a far more limited amount of packages but don't break our dependency update workflow or introduce unneccessary maintainer churn 😅

christian-heusel: Applied in the latest revision 🤗

Issue #1103, opened 2026-05-19
[TASK] Grey out label filters with no matching results in workspace create/update option selection


paulovmr:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

In the workspace create/update flow, the label filter buttons should reflect the currently visible options. Specifically:

  - Label filter buttons that would yield zero results should be greyed out (disabled) rather than hidden, so users understand the filter exists but has no matches.
  - The available filter values should respect the current "show hidden" and "show redirected" toggle states — if hidden options are not shown, label values that only exist on hidden options should be greyed out.

### Acceptance Criteria

- [ ] Filter buttons with no matching results (given current visibility toggles) are greyed out / disabled
- [ ] Filter state updates dynamically as all filters toggles change

PR #1284, opened 2026-07-29
feat: grey out label filter values with no matching visible results

thaorell:
closes: #1103

### Grey out filter labels that result in no matched results
- Introduce availableLabelValues map with key --> label name and value --> matching results
- Introduce a cleanup useEffect to sync selectedLabels with availableLabelValues whenever the available set changes.
- If a user selects hidden and/or redirected filter, the available options should respect the user's selection
- Order filters so that have available results first
- Fix tests

### DIFF
--- a/workspaces/frontend/src/app/app.css
+++ b/workspaces/frontend/src/app/app.css
@@ -55,6 +55,11 @@
   min-width: 200px;
 }
 
+.filter-label--disabled {
+  opacity: 0.5;
+  pointer-events: none;
+}
+
 .workspace-form__info-icon {
   color: var(--pf-t--global--icon--color--status--info--default);
 }
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/labelFilter/FilterByLabels.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/labelFilter/FilterByLabels.tsx
@@ -28,6 +28,19 @@ type FilterByLabelsProps<T> = {
   filterControlRef?: React.Ref<FilterControlHandle>;
 };
 
+const buildLabelMap = <T,>(objects: LabelledObject<T>[]): Map<string, Set<string>> => {
+  const labelsMap = new Map<string, Set<string>>();
+  objects
+    .flatMap((obj) => obj.labels ?? [])
+    .forEach((label) => {
+      if (!labelsMap.has(label.key)) {
+        labelsMap.set(label.key, new Set<string>());
+      }
+      labelsMap.get(label.key)?.add(label.value);
+    });
+  return labelsMap;
+};
+
 export const FilterByLabels = <T,>(props: FilterByLabelsProps<T>): React.ReactElement => {
   const [selectedLabels, setSelectedLabels] = useState<Map<string, Set<string>>>(new Map());
   const [selectedExtraFilters, setSelectedExtraFilters] = useState<Map<string, ExtraFilter<T>>>(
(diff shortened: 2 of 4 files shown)


workspaces/frontend/src/app/pages/Workspaces/Form/labelFilter/FilterByLabels.tsx
```diff
@@ -51,6 +51,28 @@ export const FilterByLabels = <T,>(props: FilterByLabelsProps<T>): React.ReactEl
     return labelsMap;
   }, [props.labelledObjects]);
 
+  const availableLabelValues = useMemo(() => {
```
paulovmr: The logic to build a `Map<string, Set<string>>` from labelled objects is nearly identical between `filterMap` (lines 41–52) and the new `availableLabelValues` (lines 54–68). Please consider extracting a shared helper like:

  ```ts
  const buildLabelMap = (objects: LabelledObject<T>[]) => {
    const labelsMap = new Map<string, Set<string>>();
    objects.flatMap((obj) => obj.labels ?? [])
      .forEach((label) => {
        if (!labelsMap.has(label.key)) {
          labelsMap.set(label.key, new Set<string>());
        }
        labelsMap.get(label.key)?.add(label.value);
      });
    return labelsMap;
  };
```

  Then both useMemo hooks become one-liners, and the map-building logic lives in one place.

File: workspaces/frontend/src/app/pages/Workspaces/Form/labelFilter/FilterByLabels.tsx [hunk 2]
```diff
@@ -172,16 +220,30 @@ export const FilterByLabels = <T,>(props: FilterByLabelsProps<T>): React.ReactEl
...
+              return aAvail ? -1 : 1;
+            })
+            .map((labelValue) => {
+              const available = isLabelValueAvailable(label, labelValue);
+              return (
+                <FilterSidePanelCategoryItem
+                  key={`${label}|||${labelValue}`}
+                  className={available ? '' : 'filter-label--disabled'}
+                  data-testid={`label-filter-${label}-${labelValue}`}
+                  checked={isLabelChecked(label, labelValue)}
```
paulovmr: When a selected filter value becomes unavailable (e.g., toggling "show hidden" off), there's one paint frame where the item renders as disabled and checked — before the cleanup useEffect runs and deselects it. You can avoid this flash by guarding the checked prop inline:

  `checked={available && isLabelChecked(label, labelValue)}`

  This makes the visual state immediately consistent, while the effect still cleans up selectedLabels for downstream consumers.

workspaces/frontend/src/app/app.css
```diff
@@ -55,6 +55,11 @@
   min-width: 200px;
 }
 
+.filter-label--disabled {
```
paulovmr, line 58: The `pointer-events: none` in `.filter-label--disabled` effectively prevents mouse interaction, but keyboard users can still Tab to the checkbox and press Space/Enter. Since `FilterSidePanelCategoryItem` is a controlled component and the onClick handler is undefined when disabled, the state won't change — but the element remains focusable and interactive from the browser's perspective (unlike a native disabled attribute). The `aria-disabled="true"` attribute correctly communicates the disabled state to assistive technology, which is the right approach. Just want to flag that if flaky keyboard interaction is observed during testing, adding `tabIndex={available ? undefined : -1}` to the disabled items would remove them from the tab order entirely.

paulovmr (changes requested):

paulovmr (approved):
/lgtm
/approve

PR #1263, opened 2026-07-20
feat: Add filterRules schema to WorkspaceKind CRD

gangli113:
## Description
Adds a new optional spec.filterRules[] field to the WorkspaceKind CRD to store admin-defined
rules that downstream consumers (the backend API server) use to dynamically filter which
WorkspaceKinds, imageConfig values, and podConfig values are visible or allowed in a given
context. The controller does not evaluate these rules — it only persists and validates them.

Schema follows the approved proposal in #683.

## Changes
- add FilterRule types with kubebuilder validation markers
- regenerate CRD manifests and DeepCopy methods
- webhook validation: at-least-one-effect, scope-appropriate match types, well-formed selectors
- unit tests covering all scope types, match combinations, and invalid configurations
- add filterRules examples to the 3 sample WorkspaceKinds

Closes #690

PR #1285, opened 2026-07-30
fix: harden testing/ and developing scripts

andyatmiami:
ℹ️ _NO GH ISSUE_

Ran into a couple minor annoyances during PR reviews that would be nice to fix.

(1) When enabling PROMETHEUS in `tilt` - I got an error (likely a race condition) that `kubectl wait ... pods` failed because the pods hadn't been created yet... waiting on Prometheus CR to be `Available` seems the obvious hardening choice here.

(2) In reviewing https://github.com/kubeflow/notebooks/pull/1263 - i had a typo in the PR "suggestion" I offered.. which lead to an extra whitespace and invalid YAML for our `codeserver` WSK that got carried into the author's fixes... and to make matters worse - all our GHA checks passed...  Added a small/idempotent/trivial addition to `sanity-check.sh` that will reasonably catch regressions in our sample files - which seems aligned with the purpose of `sanity-check.sh`

### DIFF
--- a/.github/workflows/ws-e2e-test.yml
+++ b/.github/workflows/ws-e2e-test.yml
@@ -12,6 +12,7 @@ on:
   pull_request:
     paths:
       - workspaces/**
+      - testing/**
 
 jobs:
   e2e-test:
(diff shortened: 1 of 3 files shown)


christian-heusel (approved):

christian-heusel (approved):
Thanks @andyatmiami for raising these qol fixes! 🚀 

Since we previously agreed that this is in good shape as is with your added changes we can merge this and I'll use it as a first opportunity for the additional permissions granted 😊 

/lgtm
/approve

Edit: Right, this also causes the testing folder so this doesn't work 🤣

thesuperzapper:
Thanks!

/lgtm
/approve

Issue #1286, opened 2026-07-30
Update color scheme for clickable buttons


adrianAzoitei:
### Checks

- [x] I have searched the [existing issues](https://github.com/kubeflow/notebooks/issues).
- [x] My request is related to one of the components in the [`kubeflow/notebooks`](https://github.com/kubeflow/notebooks) repository.

### Motivation

When filling in the Notebook creation web form, the "Launch" button turns from greyed-out when disabled to light-blue when clickable.

Some users get confused by this and assume the form is not yet valid.

<img width="261" height="77" alt="Image" src="https://github.com/user-attachments/assets/cc8f921b-fad5-4b46-b662-160cda8c72f7" />
<img width="215" height="75" alt="Image" src="https://github.com/user-attachments/assets/50477c3e-8d15-4f0e-b437-da16d0872c79" />

### Implementation

Use the already defined `kubeflow-blue-hue-4` color for `mat-raised-button` buttons.

### Are you willing & able to help?

- [x] I am able to submit a PR!
- [x] I can help test the feature!

PR #1287, opened 2026-07-30
chore: update clickable material button color

adrianAzoitei:
closes: https://github.com/kubeflow/notebooks/issues/1286

old:
Screenshot 2026-07-30 at 13 13 37

new:
Screenshot 2026-07-30 at 13 18 14


christian-heusel (changes requested):
Please share reproduction steps on #1286 first, I can not reproduce your issue (`ghcr.io/kubeflow/notebooks/jupyter-web-app:v1.11.0` with the [`kubeflow/community-distribution`](https://github.com/kubeflow/community-distribution/) on `master`) and I'm therefore unsure if your fix is needed:

Image

adrianAzoitei:
@christian-heusel you're right, i was running an older build; could swear i built from 1.11.0 locally and ran that in my kind cluster, but seems I did something wrong and saw the old look. will close both the PR and the issue. thanks for checking!

Issue #857, opened 2026-01-29
[TASK] Enable frontend API regeneration during Tilt live development with uncommitted backend changes


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

The `swagger.version` mechanism prevents the core benefit of Tilt—**live iterative development across components**—from working when backend API changes are introduced.

### Tilt's Purpose

Tilt provides a live-reload development environment where:
- Code changes are synced into running containers
- Components rebuild/restart automatically
- Developers iterate rapidly without manual rebuild cycles

The Kubeflow Notebooks Tiltfile (`developing/Tiltfile`) implements this well for most changes:
- Controller: Full Docker rebuild on changes
- Backend: Full Docker rebuild on changes (regenerates swagger.json via `make swag`)
- Frontend: Live sync of `src/` and `config/` with webpack HMR

### The Problem

When a developer modifies a **backend API handler** during a Tilt session:

1. Backend container rebuilds → new `swagger.json` is generated inside the container
2. Developer wants to update frontend TypeScript types to use the new API
3. Developer runs `npm run generate:api` (locally or needs it in container)
4. **Generation fails** because `swagger.version` points to a git commit hash that doesn't contain the new swagger.json

This breaks the Tilt development loop. The developer cannot:
- Test frontend code against new API endpoints
- Iterate on API design with immediate frontend feedback
- Develop a feature that spans backend and frontend in a single session

### Current Tilt Frontend Configuration

From `developing/Tiltfile` (lines 127-158):
```python
docker_build(
    "ghcr.io/kubeflow/notebooks/workspaces-frontend",
    dockerfile=os.path.join(frontend_dir, "Dockerfile.dev"),
    context=frontend_dir,
    live_update=[
        # Sync source files into the container
        sync(os.path.join(frontend_dir, "src"), "/app/src"),
        sync(os.path.join(frontend_dir, "config"), "/app/config"),

        # If package.json changes, reinstall dependencies
        run(
            "cd /app && npm install",
            trigger=[...],
        ),
    ],
    ...
)
```

The `src/` sync includes `src/generated/`, but there's no mechanism to regenerate these files when backend changes occur.

### Why This Violates Tilt's Core Value

Tilt's documentation emphasizes:
> "Tilt watches your files for edits, automatically builds your container images, and applies any changes to bring your environment up-to-date in
real-time."

[...]

PR #1291, opened 2026-07-31
feat: automate frontend API regeneration during live development

prakharsingh-74:
closes: #857 

### Description

Currently, the `swagger.version` Git-hash mechanism prevents the core benefit of Tilt—live iterative development across backend and frontend components—from working when backend API changes are introduced. Uncommitted backend changes are not captured by the Git history, causing local API client generation (`npm run generate:api`) to fail or fetch outdated schemas.

1. **Automatic Generation in Tilt**:
   - Added a `backend-swagger` `local_resource` in the `Tiltfile` that watches backend Go files and automatically runs `make swag` on the host to update the local `openapi/swagger.json`.
   - Added a `frontend-api-generate` `local_resource` in the `Tiltfile` that watches `openapi/swagger.json` and runs `npm run generate:api` on the host.
   - Updated the resource dependency graph so that the workspace containers wait for their respective generation resources to complete before deploying.

2. **Conditional Local Swagger Generation**:
   - Modified `generate-api.sh` to check for `USE_LOCAL_SWAGGER=true` or `DEV_ENV=tilt`. When active, it directly reads the local `swagger.json` path or fetches it from a running container URL (using Node.js's built-in `fetch` to keep container dependencies lean) rather than querying git history.
   - Added CRLF stripping to ensure the Git-hash check works reliably on Windows developer hosts.
   - Prefixed the `generate:api` npm script command with `bash` for native Windows execution compatibility.

3. **Documentation**:
   - Documented the new automated API client regeneration flow in `DEVELOPMENT_GUIDE.md`.

---

### How to Verify

1. **Verify Local Generation on Host**:
   - Run `$env:USE_LOCAL_SWAGGER="true"; npm run generate:api` (Windows PowerShell) or `USE_LOCAL_SWAGGER=true npm run generate:api` (macOS/Linux/Git Bash) inside `workspaces/frontend`. Verify it generates types using the local `swagger.json`.

2. **Verify Git-Based Generation (CI/Production Mode)**:
   - Run `npm run generate:api` without setting the environment variable. Verify it falls back to the original git-hash-based generation.

3. **Verify Tilt E2E Loop**:
   - Start Tilt via `make tilt-up` in `developing/`.
   - Modify a backend Go struct/handler. Verify that:
     1. Tilt runs `make swag` locally.
     2. Tilt runs `generate-api.sh` to update the TypeScript types in `src/generated/`.
     3. Webpack dev server HMR is triggered inside the frontend container.


prakharsingh-74:
cc: @andyatmiami

prakharsingh-74:
@andreyvelich sir, can you review this PR if you're free.

workspaces/frontend/package.json
```diff
@@ -26,7 +26,7 @@
     "build:bundle-analyze": "webpack-bundle-analyzer ./bundle.stats.json",
     "build:clean": "rimraf ./dist",
     "build:prod": "webpack --config ./config/webpack.prod.js",
-    "generate:api": "./scripts/generate-api.sh && npm run prettier",
+    "generate:api": "bash ./scripts/generate-api.sh && npm run prettier",
```
christian-heusel, line 29: Why do we need to explicitly execute this script with bash? 🤔 
The script already has this as shebang, so my intuition would be that this can be dropped 🤔 
https://github.com/kubeflow/notebooks/blob/07a611c3884ab205544bd3815776ce6399b36608/workspaces/frontend/scripts/generate-api.sh#L1

workspaces/frontend/scripts/generate-api.sh
```diff
@@ -3,17 +3,31 @@ set -euo pipefail
...
 TMP_SWAGGER=".tmp-swagger.json"
 
-if ! git cat-file -e "${SWAGGER_COMMIT_HASH}:${SWAGGER_JSON_PATH}"; then
-  echo "❌ Swagger file not found at commit $SWAGGER_COMMIT_HASH"
-  exit 1
+# Check if we should use local or container swagger file/URL
+if [ "${USE_LOCAL_SWAGGER:-false}" = "true" ] || [ "${DEV_ENV:-}" = "tilt" ]; then
+  echo "Using local/container swagger spec: $SWAGGER_JSON_PATH"
+  if [[ "$SWAGGER_JSON_PATH" =~ ^https?:// ]]; then
+    node -e "fetch('$SWAGGER_JSON_PATH').then(r => { if (!r.ok) throw new Error('status ' + r.status); return r.text(); }).then(t => process.stdout.write(t)).catch(e => { console.error(e); process.exit(1); })" > "$TMP_SWAGGER"
```
christian-heusel, line 14: IMO this is not the right approach to the issue we're fixing, why are we fetching this file from a remote endpoint instead of just using the one on disk or regenerating it in a temporary directory? 🤔

workspaces/frontend/scripts/generate-api.sh
```diff
@@ -3,17 +3,31 @@ set -euo pipefail
 
 GENERATED_DIR="./src/generated"
 HASH_FILE="./scripts/swagger.version"
-SWAGGER_COMMIT_HASH=$(cat "$HASH_FILE")
-SWAGGER_JSON_PATH="../backend/openapi/swagger.json"
+SWAGGER_COMMIT_HASH=$(tr -d '\r\n' < "$HASH_FILE")
```
christian-heusel, line 6: The line ending here is not guaranteed to be `\r\n`, on linux this will just be `\n`. Probably we should also fix this in `.gitattributes` so it is predictable accross both os types.

See https://stackoverflow.com/a/426404

christian-heusel (changes requested):
Thanks a lot for your contribution, I have left a few review comments below 🤗 

A few general notes:

- Did you use AI in any capacity while creating this contribution? I see that your commit does not carry a `Assisted-By: <...>` tag, so I wanted to make sure that you're aware of the [Kubeflow AI Policy](https://www.kubeflow.org/docs/about/ai_policy/). 🤗 🤖 
- Please do not ping external maintainers for your PR, it's okay to make a bit of noise if a PR goes without review, however for example Andrey does not know anything about the codebase here and will not be available to review your changes.
- This is more of a preference thing, but if you rebase your changes against `notebooks-v2` instead of merging this keeps the history more clean.

PR #1305, opened 2026-08-04
feat: Add `schedulerName` to WorkspaceKind CRD & controller reconcile

christian-heusel:
Kubernetes supports running [multiple schedulers](https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/#specify-schedulers-for-pods) in a single cluster, with the scheduler for a given Pod selected via `pod.spec.schedulerName`. Clusters running Kubeflow Notebooks often also run a batch or gang scheduler (Volcano, YuniKorn, scheduler-plugins), and typically only want it for the GPU/large-resource workspaces rather than for every workspace.

Today the `WorkspaceKind` CRD has no way to express this at all, so every Workspace Pod is necessarily handled by the cluster's default scheduler.

## Changes

This PR adds an optional `schedulerName` configuration at two levels:

- **`spec.podTemplate.schedulerName`** ➡️ the default for all Workspaces of the WorkspaceKind
- **`spec.podTemplate.options.podConfig.values[].spec.schedulerName`** ➡️ per pod config, taking precedence over the pod template

```yaml
spec:
  podTemplate:
    schedulerName: "default-scheduler"   # default for this WorkspaceKind
    options:
      podConfig:
        values:
          - id: "tiny_cpu"
            spec: {}                     # inherits "default-scheduler"
          - id: "big_gpu"
            spec:
              schedulerName: "volcano"   # wins for this pod config
```

## Behaviour notes

- When the field is unset at both levels it is omitted, and the Kubernetes API server applies its usual `default-scheduler`, exactly as before. Existing WorkspaceKinds are unaffected.
- Both fields are validated as DNS-1123 subdomains, reusing the markers already applied to `spec.podTemplate.serviceAccount.name`.
- Changing `schedulerName` on the **pod template** restarts all Workspaces of that kind, consistent with the other MUTABLE pod template fields.
- Changing it on an **in-use pod config** is rejected by the existing WorkspaceKind validating webhook, which already compares the whole `PodConfigSpec`. Admins add a new pod config value and a redirect instead, as they would for a resource change.
- No change was needed in `helper.CopyStatefulSetFields()`, which already compares the whole pod template with `equality.Semantic.DeepEqual`, so changing the scheduler rolls the StatefulSet as expected.

One caveat worth being aware of when using this: nothing validates a `schedulerName` server-side. A Pod requesting a scheduler that is not running in the cluster stays `Pending` indefinitely, with no events explaining why. The sample calls this out.

## Out of scope

- **Frontend.** No changes are required for correctness: the WorkspaceKind admin form already preserves unmapped fields round-trip through its existing spreads, and the YAML editor's schema does not reject unknown properties. Regenerating the client and adding dedicated form inputs can follow separately. Note that `scripts/swagger.version` is intentionally left pointing at the previous commit, so `src/generated/` stays self-consistent.
- **A `Workspace`-level override.** Scheduler choice is treated as an admin concern here; users select a pod config, which carries the scheduler.


workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -295,6 +295,18 @@ type WorkspaceKindPodTemplate struct {
...
 
+	// the name of the scheduler to use for Workspace Pods (MUTABLE)
+	//  - this is the default for all Workspaces of this WorkspaceKind, it may be
+	//    overridden by the `schedulerName` of a pod config value
+	//  - if not set here, or on the pod config value, the Kubernetes API server
+	//    will default to the "default-scheduler"
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinLength:=1
+	// +kubebuilder:validation:MaxLength:=253
+	// +kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
```
andyatmiami: Best I can tell...Kubernetes applies zero validation to `PodSpec.SchedulerName`. The upstream source shows:
- https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/v1/zz_generated.validations.go#L1310

This means the PR's DNS subdomain pattern validation is stricter than what Kubernetes itself enforces. 
As such, the validation is too strict. Kubernetes accepts any string, including uppercase, underscores, etc.

workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -638,6 +650,17 @@ type PodConfigSpec struct {
...
+	// the name of the scheduler to use for the pod
+	//  - this takes precedence over the `schedulerName` of the pod template
+	//  - set this to "default-scheduler" to have this pod config use the
+	//    default Kubernetes scheduler, even if the pod template sets another one
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinLength:=1
+	// +kubebuilder:validation:MaxLength:=253
+	// +kubebuilder:validation:Pattern:=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	// +kubebuilder:example="volcano"
+	SchedulerName *string `json:"schedulerName,omitempty"`
```
andyatmiami, line 677: Best I can tell...Kubernetes applies zero validation to `PodSpec.SchedulerName`. The upstream source shows:
- https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/v1/zz_generated.validations.go#L1310

This means the PR's DNS subdomain pattern validation is stricter than what Kubernetes itself enforces. 
As such, the validation is too strict. Kubernetes accepts any string, including uppercase, underscores, etc.

andyatmiami (changes requested):
Small nit on what I think is over-validation on the `schedulerName` field..

Please let me know if you have evidence to the contrary that the regex you are applying is aligned with k8s!

Everything else looks good - so once we resolve this - we can get this PR into the codebase 💯

workspaces/controller/api/v1beta1/workspacekind_types.go
christian-heusel: So what is our preference here? 🤔 Should I just strip this validation fully or do we just drop the regex?

thesuperzapper: If upstream has literally no validation, we don't either, I am not sure it even has a max length.

We could do a quick check, or just remove the validation.

christian-heusel: Lets just follow upstream then and see if it explodes at some point 😊

christian-heusel:
I just rebased this against `notebooks-v2` again and adressed your feedback 🤗 Any more work needed or are we good to go here?

andyatmiami (approved):
/lgtm
/approve

---

thanks @christian-heusel - code is consistent with existing conventions and appropriate to accomplish the intended outcome.

---

testing methodology:
- [PR-1305-TEST-RESULTS.md](https://github.com/user-attachments/files/31795787/PR-1305-TEST-RESULTS.md)
    - agent-driven, human reviewed/approved

PR #1301, opened 2026-08-04
[WIP] feat: implement Gateway API routing provider + standalone notebooks

aojea:
This commit introduces Gateway API (v1) support as an alternative to Istio for routing traffic to workspaces.

Key changes:
- workspaces/controller: Added a gateway-api routing provider in config.RoutingProviderType. When enabled (ROUTING_PROVIDER=gateway-api), the workspace controller generates HTTPRoute resources instead of Istio VirtualService resources.
- RBAC: Added permissions to the controller to manage gateway.networking.k8s.io/httproutes.
- Kustomize Overlays: Added gateway-api kustomize components and overlays for backend, frontend, and controller so the cluster can be stood up independently of Istio.
- E2E Testing: Integrated a test-e2e-gateway-api Make target that uses Envoy Gateway, standard Gateway API v1.1.0 CRDs, and cloud-provider-kind to fully test end-to-end functionality via Kubernetes HTTPRoutes.

This allows Kubeflow users to adopt any Gateway API-compliant ingress controller (like Envoy Gateway) instead of being strictly coupled to Istio.

Assisted by AI

PR #1306, opened 2026-08-05
chore: Bump golang dependencies in backend & controller

christian-heusel:
## Notable changes

`workspaces/backend/internal/auth/authentication.go`:

- `headerrequest.New()` gained a new `uidHeaders []string` parameter upstream
  in `k8s.io/apiserver`. Updated the call site to pass `nil` for it since it's
  unused (auth is keyed on name/group headers only).

---

`workspaces/controller/internal/webhook/workspace_webhook.go` / `workspacekind_webhook.go`

`controller-runtime` `v0.24` made the webhook builder generic:

- `ctrl.NewWebhookManagedBy(mgr).For(&T{})` → `ctrl.NewWebhookManagedBy(mgr, &T{})`
- `admission.Validator` methods (ValidateCreate/Update/Delete) now take the concrete type (`*Workspace`, `*WorkspaceKind`) directly instead of `runtime.Object`, so the manual `obj.(*T)` type assertions + `apierrors.NewBadRequest` fallback were dropped — that check is now enforced by the compiler instead of at request time (type mismatches are caught by the compiler; runtime failures are now JSON-decode errors, not type-assertion errors — see `sigs.k8s.io/controller-runtime/pkg/webhook/admission/validator_custom.go`).

---

`workspaces/controller/internal/helper/probe.go`:
- Switch import `k8s.io/apimachinery/pkg/util/httpstream` → `k8s.io/streaming/pkg/httpstream`

---

`workspaces/controller/internal/controller/workspace_activity_test.go` & 
- nolint the `Result.Requeue` deprecation (will open another task to refactor this deprecated type out)


christian-heusel:
/cc @aojea
who also did a few dependency updates in https://github.com/kubeflow/notebooks/pull/1301

andyatmiami (changes requested):
thanks @christian-heusel - I have one suggestion here from reviewing the various change logs across these version bumps...

---

### controller-runtime v0.21: client-side rate limiter removed

Starting with controller-runtime v0.21, the default client-side rate limiter (QPS=20, Burst=30) was [disabled by default](https://github.com/kubernetes-sigs/controller-runtime/pull/3119) — the rest config now ships with no client-side rate limiting.

**Backend**: Already mitigated. `cmd/main.go` explicitly sets `kubeconfig.QPS` and `kubeconfig.Burst` from env vars (defaults: QPS=50, Burst=100).

**Controller**: Uses `ctrl.GetConfigOrDie()` at `cmd/main.go:123` without setting explicit QPS/Burst on the rest config. After this bump, the controller will have **no client-side rate limiting** on API calls. The API server's server-side priority & fairness provides some protection, but under heavy load (many workspaces reconciling simultaneously) the controller may generate more concurrent API requests than before.

Purely out of paranoia - seems like maybe we should configure the controller similarly to the backend - but note as I say this - I'm not fully appreciative of the ramifications here... feel free to push back

> Note: The `RateLimiter` at `cmd/main.go:176/191` is the **workqueue** rate limiter (controls reconcile requeue timing), not the Kubernetes API client rate limiter.

aojea:
> Purely out of paranoia - seems like maybe we should configure the controller similarly to the backend - but note as I say this - I'm not fully appreciative of the ramifications here... feel free to push back

good catch , I think the importance of this parameters is on the number of write requests the controller does, how many objects it updates and creates , so having client side enable always help to avoid storming the apiserver and being consistent across the project seems a good call

andyatmiami (approved):
/lgtm
/approve

---

thanks @christian-heusel - really appreciate you always pushing and staying on top of _the little details_ like dependency versions to make Kubeflow Workspaces **awesome and enterprise ready**.

from reviewing the change logs for the dependencies being updated - and the minimal changes you delivered to account for these dep upgrades - I think this is looking good and ready to be merged...

timing wise - as we just shipping the beta - this is also the best time to do work like this.. in the event we missed some weird/subtle thing - we'll have plenty of time to find it (or let community find it) - and be better position for our GA near EOY!

---

testing methodology:
- [PR-1306-TEST-RESULTS.md](https://github.com/user-attachments/files/31354115/PR-1306-TEST-RESULTS.md)

Issue #1257, opened 2026-07-17
[FEATURE] Per-Workspace ServiceAccount and RBAC Role Binding


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Motivation

Today, all Workspaces of a given WorkspaceKind share a single static ServiceAccount (e.g. `default-editor`), making it impossible for Joel to grant per-workspace RBAC permissions. This is a fundamental limitation: if Bella needs her Workspace to access a namespace-scoped Role — for example, one generated by a Kubeflow TrainJob that grants `kubectl exec` to the head node of an MPI mesh — Joel has no way to bind that Role to just her Workspace without affecting every other Workspace using the same WorkspaceKind.

As established in #1124, Joel needs the controller to handle resource management correctly and validate inputs thoroughly. Per-workspace ServiceAccount management is a prerequisite for fine-grained RBAC, and it must land before GA to avoid breaking changes to the CRD schema. The controller should manage the ServiceAccount lifecycle the same way it already manages other per-Workspace resources like Service and VirtualService — generating a unique ServiceAccount for each Workspace and reconciling it alongside the StatefulSet.

With per-workspace ServiceAccounts in place, the CRD can expose fields that let Joel define cluster-wide baseline permissions at the WorkspaceKind level (via `ClusterRole` references) and let Bella bind namespace-scoped Roles to her individual Workspace. To prevent privilege escalation, the Workspace webhook must
perform a SubjectAccessReview to confirm the caller has permission to create/delete RoleBindings before allowing changes to the roles list.

#### What This Feature Does NOT Provide

- UI or backend API endpoints for browsing/selecting Roles (future work)
- TrainJob lifecycle management from the Workspaces UI (future work)
- A `ResourceProviderConnection` CRD to bundle related resources like Roles and ConfigMaps (future work)

#### CRD Schema Changes

**WorkspaceKind** — new field at `spec.podTemplate.serviceAccount`:

```yaml
spec:
  podTemplate:
    serviceAccount:
      ## ClusterRoles to bind to the Workspace Pod's ServiceAccount.
      ## Bound via a namespace-scoped RoleBinding, so access is limited
      ## to the Workspace's namespace (not cluster-wide).
      clusterRoles:
        - name: "kubeflow-view"
```

**Workspace** — new field at spec.podTemplate.serviceAccount:

```yaml
spec:
  podTemplate:
    serviceAccount:
[...]

PR #1307, opened 2026-08-05
chore: remove serviceAccount.name from WorkspaceKind CRD

christian-heusel:
## Description

Removes `spec.podTemplate.serviceAccount.name` from the WorkspaceKind CRD and hardcodes the ServiceAccount name in the controller instead.

The implementation of per-Workspace ServiceAccounts (#1257) will have the controller create and own a unique ServiceAccount for every Workspace, and repurpose `spec.podTemplate.serviceAccount` to
carry `clusterRoles`. Given the new design a user-supplied `name` is meaningless therefore it would have to be deprecated, ignored, or removed from the CRD *after* the beta release, which would be backwards-incompatible CRD change at a point where we no longer want to make them.

Instead we'll break this now so that #1257 only has to *add* fields to `serviceAccount` later, which is purely additive and hence non-breaking.

## Changes

- drop `name` from `WorkspaceKindServiceAccount`, leaving the struct empty and the `serviceAccount`
  field optional; the immutability CEL rule and validation markers go with it
- add the `workspaceServiceAccountName` constant (`"default-editor"`) to the controller and use it
  when generating the Workspace StatefulSet pod spec
- assert the generated StatefulSet's `ServiceAccountName` in the Workspace controller tests
- drop the `serviceAccount` blocks from the 3 sample WorkspaceKinds, and document in
  `samples/common/workspace_service_account.yaml` that its name is fixed for the time being
- regenerate the CRD, DeepCopy methods, and the backend OpenAPI spec

## Breaking change

WorkspaceKinds that set `spec.podTemplate.serviceAccount.name` are now **rejected** by the API
server, the two lines have to be removed from them:

## Follow-up

The generated frontend client still declares the field. `workspaces/frontend/scripts/swagger.version` pins a merged commit hash, so it can only be bumped once this PR is on `notebooks-v2`.
A follow-up PR will need to bump it, run `npm run generate:api`, and drop `serviceAccount` from the two frontend mocks.
Nothing should break in the meantime since the frontend never had a form field for `serviceAccount`, and `Form/helpers.ts` never set it on the create path.

Relates to #1257

### DIFF
--- a/workspaces/backend/api/suite_test.go
+++ b/workspaces/backend/api/suite_test.go
@@ -224,9 +224,6 @@ func NewExampleWorkspaceKind(name string) *kubefloworgv1beta1.WorkspaceKind {
 			},
 			PodTemplate: kubefloworgv1beta1.WorkspaceKindPodTemplate{
 				PodMetadata: &kubefloworgv1beta1.WorkspaceKindPodMetadata{},
-				ServiceAccount: kubefloworgv1beta1.WorkspaceKindServiceAccount{
-					Name: "default-editor",
-				},
 				ActivityProbe: &kubefloworgv1beta1.ActivityProbe{
 					MinProbeIntervalSeconds: new(int32(300)),
 					ProbeIntervalSeconds:    new(int32(3600)),
--- a/workspaces/backend/api/workspacekinds_handler_test.go
+++ b/workspaces/backend/api/workspacekinds_handler_test.go
@@ -356,8 +356,6 @@ spec:
     logo:
       url: "https://upload.wikimedia.org/wikipedia/commons/3/38/Jupyter_logo.svg"
   podTemplate:
-    serviceAccount:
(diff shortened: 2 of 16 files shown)


christian-heusel:
cc @jefftree regarding the related issue

Jefftree:
Looks good, aligns with what I had in mind with #1257 and ill pull this to rebase on top of my changes.

andyatmiami (approved):
/lgtm
/approve

---

thanks @christian-heusel for a clean/straightforward  "prep" PR to clear way for @Jefftree 's follow up work.

i wish all PRs were as conceptually straightforward as this 😄 

test plan:
- [PR-1307-TEST-RESULTS.md](https://github.com/user-attachments/files/30804364/PR-1307-TEST-RESULTS.md)

also did simple "clicks" tests through frontend in Tilt and make sure all samples imported clean

Image

Issue #1289, opened 2026-07-30
[TASK] Implement Simple Log Viewer in UI


thaorell:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description
Introduce a PatternFly Log Viewer (@patternfly/react-log-viewer) in the Workspace details overlay panel to allow users to view container logs for their Workspaces directly from the UI, without requiring kubectl access.

This is a frontend-only task. It consumes the backend GET /api/v1/workspaces/{namespace}/{name}/podtemplate/logs/batch endpoint (see #885, #886) and provides a read-only, non-streaming log viewing experience. Live streaming (/logs/stream) is out of scope.

### Context
  - The Workspace details panel (WorkspaceDetails.tsx) already has a commented-out Logs tab placeholder (line 81).
  - The backend logs batch endpoint returns log lines as a JSON array via the standard Envelope[D] pattern.
  - The /podtemplate/details endpoint (#1029) provides the list of available containers for the container selector.
  - PatternFly provides @patternfly/react-log-viewer (v6.x, MIT licensed) with built-in search, line wrapping, and scroll-to-bottom.\
 
###  High-Level Design

  Logs Tab (in Workspace details panel)
  1. Uncomment and wire up the existing Logs tab in WorkspaceDetails.tsx.
  2. Create a LogsTab component that:
    - Fetches logs from GET /podtemplate/logs/batch on mount and via a manual refresh button.
    - Renders log output using <LogViewer> from @patternfly/react-log-viewer.
    - Provides a container selector dropdown populated from the /podtemplate/details response (defaults to the first container).
    - Provides a "Previous logs" toggle (?previous=true) for viewing terminated container logs.
    - Provides a tail lines selector (e.g., 100, 500, 1000, all) via ?tail=N.
    - Includes a refresh button to re-fetch logs on demand.
    - Uses LogViewerSearch for in-viewer text search.

### Acceptance Criteria

  - [ ] @patternfly/react-log-viewer is added as a dependency
  - [ ] The Logs tab is enabled in the Workspace details panel (uncommented and functional)
  - [ ] Logs are fetched from GET /api/v1/workspaces/{namespace}/{name}/podtemplate/logs/batch and rendered in a <LogViewer> component
  - [ ] A container selector dropdown allows switching between containers (populated from /podtemplate/details pod containers list); defaults to the first container
[...]

PR #1269, opened 2026-07-22
feat: add gzip/zstd response compression to backend

christian-heusel:
## Summary
- Wraps the backend HTTP handler with `github.com/klauspost/compress/gzhttp`, negotiating gzip or zstd per-request based on the client's `Accept-Encoding` header.
- Uses library defaults: 1024-byte minimum response size before compressing (avoids making tiny responses larger from framing overhead), zstd preferred over gzip when a client accepts both at equal quality.

This will be especially useful when we proceed with the work in #885, as this way the payload size towards the client can be drastically reduced.

## Impact / testing results

Benchmarked every non-mutating endpoint via repeated requests with `Accept-Encoding: identity|gzip|zstd`:

| Endpoint | Uncompressed | gzip | zstd |
|---|---:|---:|---:|
| `GET /workspacekinds` | 7,143 B | 1,238 B (-82.7%) | 1,332 B (-81.3%) |
| `GET /workspacekinds/{name}` | 5,108 B | 1,517 B (-70.3%) | 1,653 B (-67.6%) |
| `GET /persistentvolumeclaims/{ns}` | 5,488 B | 802 B (-85.4%) | 792 B (-85.6%) |
| `POST /workspacekinds/{name}/podtemplate/options/listvalues` | 3,149 B | 752 B (-76.1%) | 805 B (-74.4%) |
| `GET /workspaces` | 1,322 B | 670 B (-49.3%) | 713 B (-46.1%) |

Small responses (healthcheck, namespaces, secrets, single-workspace detail which are all under ~450 B) correctly stayed uncompressed on every variant, confirming the 1024-byte threshold works as intended.

**Load test** ([`hatoo/oha`](https://github.com/hatoo/oha)):

```
$ oha -z 5sec --insecure https://localhost:8443/workspaces/api/v1/workspacekinds\?namespaceFilter\=default -H "kubeflow-userid: admin" [--disable-compression] --no-tui
```

| Variant | Req/sec | Mean latency | p99 latency |
|---|---:|---:|---:|
| Uncompressed | 1,234.6 | 40.60 ms | 110.71 ms |
| gzip | 1,449.5 (+17.4%) | 34.64 ms (-14.7%) | 53.62 ms (-51.6%) |
| zstd | 1,396.4 (+13.1%) | 35.92 ms (-11.5%) | 69.23 ms (-37.5%) |

Under concurrency, compression measurably reduces contention for the shared network/tunnel bandwidth. Single uncontended requests showed no latency difference either way, as expected.

[...]

PR #1310, opened 2026-08-06
feat: viewing logs in frontend details view

christian-heusel:
Add a Logs tab to the workspace details drawer, so users can read container logs from the UI without `kubectl` access. The tab consumes the batch endpoint added in #1267 and is read-only, live streaming (`/logs/stream`) stays out of scope.

closes: https://github.com/kubeflow/notebooks/issues/1289
related: https://github.com/kubeflow/notebooks/issues/885, https://github.com/kubeflow/notebooks/issues/886

> [!IMPORTANT]
> **Stacked on #1267 and #1269**: this branch contains that PR's backend commits. It should be rebased onto `notebooks-v2` once they are merged, and `scripts/swagger.version` re-pinned to the resulting merge commit (see below). Only the two frontend commits are new here.

- [x] #1269 rebased and dropped
- [x] #1267 rebased
- [x] #1314 rebased and dropped

### Commit 1: Add the Logs tab

`feat(frontend): add a Logs tab to the workspace details drawer`

- Enables the placeholder `<Tab>` in `WorkspaceDetails.tsx` (its `eventKey` collided with the Resources tab, and is now `3`, matching the existing `logsTabContent`)
- New `useWorkspaceLogs` hook, following the `useFetchState` pattern of `useWorkspaceDetails`
- New `WorkspaceDetailsLogs` component built on `@patternfly/react-log-viewer` (new dependency, v6, MIT)

Controls:

| Control | Query parameter | Notes |
| --- | --- | --- |
| Container | `container` | Populated from `details.pod.containers` + `initContainers`, so `istio-proxy` is reachable; defaults to the primary container |
| Lines | `tailLines` | 100 / 500 / 1000 / 5000. The backend validates the value as a *positive* integer, so there is no "all lines" option |
| Time range | `sinceTime` | All time / 15 minutes / 1 hour / 24 hours, sent as RFC3339 |
| Previous container | `previous` | For crash-loop debugging |
| Search | | `LogViewerSearch`, scoped to the fetched snapshot |
| Wrap lines, refresh, download |  | Download saves `<workspace>-<container>.log` |

States: a spinner while loading, an empty state when the workspace has no pod (with separate wording for paused workspaces) or when the container produced no output, and an error state that surfaces the backend's message, e.g. the 409 for a pod that is not running.

### Implementation notes for reviewers

- **The response is a raw `text/plain` stream**, so the hook does not unwrap an envelope. The request format is deliberately left as JSON rather than `text`: error responses *are* JSON envelopes, and axios hands back the raw string whenever a body is not valid JSON. With `format: 'text'` the envelope stays an unparsed string and every failure degrades to a generic status message.
- **The log viewer measures itself on mount**, so it is mounted only while its tab is active. All `TabContent` bodies in the drawer render eagerly, and mounting the viewer hidden makes it size itself to a zero-width container and render nothing.
- **The toolbar is a plain wrapping flex row, not a PatternFly `Toolbar`.** A `Toolbar`'s content row is a flex item with `min-width: auto`, so in the narrow, resizable drawer it sizes to its content and pushes half the controls outside the panel instead of wrapping. Note that `ToolbarContent`'s `style` lands on a wrapper rather than the row that overflows, so constraining it from the outside does not help.
- The line-number gutter is turned off since it costs more width than it is worth at the current drawer size. In the future we should make this something else like a popover modal I guess.

### Testing

- `useWorkspaceLogs` unit tests: API unavailable, not-ready states, raw-string response, query parameter forwarding
- `WorkspaceDetailsLogs` unit tests: loading, error, empty (no pod / paused / no output) and populated states, container defaulting, previous-container refetch, manual refresh
- Cypress: opening the Logs tab renders the log text and requests the primary container; a 409 renders the error state
- `npm run test` (prettier, lint, type-check, jest, Cypress) passes
- Verified manually in the browser against mock data: container switching (including the init container), line wrapping, refresh, download, and the toolbar layout at drawer width
    - Also checked that with this setup the compression introduced in https://github.com/kubeflow/notebooks/pull/1269 already works

### Screenshots

grafik

grafik


christian-heusel:
cc @YoyinZyc @andyatmiami regarding the ongoing work in the other PR 🤗

YoyinZyc:
The UI looks pretty cool! Just some general questions:
1. What is the default selection for "Lines" and "Time range"?
2. How is the "search" behavior implemented? Is it all on the frontend?
3. What is going to happen after click "Wrap lines"?

christian-heusel:
Thanks for the positive feedback! 🤗 
Generally you should be able to just checkout the branch and run `make tilt-up` to play around with this live if you wanna check it out a bit more!

1. Lines is 1000 to match the backend default and "Time range" is "All time"
2. Yes, its part of the patternfly library and not a "proper" search function: https://www.patternfly.org/extensions/log-viewer/#with-search
3. It wraps the lines, see the screenshot below

image

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetails.tsx
```diff
@@ -83,14 +84,13 @@ export const WorkspaceDetails: React.FunctionComponent<WorkspaceDetailsProps> =
             aria-label="Resources"
             data-testid="resources-tab"
           />
-          {/* TODO: Uncomment when Logs visualization is fully supported
```
Snehadas2005, line 86: Looking at the diff, it looks like the `<Tab>` for Logs might still be inside the multi-line commented-out block (`{/* TODO: Uncomment when Logs visualization... */}`). It could just be a diff rendering artifact on my end, but could you double-check to make sure the tab is actually uncommented and rendering in the UI as expected?

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
```diff
@@ -0,0 +1,340 @@
...
+  const containerNames = useMemo(
+    () => (details?.pod?.containers ?? []).map((c) => c.name),
+    [details?.pod?.containers],
+  );
+  const initContainerNames = useMemo(
+    () => (details?.pod?.initContainers ?? []).map((c) => c.name),
+    [details?.pod?.initContainers],
+  );
+
+  const [selectedContainer, setSelectedContainer] = useState<string | undefined>();
```
Snehadas2005, line 72: I could see a potential bug here when switching between workspaces while keeping the drawer open on the Logs tab.

Because `WorkspaceDetailsLogs` stays mounted, `selectedContainer` doesn't reset when the `workspace` changes. If a user selects a specific container (like `istio-proxy`) on one workspace and then clicks on another workspace that doesn't have that container, `selectedContainer` will still hold the old name. This ends up sending a request to the backend for a container name that doesn't exist on the new pod, which returns a `400 Bad Request`.

Could we add a quick `useEffect` to reset the selected container (and `scrollToRow`) whenever the workspace changes? Something like:

```
useEffect(() => {
  setSelectedContainer(undefined);
  setScrollToRow(undefined);
}, [workspace.namespace, workspace.name]);
```

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
```diff
@@ -0,0 +1,340 @@
...
+  const [tailLines, setTailLines] = useState<number>(1000);
+  const [sinceLabel, setSinceLabel] = useState<string>(SINCE_OPTIONS[0].label);
+  const [previous, setPrevious] = useState(false);
+  const [isTextWrapped, setIsTextWrapped] = useState(false);
+  const [scrollToRow, setScrollToRow] = useState<number | undefined>();
```
Snehadas2005, line 77: I could see a minor edge case here: `scrollToRow` is not reset when someone switches `container`, `tailLines`, or the time range. If a user scrolls all the way down in a 5,000-line log and then filters down to `tailLines: 100, scrollToRow` holds onto the old out-of-range index.

Not a crash or anything breaking, but resetting `setScrollToRow(undefined)` whenever the filter setters are called would keep the scroll position super clean!

workspaces/frontend/src/app/hooks/useWorkspaceLogs.ts
```diff
@@ -0,0 +1,44 @@
...
+    // The response format is left as JSON so that error responses (which *are* JSON envelopes)
+    // stay parsed; axios hands back the raw string whenever the body is not valid JSON.
+    const logs = await api.workspaces.getWorkspacePodTemplateLogsBatch(namespace, name, {
+      container,
+      tailLines,
+      sinceTime,
+      previous,
+    });
+    // Guard against a log body that happens to be valid JSON and was therefore parsed.
+    return typeof logs === 'string' ? logs : JSON.stringify(logs);
```
Snehadas2005, line 40: Catching the axios auto-parsed JSON edge case here is super sharp!
Just a small heads-up: `JSON.stringify` can be a bit lossy with original formatting/whitespace if the raw log string happens to be pure JSON. Since that is extremely rare in real-world log output, it is probably fine as-is, but I thought I would mention it just as a minor limitation to keep in mind.

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
```diff
@@ -0,0 +1,340 @@
...
+  detailsError,
+}) => {
+  const containerNames = useMemo(
+    () => (details?.pod?.containers ?? []).map((c) => c.name),
+    [details?.pod?.containers],
+  );
+  const initContainerNames = useMemo(
+    () => (details?.pod?.initContainers ?? []).map((c) => c.name),
+    [details?.pod?.initContainers],
+  );
```
Snehadas2005, line 70: Just a quick observation while reading through: `containerOptions` appends `(init)` to the display label, but uses the bare container name as the value.

If a pod ever happened to have an init container and a regular container with the exact same name, selecting between them might get ambiguous. Definitely a super rare edge case in Kubernetes and not a blocker, but wanted to throw it out there!

workspaces/backend/api/workspace_podtemplate_logs_handler.go
```diff
@@ -0,0 +1,155 @@
...
+	opts, valErrs := parseLogOptions(r)
+	if len(valErrs) > 0 {
+		a.failedValidationResponse(w, r, errMsgQueryParamsInvalid, valErrs, nil)
+		return
+	}
+
+	// =========================== AUTH ===========================
+	authPolicies := []*auth.ResourcePolicy{
+		auth.NewResourcePolicy(auth.VerbGet, auth.Workspaces, auth.ResourcePolicyResourceMeta{Namespace: namespace, Name: workspaceName}),
+	}
```
Snehadas2005, line 79: I had a quick architectural question regarding RBAC permissions here.
Right now, log access is gated on `auth.VerbGet` for `auth.Workspaces`, which shares the exact same permission check as workspace metadata endpoints (like `/podtemplate/details`). Upstream Kubernetes explicitly separates `pods` and `pods/log` into distinct RBAC subresources because container logs can often leak sensitive runtime environment variables, tokens, or PII that wouldn't normally appear in spec details.

Collapsing the distinction means anyone who can view workspace details automatically gets full access to read all container logs (including sidecars like `istio-proxy`). This might be completely intentional and consistent with the threat model of this project! Just wanted to check: was this a deliberate design decision, or should it be considered introducing a distinct policy check for log streaming?

workspaces/backend/internal/repositories/logs/repo.go
```diff
@@ -0,0 +1,220 @@
...
+	ErrGetPod               = fmt.Errorf("failed to get pod")
+)
+
+var (
+	// safeLimitBytes is the maximum number of bytes the Kubernetes API will return
+	// for a single log request, bounding the size of the proxied stream.
+	safeLimitBytes = int64(100 * 1024 * 1024) // 100 MB
+	// defaultTailLines is the default number of lines to retrieve from the end of the logs.
+	defaultTailLines = int64(1000)
+)
```
Snehadas2005, line 52: I am quite curious if we might want to make these `const` instead of `var` at package scope? Since neither `safeLimitBytes` nor `defaultTailLines` are ever reassigned anywhere in this file or package, converting them to `const` would be slightly more idiomatic Go and prevent any accidental mutations down the road.

workspaces/backend/internal/repositories/logs/repo.go
```diff
@@ -0,0 +1,220 @@
...
+		Timestamps: true,
+		SinceTime:  opts.SinceTime,
+	})
+
+	stream, err := req.Stream(ctx)
+	if err != nil {
+		return nil, ErrOpenLogStream
+	}
+	return stream, nil
+}
```
Snehadas2005, line 102: I noticed a minor edge case here! What if a pod gets deleted or the container terminates in the tiny window right between our `ensureLogsAvailable` check and this `req.Stream(ctx)` call?
If that happens, `req.Stream()` will fail with a generic K8s API error, which gets wrapped as `ErrOpenLogStream`. Over in the HTTP handler (`GetWorkspacePodTemplateLogsHandler`), `ErrOpenLogStream` is not explicitly matched by any `errors.Is` case in the switch statement, so it degrades to `default: a.serverErrorResponse` (500) rather than a more descriptive 404/409.
It is super rare and definitely low severity, but I wanted to flag it in case you prefer to catch `ErrOpenLogStream` in the handler or map it more accurately!

workspaces/backend/api/workspace_podtemplate_logs_handler.go
```diff
@@ -0,0 +1,155 @@
...
+
+	if raw := query.Get(constants.TailLinesQueryParam); raw != "" {
+		tail, errs := helper.ValidateFieldIsPositiveInt64(field.NewPath(constants.TailLinesQueryParam), raw)
+		if len(errs) > 0 {
+			valErrs = append(valErrs, errs...)
+		} else {
+			opts.TailLines = tail
+		}
+	}
```
Snehadas2005, line 134: One small thought on `tailLines`: currently `ValidateFieldIsPositiveInt64` accepts any positive 64-bit integer (e.g. `9223372036854775807`). While `safeLimitBytes` (100MB) safely bounds the resulting payload size, the K8s API server still has to process and read up to that requested line count first.
Do you think it might be worth adding a reasonable upper cap (maybe matching the UI's max of 5000 lines or slightly above) as a cheap defense-in-depth measure?

workspaces/backend/manifests/kustomize/base/rbac.yaml
```diff
@@ -44,6 +44,12 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods/log
+  verbs:
+  - get
```
Snehadas2005, line 52: The diff snippet shows 6 lines added here, but the specific contents were collapsed in the view. Could you please double-check to make sure the backend's service account is granted `get` permissions on both `pods` and `pods/log`?
Since `.Pods().Get()` (for live-status checks) and `.GetLogs()` (for streaming) hit two distinct subresources in K8s RBAC terms, granting only `pods` would actually pass all our unit/envtest suites (since fake clientsets bypass RBAC enforcement) but fail when deployed onto a real cluster.

Snehadas2005 (commented):
Hey @christian-heusel, thanks for the detailed writeup and the implementation notes genuinely made this easy to follow. Went through both the frontend and backend parts and found a few things worth a look. 

Everything else from test coverage, the live-vs-cached client split to the pre-flight status check for precise errors, looked really solid. Looks really nice, especially given how much surface area this PR covers. 

Feel free to push back on any of these, happy to be wrong here! 😊

workspaces/frontend/src/generated/data-contracts.ts
```diff
@@ -427,9 +427,9 @@ export interface DetailsWorkspaceDetailPod {
 }
 
 export interface DetailsWorkspaceDetailVolumes {
-  data?: CommonPodVolumeInfo[];
-  home: CommonPodVolumeInfo;
-  secrets?: CommonPodSecretInfo[];
+  data?: DetailsPodVolumeInfo[];
```
thaorell, line 430: Should this be part of this PR? It is creating changes that are out-of-scope for this issue

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
```diff
@@ -0,0 +1,340 @@
+import React, { useCallback, useMemo, useState } from 'react';
```
thaorell: In order to keep this LogViewer component as minimal as possible, please consider not have filters for the first pass, as it could introduce tech debt. For a first pass, we should just have all log lines from all time and just the refresh button.
We can implement other features as well as comprehensive testing as an enhancement.

thaorell (changes requested):

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
christian-heusel: I disagree with the argument that this is introducing potential tech debt since this is exposing a well specified API in the UI that we maintain ourselves and hence we shouldn't be fearful of that this breaks in the future🤔 Could you go into more detail what you mean? Happy to slim things down if it makes the review easier though 🤗  

As I said in the PR description I think this should be a modal in the future (so it's not so narrow etc.) but that I future work IMO.

workspaces/frontend/src/generated/data-contracts.ts
christian-heusel: Will create a separate PR, I already split this into two commits in this PR so it's easy to do 🤗
Edit: ➡️ #1314

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
```diff
@@ -0,0 +1,340 @@
...
+        </EmptyStateBody>
+      </EmptyState>
+    );
+  }
+
+  // A PatternFly Toolbar is not used here: its content row is a flex item with `min-width: auto`,
+  // so in the (resizable, often narrow) details drawer it sizes to its content and overflows the
+  // panel instead of wrapping. A plain wrapping flex row keeps every control reachable.
+  // The search field only works inside the log viewer, as it relies on its context.
+  const renderToolbar = (withSearch: boolean) => (
```
thaorell: This component is almost 100 lines. Can you abstract it out to either a separate component and/or file?

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
thaorell: Agreed with Sneha here. This could be very confusing. A regular user (a data scientist) who rarely cares about log, unless the workspace is broken, might not know how to navigate the container selection

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
```diff
@@ -0,0 +1,340 @@
...
+const LabelledControl: React.FC<{ label: string; id: string; children: React.ReactNode }> = ({
+  label,
+  id,
+  children,
+}) => (
+  <FlexItem style={{ flex: '1 1 0', minWidth: 0 }}>
+    <div
+      id={id}
+      style={{
+        fontSize: 'var(--pf-t--global--font--size--sm)',
```
thaorell: Consider using utilities PF classes. Like className="pf-v6-u-text-color-subtle pf-v6-u-font-size-sm"

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
```diff
@@ -0,0 +1,340 @@
...
+  detailsError?: Error;
+}
+
+export const WorkspaceDetailsLogs: React.FC<WorkspaceDetailsLogsProps> = ({
+  workspace,
+  details,
+  detailsLoaded,
+  detailsError,
+}) => {
+  const containerNames = useMemo(
```
thaorell: This component is handling way too complex of a state management: containerNames, initContainerNames, container selection, tailLines, sinceLabels, etc. Consider moving those to `useWorkspaceLogs.ts`

thaorell (changes requested):
The API has seen quite rapid changes in the last month, so the introduction of rich features for a component that could be simplified does concern me. However, if you insist on keeping them. Can you please address the issues below?

In addition, the Cypress test does not cover the toolbar interactions (revert, log lines, time selection, downloads, etc). Can you please add those tests as well?

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
```diff
@@ -0,0 +1,340 @@
...
+  const [tailLines, setTailLines] = useState<number>(1000);
+  const [sinceLabel, setSinceLabel] = useState<string>(SINCE_OPTIONS[0].label);
+  const [previous, setPrevious] = useState(false);
+  const [isTextWrapped, setIsTextWrapped] = useState(false);
+  const [scrollToRow, setScrollToRow] = useState<number | undefined>();
+
+  // Default to the primary (first) container of the pod.
+  const container = selectedContainer ?? containerNames.at(0) ?? initContainerNames.at(0);
+
+  const sinceTime = useMemo(() => {
```
paulovmr: `sinceTime` is computed via `useMemo([sinceLabel])`, but captures `Date.now()` at memo-evaluation time. When the user clicks Refresh without changing the time range dropdown, `sinceTime` isn't recomputed — the API receives the original timestamp. For example, selecting "15 minutes" at 10:00 and refreshing at 10:30 returns 45 minutes of logs instead of 15. Consider computing `sinceTime` at fetch time inside the `useWorkspaceLogs` callback, or passing the `windowMs` value instead and computing the timestamp at request time.

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
christian-heusel: State now lives in a `useWorkspaceLogsController` hook with a `useEffect` on `[namespace, name]` that resets the selected container (and scroll position) whenever the workspace changes, so a container picked on one workspace can't leak onto the next pod and 400.

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
christian-heusel: Great catch! Fixed in the latest revision 🤗

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
christian-heusel: I don't really know how to improve the user experience here, do you have any suggestions? 🤔 
Should we just explain what a Pod is in a tooltip?

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
christian-heusel: Extracted the toolbar into `WorkspaceLogsToolbar.tsx` and moved the state into `useWorkspaceLogsController` ✅

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceDetailsLogs.tsx
christian-heusel: Done!

workspaces/frontend/src/__tests__/cypress/cypress/pages/workspaces/workspaces.ts
```diff
@@ -347,6 +347,51 @@ class WorkspaceDetailsDrawer {
...
+  }
+
+  findLogsDownloadButton() {
+    return this.findLogsTabContent().findByTestId('logs-download-button');
+  }
+
+  // The toolbar dropdowns are PatternFly SimpleSelects: click the toggle, then pick
+  // the option (which renders in a popper appended to the document body).
+  selectLogsContainer(optionName: string) {
+    this.findLogsTabContent().find('#logs-container-select').click();
```
thaorell, line 377: The project convention is `data-testid`. The Cypress page object works around this with `find('#logs-container-select')`  instead of the standard `findByTestId()`. Consider adding `'data-testid': 'logs-container-select'` to `toggleProps` alongside the `id` (the `id` should remain for `aria-labelledby`), and update the page object methods to use `findByTestId()`.

thaorell (commented):
Small comment below

thaorell (commented):
/lgtm

@christian-heusel I will open up a follow up for the comment above

paulovmr (approved):
/lgtm
/approve

Issue #1293, opened 2026-07-31
[TASK] [Backend] Expose GET `/podTemplate/resources` HTTP endpoint and OpenAPI schema


richabanker:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

This task exposes the workspace resource utilization repository as a REST API endpoint (GET `/api/v1/workspaces/:namespace/:name/podTemplate/resources`). It implements user authorization via Kubernetes SubjectAccessReview (SAR), formats responses using standard envelopes, and generates the OpenAPI/Swagger specification required for frontend client generation.

### Acceptance Criteria

- [ ]  A new HTTP route and handler are registered for `GET /api/v1/workspaces/:namespace/:name/podTemplate/resources`
- [ ] The endpoint handler performs a delegating SubjectAccessReview check on verb get for resource workspaces in `:namespace` for `:name `
- [ ] Successful and degraded (200 OK) responses are wrapped in `Envelope[WorkspaceResourceUsage] ({"data": ...})`
- [ ] Non-2xx HTTP errors (401 Unauthorized, 403 Forbidden, 404 Not Found, 422 Unprocessable Entity, 500 Internal Server Error) are wrapped in ErrorEnvelope
- [ ] OpenAPI/Swagger schema definitions (openapi/swagger.json, openapi/docs.go) are updated and valid for `/api/v1/workspaces/{namespace}/{name}/podTemplate/resources`

christian-heusel:
Resolved by https://github.com/kubeflow/notebooks/pull/1312, great work @richabanker ✅

PR #1312, opened 2026-08-07
feat: expose workspace podtemplate resources endpoint and OpenAPI schema

richabanker:
This PR exposes the workspace resource utilization repository from https://github.com/kubeflow/notebooks/pull/1241 as a REST endpoint `GET /api/v1/workspaces/:namespace/:name/podtemplate/resources` 

closes: https://github.com/kubeflow/notebooks/issues/1293

Tested on Tilt via `ENABLE_METRICS_SERVER=true make tilt-up`:

1. single container in workspace pod
```
curl -k -s -H "kubeflow-userid: admin" https://localhost:8443/workspaces/api/v1/workspaces/default/test/podtemplate/resources | jq .
{
  "data": {
    "containers": {
      "main": {
        "metricsFromMetricsServer": {
          "timestamp": "2026-08-06T22:35:53Z",
          "usage": {
            "cpu": "496767n",
            "memory": "181828Ki"
          }
        },
        "resources": {
          "requests": {
            "cpu": "100m",
            "memory": "128Mi"
          }
        }
      }
    }
  }
}
```

2. multiple containers in workspace pod
```
$ kubectl patch workspace test -n default --type='json' -p='[
  {
    "op": "add",
    "path": "/spec/podTemplate/podMetadata/annotations",
    "value": {
      "sidecar.istio.io/nativeSidecar": "false"
    }
  }
]'
workspace.kubeflow.org/test patched 

$ curl -k -s -H "kubeflow-userid: admin" https://localhost:8443/workspaces/api/v1/workspaces/default/test/podtemplate/resources | jq .
{
  "data": {
    "containers": {
      "istio-proxy": {
        "metricsFromMetricsServer": {
          "timestamp": "2026-08-06T22:52:41Z",
          "usage": {
            "cpu": "3864361n",
            "memory": "42776Ki"
          }
        },
        "resources": {
          "limits": {
            "cpu": "2",
            "memory": "1Gi"
          },
          "requests": {
            "cpu": "100m",
            "memory": "128Mi"
          }
        }
      },
      "main": {
        "metricsFromMetricsServer": {
          "timestamp": "2026-08-06T22:52:41Z",
          "usage": {
            "cpu": "0",
            "memory": "80300Ki"
          }
        },
        "resources": {
          "requests": {
            "cpu": "100m",
            "memory": "128Mi"
          }
        }
      }
    }
  }
}
```
3. For a paused workspace
```
$ kubectl patch workspace test -n default --type='merge' -p '{"spec":{"p
aused":true}}'
workspace.kubeflow.org/test patched

$ curl -k -s -H "kubeflow-userid: admin" \
  https://localhost:8443/workspaces/api/v1/workspaces/default/test/podtemplate/resources | jq .

{
  "error": {
    "code": "400",
    "message": "workspace pod is not running"
  }
}
```

4. When metrics.k8s.io API is not available
```
$ kubectl delete apiservice v1beta1.metrics.k8s.io

# Restart the backend pod
$ kubectl delete pod -n kubeflow-workspaces -l app=workspaces-backend
$ curl -k -s -H "kubeflow-userid: admin" \
  https://localhost:8443/workspaces/api/v1/workspaces/default/test/podtemplate/resources | jq .

{
  "data": {
    "containers": {
      "istio-proxy": {
        "resources": {
          "limits": {
            "cpu": "2",
            "memory": "1Gi"
          },
          "requests": {
            "cpu": "100m",
            "memory": "128Mi"
          }
        }
      },
      "main": {
        "resources": {
          "requests": {
            "cpu": "100m",
            "memory": "128Mi"
          }
        }
      }
    }
  }
}

```

### DIFF
--- a/workspaces/backend/api/response_errors.go
+++ b/workspaces/backend/api/response_errors.go
@@ -135,20 +135,6 @@ func (a *App) serverErrorResponse(w http.ResponseWriter, r *http.Request, err er
 	a.errorResponse(w, r, httpError)
 }
 
-// HTTP: 503
-func (a *App) serviceUnavailableResponse(w http.ResponseWriter, r *http.Request, err error) {
-	a.LogWarn(r, err.Error())
-
-	httpError := &HTTPError{
-		StatusCode: http.StatusServiceUnavailable,
-		ErrorResponse: ErrorResponse{
-			Code:    strconv.Itoa(http.StatusServiceUnavailable),
-			Message: err.Error(),
-		},
-	}
-	a.errorResponse(w, r, httpError)
-}
-
 // HTTP: 400
 func (a *App) badRequestResponse(w http.ResponseWriter, r *http.Request, err error) {
 	httpError := &HTTPError{
--- a/workspaces/backend/api/response_errors.go
+++ b/workspaces/backend/api/response_errors.go
@@ -135,6 +135,20 @@ func (a *App) serverErrorResponse(w http.ResponseWriter, r *http.Request, err er
 	a.errorResponse(w, r, httpError)
 }
 
+// HTTP: 503
+func (a *App) serviceUnavailableResponse(w http.ResponseWriter, r *http.Request, err error) {
+	a.LogWarn(r, err.Error())
+
+	httpError := &HTTPError{
+		StatusCode: http.StatusServiceUnavailable,
+		ErrorResponse: ErrorResponse{
+			Code:    strconv.Itoa(http.StatusServiceUnavailable),
+			Message: err.Error(),
+		},
+	}
+	a.errorResponse(w, r, httpError)
+}
+
 // HTTP: 400
 func (a *App) badRequestResponse(w http.ResponseWriter, r *http.Request, err error) {
 	httpError := &HTTPError{
--- a/workspaces/backend/api/workspace_podtemplate_resources_handler.go
+++ b/workspaces/backend/api/workspace_podtemplate_resources_handler.go
@@ -26,9 +26,8 @@ import (
 	"github.com/kubeflow/notebooks/workspaces/backend/api/constants"
 	"github.com/kubeflow/notebooks/workspaces/backend/internal/auth"
 	"github.com/kubeflow/notebooks/workspaces/backend/internal/helper"
-	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/metrics"
+	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/workspaces/podtemplate/resources"
 	repoCommon "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/common"
-	repoMetrics "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/metrics"
 )
 
 // WorkspaceResourceUsageEnvelope is the response envelope for workspace resource usage.
@@ -37,20 +36,19 @@ type WorkspaceResourceUsageEnvelope Envelope[*models.WorkspaceResourceUsage]
 // GetWorkspacePodTemplateResourcesHandler returns point-in-time resource usage for a workspace.
 //
 //	@Summary		Get workspace pod template resources
-//	@Description	Returns point-in-time CPU and memory usage for each container of the workspace pod, alongside the requests and limits configured in the pod spec. Usage is read from the Kubernetes Metrics Server.
+//	@Description	Returns point-in-time CPU and memory usage for each container of the workspace pod when available, alongside the requests and limits configured in the pod spec.
 //	@Tags			workspaces
 //	@ID				getWorkspacePodTemplateResources
 //	@Produce		json
 //	@Param			namespace	path		string							true	"Namespace of the workspace"	extensions(x-example=kubeflow-user-example-com)
 //	@Param			name		path		string							true	"Name of the workspace"			extensions(x-example=my-workspace)
-//	@Success		200			{object}	WorkspaceResourceUsageEnvelope	"Successful operation. Returns per-container usage and configured resources."
+//	@Success		200			{object}	WorkspaceResourceUsageEnvelope	"Successful operation. Returns per-container configured resources and live usage metrics when available."
 //	@Failure		400			{object}	ErrorEnvelope					"Bad Request. Workspace pod is not running."
 //	@Failure		401			{object}	ErrorEnvelope					"Unauthorized."
 //	@Failure		403			{object}	ErrorEnvelope					"Forbidden."
 //	@Failure		404			{object}	ErrorEnvelope					"Workspace not found."
 //	@Failure		422			{object}	ErrorEnvelope					"Unprocessable Entity. Validation error."
 //	@Failure		500			{object}	ErrorEnvelope					"Internal server error."
-//	@Failure		503			{object}	ErrorEnvelope					"Service Unavailable. Metrics API is not available."
 //	@Router			/workspaces/{namespace}/{name}/podtemplate/resources [get]
 func (a *App) GetWorkspacePodTemplateResourcesHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	namespace := ps.ByName(constants.NamespacePathParam)
@@ -79,9 +77,7 @@ func (a *App) GetWorkspacePodTemplateResourcesHandler(w http.ResponseWriter, r *
 		switch {
 		case errors.Is(err, repoCommon.ErrWorkspaceNotFound):
 			a.notFoundResponse(w, r)
-		case errors.Is(err, repoMetrics.ErrMetricsAPINotAvailable):
-			a.serviceUnavailableResponse(w, r, err)
-		case errors.Is(err, repoMetrics.ErrWorkspaceNotRunning):
+		case errors.Is(err, repoCommon.ErrWorkspacePodNotRunning):
 			a.badRequestResponse(w, r, err)
 		default:
 			a.serverErrorResponse(w, r, err)
--- a/workspaces/backend/api/workspace_podtemplate_resources_handler.go
+++ b/workspaces/backend/api/workspace_podtemplate_resources_handler.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+import (
+	"errors"
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	"github.com/kubeflow/notebooks/workspaces/backend/api/constants"
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/auth"
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/helper"
+	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/metrics"
+	repoCommon "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/common"
+	repoMetrics "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/metrics"
+)
+
+// WorkspaceResourceUsageEnvelope is the response envelope for workspace resource usage.
+type WorkspaceResourceUsageEnvelope Envelope[*models.WorkspaceResourceUsage]
+
+// GetWorkspacePodTemplateResourcesHandler returns point-in-time resource usage for a workspace.
+//
+//	@Summary		Get workspace pod template resources
+//	@Description	Returns point-in-time CPU and memory usage for each container of the workspace pod, alongside the requests and limits configured in the pod spec. Usage is read from the Kubernetes Metrics Server.
+//	@Tags			workspaces
+//	@ID				getWorkspacePodTemplateResources
+//	@Produce		json
+//	@Param			namespace	path		string							true	"Namespace of the workspace"	extensions(x-example=kubeflow-user-example-com)
+//	@Param			name		path		string							true	"Name of the workspace"			extensions(x-example=my-workspace)
+//	@Success		200			{object}	WorkspaceResourceUsageEnvelope	"Successful operation. Returns per-container usage and configured resources."
+//	@Failure		400			{object}	ErrorEnvelope					"Bad Request. Workspace pod is not running."
+//	@Failure		401			{object}	ErrorEnvelope					"Unauthorized."
+//	@Failure		403			{object}	ErrorEnvelope					"Forbidden."
+//	@Failure		404			{object}	ErrorEnvelope					"Workspace not found."
+//	@Failure		422			{object}	ErrorEnvelope					"Unprocessable Entity. Validation error."
+//	@Failure		500			{object}	ErrorEnvelope					"Internal server error."
+//	@Failure		503			{object}	ErrorEnvelope					"Service Unavailable. Metrics API is not available."
+//	@Router			/workspaces/{namespace}/{name}/podtemplate/resources [get]
+func (a *App) GetWorkspacePodTemplateResourcesHandler(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	namespace := ps.ByName(constants.NamespacePathParam)
+	workspaceName := ps.ByName(constants.ResourceNamePathParam)
+
+	// validate path parameters
+	var valErrs field.ErrorList //nolint:prealloc
+	valErrs = append(valErrs, helper.ValidateKubernetesNamespaceName(field.NewPath(constants.NamespacePathParam), namespace)...)
+	valErrs = append(valErrs, helper.ValidateWorkspaceName(field.NewPath(constants.ResourceNamePathParam), workspaceName)...)
+	if len(valErrs) > 0 {
+		a.failedValidationResponse(w, r, errMsgPathParamsInvalid, valErrs, nil)
+		return
+	}
+
+	// =========================== AUTH ===========================
+	authPolicies := []*auth.ResourcePolicy{
+		auth.NewResourcePolicy(auth.VerbGet, auth.Workspaces, auth.ResourcePolicyResourceMeta{Namespace: namespace, Name: workspaceName}),
+	}
+	if _, ok := a.requireAuth(w, r, authPolicies); !ok {
+		return
+	}
+	// ============================================================
+
+	usage, err := a.repositories.Metrics.GetWorkspaceResourceUsage(r.Context(), namespace, workspaceName)
+	if err != nil {
+		switch {
+		case errors.Is(err, repoCommon.ErrWorkspaceNotFound):
+			a.notFoundResponse(w, r)
+		case errors.Is(err, repoMetrics.ErrMetricsAPINotAvailable):
+			a.serviceUnavailableResponse(w, r, err)
+		case errors.Is(err, repoMetrics.ErrWorkspaceNotRunning):
+			a.badRequestResponse(w, r, err)
+		default:
+			a.serverErrorResponse(w, r, err)
+		}
+		return
+	}
+
+	responseEnvelope := &WorkspaceResourceUsageEnvelope{Data: usage}
+	a.dataResponse(w, r, responseEnvelope)
+}
--- a/workspaces/backend/internal/models/workspaces/podtemplate/resources/types.go
+++ b/workspaces/backend/internal/models/workspaces/podtemplate/resources/types.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package metrics
+package resources
 
 import (
 	corev1 "k8s.io/api/core/v1"
--- a/workspaces/backend/internal/repositories/common/errors.go
+++ b/workspaces/backend/internal/repositories/common/errors.go
@@ -1,0 +1,27 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package common holds values shared between repositories.
+//
+// Sentinel errors live here when more than one repository can return them, so that callers
+// have a single value to match with errors.Is. Sentinels raised by exactly one repository
+// stay in that repository's own package.
+package common
+
+import "errors"
+
+// ErrWorkspaceNotFound is returned when a Workspace does not exist.
+var ErrWorkspaceNotFound = errors.New("workspace not found")
--- a/workspaces/backend/internal/repositories/common/errors.go
+++ b/workspaces/backend/internal/repositories/common/errors.go
@@ -25,3 +25,6 @@ import "errors"
 
 // ErrWorkspaceNotFound is returned when a Workspace does not exist.
 var ErrWorkspaceNotFound = errors.New("workspace not found")
+
+// ErrWorkspacePodNotRunning is returned when a workspace pod is not running.
+var ErrWorkspacePodNotRunning = errors.New("workspace pod is not running")
--- a/workspaces/backend/internal/repositories/metrics/repo.go
+++ b/workspaces/backend/internal/repositories/metrics/repo.go
@@ -22,7 +22,9 @@ import (
 	"sync"
 	"time"
 
+	kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	metricsv1beta1 "k8s.io/metrics/pkg/apis/metrics/v1beta1"
@@ -31,6 +33,12 @@ import (
 	"github.com/kubeflow/notebooks/workspaces/backend/internal/config"
 	modelsCommon "github.com/kubeflow/notebooks/workspaces/backend/internal/models/common"
 	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/metrics"
+	repoCommon "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/common"
+)
+
+var (
+	ErrMetricsAPINotAvailable = fmt.Errorf("metrics API is not available")
+	ErrWorkspaceNotRunning    = fmt.Errorf("workspace pod is not running")
 )
 
 const (
@@ -57,13 +65,24 @@ func NewMetricsRepository(cfg *config.EnvConfig, c client.Client) *MetricsReposi
 
 // GetWorkspaceResourceUsage returns the resource usage for all pods in the given namespace and workspace.
 func (r *MetricsRepository) GetWorkspaceResourceUsage(ctx context.Context, ns, workspace string) (*models.WorkspaceResourceUsage, error) {
+	// Confirm the workspace exists first. Usage is resolved by pod label, so without this a
+	// workspace that never existed would be indistinguishable from one that is merely paused,
+	// and both would report WORKSPACE_NOT_RUNNING. This read is served from the informer cache.
+	ws := &kubefloworgv1beta1.Workspace{}
+	if err := r.client.Get(ctx, client.ObjectKey{Namespace: ns, Name: workspace}, ws); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, repoCommon.ErrWorkspaceNotFound
+		}
+		return nil, err
+	}
+
 	available, err := r.apiAvailable()
 	if err != nil {
 		return nil, err
 	}
 
 	if !available {
-		return models.NewErrorResourceUsage(models.ErrorCodeMetricsAPINotAvailable), nil
+		return nil, ErrMetricsAPINotAvailable
 	}
 
 	selector := client.MatchingLabels{modelsCommon.LabelWorkspaceName: workspace}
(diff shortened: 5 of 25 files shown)


richabanker:
Some points / questions I'd like to highlight:

1. This is more of a fix done in this PR which I want to call out: Previously in `metrics/repo.go`, the backend [searches for pods with a matching label for workspace-name](https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/internal/repositories/metrics/repo.go#L69-L84), but for a workspace that does not exist, this would return 0 pods,  which was mistakenly returning `200 OK` with `WORKSPACE_NOT_RUNNING`. So in this PR, I added an existence check that ensures the API distinguishes non-existent workspaces from paused ones and returns a proper `404 Not Found`
2. For the above, I moved `ErrWorkspaceNotFound` out of `workspaces/backend/internal/repositories/workspaces/repo.go` and into a new `workspaces/backend/internal/repositories/common/errors.go` since it is now raised by both the `workspaces` repository and the `metrics` repository, so wanted to keep it as a single shared sentinel error across repositories
3. This is more of a limitation of the current tests added in `workspaces/backend/api/workspace_podtemplate_resources_handler_test.go` - The API handler test suite runs against `envtest`, which provides an in-process kube-apiserver without aggregated API servers like Metrics Server. So currently this lacks a happy path test, but I included one using a fake client for metrics-server in `workspaces/backend/internal/repositories/metrics/repo_test.go`. If there are any suggestions to add a happy path test in `workspace_podtemplate_resources_handler_test.go`, I'm open to that
4. Finally one open question that I realized while testing the new endpoint in Tilt: Because controller-runtime caches API discovery in memory upon startup, if Metrics Server crashes, restarts, or is uninstalled while the backend is running, the discovery probe still reports true, but `client.List` for `podMetricsList` receives a `404` or `503` from kube-apiserver, surfacing as an HTTP `500` error. Should we update client.List in [repo.go](https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/internal/repositories/metrics/repo.go#L84) to catch `apierrors.IsNotFound(err)` and `apierrors.IsServiceUnavailable(err)` and return `METRICS_API_NOT_AVAILABLE (200 OK)` so the UI always degrades gracefully during cluster maintenance or pod restarts? Or is `500` error acceptable for this case?

workspaces/backend/api/workspace_podtemplate_resources_handler.go
```diff
@@ -0,0 +1,86 @@
...
+// GetWorkspacePodTemplateResourcesHandler returns point-in-time resource usage for a workspace.
+//
+//	@Summary		Get workspace pod template resources
+//	@Description	Returns point-in-time CPU and memory usage for each container of the workspace pod, alongside the requests and limits configured in the pod spec. Usage is read from the Kubernetes Metrics Server. When usage cannot be retrieved the response is still 200 OK, with a `data.error` code explaining why (`METRICS_API_NOT_AVAILABLE` when the Metrics Server is not installed, `WORKSPACE_NOT_RUNNING` when the workspace has no pods).
+//	@Tags			workspaces
+//	@ID				getWorkspacePodTemplateResources
+//	@Produce		json
+//	@Param			namespace	path		string							true	"Namespace of the workspace"	extensions(x-example=kubeflow-user-example-com)
+//	@Param			name		path		string							true	"Name of the workspace"			extensions(x-example=my-workspace)
+//	@Success		200			{object}	WorkspaceResourceUsageEnvelope	"Successful operation. Returns per-container usage and configured resources, or an error code when usage is unavailable."
```
christian-heusel: I think we should add a bit more error handling here, i.e. if the metrics server is not available this endpoint shouldn't return 200:
```
$ curl -kvX 'GET' \
  'https://localhost:8443/workspaces/api/v1/workspaces/default/test/podtemplate/resources' \
  -H 'accept: application/json' \
  -H 'kubeflow-userid: admin'
[...]
< HTTP/2 200
< access-control-allow-origin: *
< content-type: application/json
< vary: Accept-Encoding
< date: Fri, 07 Aug 2026 10:08:32 GMT
< content-length: 46
< x-envoy-upstream-service-time: 1
< server: istio-envoy
<
* Connection #0 to host localhost:8443 left intact
{"data":{"error":"METRICS_API_NOT_AVAILABLE"}}
```

Maybe something like [503 Service Unavailable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/503) would be more appropriate?

workspaces/backend/api/workspace_podtemplate_resources_handler.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/kubeflow/notebooks/workspaces/backend/api/constants"
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/auth"
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/helper"
+	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/metrics"
+	repoCommon "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/common"
+)
+
+// WorkspaceResourceUsageEnvelope is the response envelope for workspace resource usage.
+type WorkspaceResourceUsageEnvelope Envelope[*models.WorkspaceResourceUsage]
```
christian-heusel, line 34: The schema for this doesn't adequately model the response right now:
```json
{
  "data": {
    "containers": {
      "additionalProp1": {
        "metricsFromMetricsServer": "string",
        "resources": "string"
      },
      "additionalProp2": {
        "metricsFromMetricsServer": "string",
        "resources": "string"
      },
      "additionalProp3": {
        "metricsFromMetricsServer": "string",
        "resources": "string"
      }
    },
    "error": "string"
  }
}
```

vs the actual return:
```json
{
  "data": {
    "containers": {
      "main": {
        "metricsFromMetricsServer": {
          "timestamp": "2026-08-07T10:22:04Z",
          "usage": {
            "cpu": "15143n",
            "memory": "53952Ki"
          }
        },
        "resources": {
          "requests": {
            "cpu": "1",
            "memory": "2Gi"
          }
        }
      }
    }
  }
}
```

christian-heusel: See https://localhost:8443/workspaces/api/v1/swagger/index.html#/workspaces/getWorkspacePodTemplateResources

christian-heusel (changes requested):
Hey @richabanker, thanks a lot for driving this forward 🤗 

I did a brief review, see the comments below!

workspaces/backend/api/workspace_podtemplate_resources_handler.go
richabanker: Not sure I understand, for a non-existent workspace, we already return 404 NOT FOUND. 

For the case when metrics-server wasnt available, we agreed in the design discussion to indicate this degraded state behind a 200 OK still - check "200 OK — Degraded States" section of the comment  https://github.com/kubeflow/notebooks/issues/1223#issuecomment-5072632630

Are we saying we should not have the "error" field in the response at all and denote those errors as [503 Service Unavailable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/503) ?

cc @andyatmiami

christian-heusel: Yeah I think it makes far more sense to return a 503 here, since indeed the service is unavailable 🤔 
See https://github.com/kubeflow/notebooks/pull/1328/changes/ece95299d49147edb64430cbdf65bd3637b61cb9 for my sample implementation 🤗

workspaces/backend/api/workspace_podtemplate_resources_handler.go
richabanker: Hmm, I see the following schema for the 200 OK response:
```
{
  "data": {
    "containers": {
      "additionalProp1": {
        "metricsFromMetricsServer": {
          "timestamp": "string",
          "usage": {
            "cpu": "string",
            "memory": "string"
          }
        },
        "resources": {
          "claims": [
            {
              "name": "string",
              "request": "string"
            }
          ],
          "limits": {
            "additionalProp1": {
              "Format": "DecimalExponent"
            }
          },
          "requests": {
            "additionalProp1": {
              "Format": "DecimalExponent"
            }
          }
        }
      }
    },
    "error": "METRICS_API_NOT_AVAILABLE"
  }
```

While on my end `metricsFromMetricsServer` resolves its nested fields, I agree that the schema definition for resources is problematic:

1. requests and limits are rendered as { "Format": "DecimalExponent" } objects instead of string quantities
2. It includes an unwanted claims array from Kubernetes DRA

Its likely because we are embedding Kubernetes' internal `corev1.ResourceRequirements` into `ContainerResourceUsage.Resources` directly, which swaggo seems to have issues un-nesting...

A potential workaround could be to stop embedding `corev1.ResourceRequirements` in the response model, rather use a new explicit `ContainerResources` DTO with Requests/Limits as map[string]string. WDYT?

cc @andyatmiami

workspaces/backend/api/workspace_podtemplate_resources_handler.go
richabanker: One thing to consider maybe: https://datatracker.ietf.org/doc/html/rfc9110#name-503-service-unavailable defines 503 error code to be a temporary retry-able error, but if the metrics.k8s.io API is not available, i.e. the metrics-server is not installed, its more of a permanent cluster config, until the operator installs one. If we made this a 503, we may indicate to the UI that you could trigger retries for this endpoint, even though we know it wouldnt help...

andyatmiami: sorry i overlooked this..

from an API/schema perspective - I think we **must** have  these `200 OK — Degraded States` scenarios actually **NOT** return `200` - just to keep the schema-to-responde-code-mapping sensible..

the 2 "degraded states" already follow the error envelope schema itself - we need to be in a state whereby 2xx == "happy data envelope" and non-2xx == "error envelope"

workspaces/backend/api/workspace_podtemplate_resources_handler.go
andyatmiami: sorry @christian-heusel - can you be explicit on what exactly you are raising here?  @richabanker response touches on some potential schema-related things - but I still wanna understand the discrepancy b/w her schema and your schema as they aren't remotely the same... 🤔

workspaces/backend/api/workspace_podtemplate_resources_handler.go
richabanker: Ok sg. I have refactored the endpoint and data models to strictly follow the `2xx == data envelope` and `non-2xx == error envelope` contract:

1. HTTP Status Code Mapping
-`503 Service Unavailable`: Returned when the cluster's Metrics Server / `metrics.k8s.io` API is not available.
- `400 Bad Request`: Returned when the workspace pod is not running (e.g., paused), aligning with the pattern used by `/podtemplate/logs/batch`.
- `404 Not Found`: Returned when the workspace CR does not exist.
- `200 OK`: Strictly returns the happy data envelope (`{"data": {"containers": {...}}}`).

2. Schema & Model cleanup
- Dropped `data.error` and the `ErrorCode` enum from `WorkspaceResourceUsage`.
- `containers` is now marked as required (non-`omitempty`) in `WorkspaceResourceUsage`, ensuring frontend generators type it as a guaranteed dict

richabanker: oh btw I just remembered why we did have an `Error` field in the `WorkspaceResourceUsage` struct originally. Even when the metrics-server is unavailable, we still have the requests/limits data from the pod spec, which we would be throwing away if we returned a blanket 503 error based on metrics-server availability.. which seems like throwing away useful info.

andyatmiami: Ah, yes... thanks for calling that out @richabanker ... as that fact lets us still preserve '2xx == data envelope' + 'non-2xx == error envelope' - we just need to be more crisp in our definition/expectation on what populates `error` on the `2xx` case...

will stew on this a bit and follow up

workspaces/backend/api/workspace_podtemplate_resources_handler.go
christian-heusel: The main thing I wanted to point out is this difference:
```json
          "requests": {
            "additionalProp1": {
              "Format": "DecimalExponent"
            }
          }
```
vs the actual response:
```json
        "resources": {
          "requests": {
            "cpu": "1",
            "memory": "2Gi"
          }
        }
```

However this is more developer-than user facing, I just wanted to get us thinking if we should use a more narrow definition here/if we actually make use of the more general type.

image
https://localhost:8443/workspaces/api/v1/swagger/index.html#/workspaces/getWorkspacePodTemplateResources

workspaces/backend/api/workspace_podtemplate_resources_handler.go
christian-heusel: Discussed (at length) with @andyatmiami in the leads call: I'm fine with the non-503 return here, essentially with the reasoning that we still have something sensible (i.e. successfull to return) that Richa provided here: https://github.com/kubeflow/notebooks/pull/1312#discussion_r3800929052

workspaces/backend/api/response_errors.go
```diff
@@ -135,6 +135,20 @@ func (a *App) serverErrorResponse(w http.ResponseWriter, r *http.Request, err er
 	a.errorResponse(w, r, httpError)
 }
 
+// HTTP: 503
+func (a *App) serviceUnavailableResponse(w http.ResponseWriter, r *http.Request, err error) {
+	a.LogWarn(r, err.Error())
```
andyatmiami: I don't love the idea of logging a `warn` statement **per request** - even in this particular use case with Metrics API - and then even moreso as a "general contract" for future callers of this function.

I think this could lead to overly chatty (and uninformative) logs...

Admittedly there isn't a great alternative right now - but in general an "access log" (via istio) with a `503` return code is a signal that cluster admins/operators could key off of..

So, i'd say for now we simply drop this log line.. and do nothing else..

Thoughts?

workspaces/backend/api/workspace_podtemplate_resources_handler.go
andyatmiami: The problem we have here is that there is **NO** strict type in the k8s layer..

early discussions had us agree that we'd simply re-use the map-like structure inherent to the k8s definition... and that is then how this gets rendered.. the key names themselves are dynamic/undefined-until-runtime - which then seems to get rendered in swagger as `additionalProp1` ..

I'd say we continue with this as-is (as we originally agreed to) - and then let the `frontend` be the _ultimate decider_ here (since they are the one real/true client we care to serve with the API)

If their generated code _works for them_ to support the feature afforded by this API - then nothing needs to change (because nothing is inherently broken 😄 )

workspaces/backend/internal/repositories/common/errors.go
```diff
@@ -0,0 +1,27 @@
...
+//
+// Sentinel errors live here when more than one repository can return them, so that callers
+// have a single value to match with errors.Is. Sentinels raised by exactly one repository
+// stay in that repository's own package.
+package common
+
+import "errors"
+
+// ErrWorkspaceNotFound is returned when a Workspace does not exist.
+var ErrWorkspaceNotFound = errors.New("workspace not found")
```
andyatmiami, line 27: In parallel to this PR being worked - I had merged a PR introducing the Pod logs API... which then introduced _yet another_ `ErrWorkspaceNotFound` variant 😇 
- 	https://github.com/kubeflow/notebooks/blob/notebooks-v2/workspaces/backend/internal/repositories/podlogs/repo.go#L37

can you rebase this PR to pick up that change and then flip that over to this common error as well?

thanks!

workspaces/backend/internal/repositories/metrics/repo.go
```diff
@@ -82,6 +101,9 @@ func (r *MetricsRepository) GetWorkspaceResourceUsage(ctx context.Context, ns, w
 	pod := &podList.Items[0]
 	podMetricsList := &metricsv1beta1.PodMetricsList{}
 	if err := r.client.List(ctx, podMetricsList, client.InNamespace(ns), selector); err != nil {
+		if apierrors.IsNotFound(err) || apierrors.IsServiceUnavailable(err) || apierrors.IsTimeout(err) {
+			return nil, ErrMetricsAPINotAvailable
+		}
```
andyatmiami, line 94: Open for discussion here...

I don't love the idea of having 3 different "sources" for the `ErrMetricsAPINotAvailable` sentinel error - but not having any way to really distinguish what led to that error being returned..

**ON THE OTHER HAND** - prior in this PR review I said I **ALSO** don't like any type of per-request error logging as that gets too chatty for an error that fundamentally is not tied to any given request lifecycle..

It kinda feels like the `memoize` function we introduced for metrics is **probably?** the right place to address my concerns.. but that itself has its drawbacks:
- that function is presently more generic in nature
- that function does not have access to a logger

However, `memoize` is the right operational pattern i'd be interested in to given app admins insight into _potentially deviant behavior_ ..

- it only runs on a TTL (so totally decoupled from requests / reasonable log volume)
- could fire once on initial app load
- could be provided some semblance of "prior state" to then only log additionally on "state transitions"
    - i.e. from 'running/healthy' -> 'unavailable' (as this is the case that an admin would care about)

This gets very granular and could certainly be delivered in a follow up PR (if you think its warranted)... so for now not requesting change... only wanna get your thoughts on the subject.

workspaces/backend/api/workspace_podtemplate_resources_handler.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/julienschmidt/httprouter"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	"github.com/kubeflow/notebooks/workspaces/backend/api/constants"
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/auth"
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/helper"
+	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/metrics"
+	repoCommon "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/common"
+	repoMetrics "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/metrics"
```
andyatmiami:
```suggestion
	repository "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/metrics"
```

General convention on handlers is that (to date) they typically map to a single/primary repository... and we then alias that import as `repository`

so, essentially, every handler will have a `models` and `repository` alias (although obviously the packages _actually being imported_ will differ across handlers.

workspaces/backend/internal/models/workspaces/podtemplate/resources/types.go
andyatmiami, line 1: ℹ️ _No good spot to put this review comment - so just attaching it to this file_

Apologies I had overlooked this on the original PR establishing this `models/` directory... as we iterated the API Proposal - I should have appreciated the ramifications..

The endpoint is at `/workspaces/{ns}/{name}/podtemplate/resources`, but types live in `internal/models/metrics/` instead of `internal/models/workspaces/podtemplate/resources/`. The other podtemplate sub-resources follow the convention: 
- `details` -> `models/workspaces/podtemplate/details/`
- `logs` -> `models/workspaces/podtemplate/logs/`.

So, we should move `internal/models/metrics/` to:
- `internal/models/workspaces/podtemplate/resources/`

workspaces/backend/internal/repositories/metrics/repo.go
```diff
@@ -31,6 +33,12 @@ import (
 	"github.com/kubeflow/notebooks/workspaces/backend/internal/config"
 	modelsCommon "github.com/kubeflow/notebooks/workspaces/backend/internal/models/common"
 	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/metrics"
+	repoCommon "github.com/kubeflow/notebooks/workspaces/backend/internal/repositories/common"
+)
+
+var (
+	ErrMetricsAPINotAvailable = fmt.Errorf("metrics API is not available")
+	ErrWorkspaceNotRunning    = fmt.Errorf("workspace pod is not running")
```
andyatmiami:
```suggestion
	ErrMetricsAPINotAvailable = errors.New("metrics API is not available")
	ErrWorkspaceNotRunning    = errors.New"workspace pod is not running")
```

Just as you did in https://github.com/kubeflow/notebooks/pull/1312/changes#diff-8206edcd75fc2bcc3109967d8c56a6b92d4b4a323da0bad014d5396b69c9dac1R27 , if we don't require parameterization on the sentinel error string - we should prefer usage of `errors.New`

workspaces/backend/internal/repositories/metrics/repo_test.go
```diff
@@ -24,7 +24,9 @@ import (
 
 	modelsCommon "github.com/kubeflow/notebooks/workspaces/backend/internal/models/common"
 	modelsmetrics "github.com/kubeflow/notebooks/workspaces/backend/internal/models/metrics"
```
andyatmiami:
```suggestion
	models "github.com/kubeflow/notebooks/workspaces/backend/internal/models/metrics"
```

Again, sorry this was missed in initial review - but I think we should keep consistency of having the "primary" models package be aliased as `models` (as in the production code)
- and, as an aside - if we did want to give a more granular suffix (even though in this case **we do not**) - per convention it should be camelCased

andyatmiami (changes requested):
@richabanker - thanks for driving great discussion on this PR as well as delivering another high quality contribution..

as per _norm_ - almost all my comments are trivial syntactic/consistency things..

one fun open question on logging frequency - but that won't necessarily need to hold up merging of this PR... looking forward to getting your thoughts and getting this merged!

workspaces/backend/api/workspace_podtemplate_resources_handler.go
andyatmiami: Ok, I want to "resurrect" this thread one more time 😈 

Looking at the code as checked out, the current implementation still returns a blanket `503` for `ErrMetricsAPINotAvailable` — which throws away the `requests`/`limits` data  entirely. So the code doesn't match where (I thought?) the discussion landed - but that could have been due to poor specificity on my part (for which I apologize!)

The two cases where a 200 with partial data would make sense:

1. Metrics-server unavailable, pod running — we have `requests`/`limits` from the pod spec, just no live usage metrics. Return `200` with `containers` populated (`requests`/`limits` filled, `metricsFromMetricsServer` null/absent).

2. Metrics-server available but hasn't scraped yet (fresh pod, <15s) — same situation: `requests`/`limits` exist, `usage` data doesn't yet. Return 200 with the same partial shape.

Both fit the "2xx = data envelope" rule since there is meaningful data to return. The `metricsFromMetricsServer` field being null/absent is the signal to the frontend that live  metrics aren't available — no need for a separate error code.

Now, I realize a downside to this proposal here is that scenarios 1 and 2 have vastly different cluster states that lead to the **same JSON response** - which isn't great.. but I'd think if frontend would ever need/want to distinguish these - a better solution (vs. trying to report this information on the response of the this API) - is to have a general (maybe the `/healthcheck` API?) endpoint that reports "cluster-wide" information - like presence/absence of Metrics Server... and certainly in the short term - we don't really have a need to differentiate... our simply rendering will either show usage (or not)..

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Since now we dont return ErrMetricsAPINotAvailable, so that we can atleast return the data that we do have (pod requests / limits) even when metrics-server is unreachable, we wont have this `if apierrors.IsNotFound(err) || apierrors.IsServiceUnavailable(err) || apierrors.IsTimeout(err) ` block anymore (pushing that change in a moment)

But, I think if possible, we should still keep some debug log or log at a higher verbosity levels (via `slog.DebugContext(ctx, ...)`), whenever we do encounter errors while trying to reach metrics-server in order to have some visibility into "Why are live metrics missing for this workspace?".

workspaces/backend/api/workspace_podtemplate_resources_handler.go
richabanker: ok finally got to pushing the changes addressing the above (sorry for the delay here!). Summarizing the changes:

#### 200 OK` (Full vs Partial Data)
Once a running pod is found, the endpoint always returns `200 OK` with the pod's configured `requests`/`limits` (`resources`). Live utilization is enriched on a best-effort basis:
* **Full Data (`200 OK`)**: Metrics Server is healthy and query succeeds $\rightarrow$ returns configured `resources` + live `metricsFromMetricsServer` (CPU/Memory usage & timestamp).
* **Partial Data (`200 OK`)**: Metrics Server is uninstalled, warming up, failing, or the backend service account lacks RBAC for `metrics.k8s.io` $\rightarrow$ gracefully degrades to returning configured `resources` with `metricsFromMetricsServer: null` (never loses configured pod specs).

workspaces/backend/api/workspace_podtemplate_resources_handler.go
richabanker: Sg, keeping things ad they are for now. Will revisit if frontend expresses any issues.

workspaces/backend/api/response_errors.go
richabanker: The whole `serviceUnavailableResponse()` became obsolete since we decided not to return blanket 503s when metrics API is not available, so removed this function entirely.

workspaces/backend/internal/repositories/common/errors.go
richabanker: Done

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: I didnt add the debug logs yet though. Right now `GetWorkspaceResourceUsage()` is swallowing err from `r.apiAvailable()` and `r.client.List(ctx, podMetricsList, client.InNamespace(ns), selector)` without surfacing those anywhere. Lmk if you think its fine to keep that as is.

workspaces/backend/api/workspace_podtemplate_resources_handler.go
richabanker: Done

workspaces/backend/internal/models/workspaces/podtemplate/resources/types.go
richabanker: Done

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Done

workspaces/backend/internal/repositories/metrics/repo_test.go
richabanker: Done

workspaces/backend/internal/repositories/metrics/repo.go
andyatmiami, line 120: Sorry, I think I missed this in the initial review 😢 

When `probe()` returns an error, `checkedAt` is never updated (line 123-126):

```
v, err := probe()
if err != nil {
    return false, err   // checkedAt stays at zero/stale
}
val, checkedAt = v, time.Now()  // only set on success
```

This means if the REST mapper is broken (not `NoMatchError` — some other failure like connectivity issues or API server overload), the memoize cache never takes effect. Every incoming request to the `/resources` endpoint re-triggers a fresh discovery call to the API server.

The caller at line 93-95 then swallows the error:

```
available, err := r.apiAvailable()
if err != nil || !available {
    return models.NewWorkspaceResourceUsage(pod, nil), nil
}
```

So the error silently disappears, and the next request hits the API server again. Under load, this creates an amplification loop: the API server is struggling → discovery calls fail → every request retries discovery → more pressure on the API server.

**Suggested Fix**: Cache the error result with the same TTL (or a shorter one). Something like:

```
v, err := probe()
if err != nil {
    val, checkedAt = false, time.Now()  // cache the negative result
    return false, nil                    // or return false, err if caller cares
}
```

Whether the error itself should be returned to the caller is a design choice you can make - I'm indifferent given current structure of the code (with partial responses being returned). Since `GetWorkspaceResourceUsage` treats `err != nil` and `!available` identically (both → partial
data), caching the `false` result and returning `nil` error is probably the cleanest path.

workspaces/backend/internal/repositories/metrics/repo.go
```diff
@@ -57,32 +65,39 @@ func NewMetricsRepository(cfg *config.EnvConfig, c client.Client) *MetricsReposi
...
+	available, err := r.apiAvailable()
+	if err != nil || !available {
+		return models.NewWorkspaceResourceUsage(pod, nil), nil
+	}
+
 	podMetricsList := &metricsv1beta1.PodMetricsList{}
 	if err := r.client.List(ctx, podMetricsList, client.InNamespace(ns), selector); err != nil {
-		return nil, err
+		return models.NewWorkspaceResourceUsage(pod, nil), nil
 	}
```
andyatmiami: Oof - my initial review apparently wasn't great as I overlooked this too (albeit it is minor)...

This catches every error and returns partial data. The tests cover the intended cases well (`NotFound`, `ServiceUnavailable`, `Forbidden`), but this also swallows `context.Canceled` and `context.DeadlineExceeded`. When a client disconnects mid-request, the handler still constructs a response and attempts to write it — which will be destined to fail since the socket will have already been closed.

**Suggested Fix**:  Propagate context errors

```
if err := r.client.List(ctx, podMetricsList, ...); err != nil {
    if ctx.Err() != nil {
        return nil, ctx.Err()
    }
    return models.NewWorkspaceResourceUsage(pod, nil), nil
}
```

andyatmiami (changes requested):
Hey @richabanker - this is looking **really good**...

In my last _invasive_ set of review/testing - found a couple pre-existing issues I missed in `repo.go` that I hope we can just address now...

Not critically impacting - but good to just knock it out _now that I can't unsee it_ .

After this last batch of changes - I will simply re-run my crafted test suite and expect to approve this PR! 🎉

As always, let me know if you have any question and/or disagreements!

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Hmm, your point about current code treating `err != nil` and `!available` identical is valid. I think this is going back to whether we should in fact have some sort of an indication in the response payload about why the usage metrics are missing. Just plainly returning bool while swallowing all errors while trying to check if metrics API available seems ~not great. We are losing all visibility into why the metrics are not available. There should be some record of that, either in the API response or in the logs. Currently we have none.

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Good point, incorporated the fix. Thanks!

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: ok discussed offline, lets track the "losing visibility into errors" as a separate discussion / effort. For now, I am now caching the metrics-api availability as false upon `error != nil` as well. I also went ahead and removed the err from the return value for `metricsAPIServed()` since we dont meaningfully distinguish between the `err != nil` and `!available` cases, to keep things readable.

andyatmiami (approved):
/lgtm
/approve

--- 

@richabanker - been a **real run ride** getting to this point - but we are finally ready to merge this code!!  Appreciate all your engagement/responsiveness as we figured out this feature and its code structure.  Also, thanks for bringing your k8s knowledge/expertise as it relates to Metrics Server to this community.

very valuable - look forward to continuing to collaborate and build great things!

---

testing methodology:
- [PR-1312-TEST-RESULTS.md](https://github.com/user-attachments/files/31554060/PR-1312-TEST-RESULTS.md)

PR #1314, opened 2026-08-10
chore: regenerate frontend API client

christian-heusel:
Regenerating also pulls in the backend API drift that accumulated since the old pin: `pendingRestart` is gone from the workspace list item, `podMetadata` and `volumes` moved off `podTemplate` onto the `podtemplate/details` endpoint, `displayName` was added, and the shared `common.Pod*` definitions are now `details.Pod*`. No application code referenced the removed fields, so the fallout is limited to the mock builders and the Cypress specs that seeded workspace list items with volumes or secrets - that data is already provided to those tests through the workspace update and details mocks.

---

Split out from https://github.com/kubeflow/notebooks/pull/1310


thaorell (approved):
/lgtm

paulovmr (approved):
/lgtm
/approve

Issue #846, opened 2026-01-27
[TASK] Implement Compatibility Selectors /listvalues API endpoint


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

🛑 **Depends on:** #690 (filterRules CRD schema must be merged first)

Evolves the stub `POST /listvalues` endpoint into a full implementation that evaluates
`spec.filterRules[]` from the WorkspaceKind CRD to compute `hidden` and `restrictions` for each
imageConfig and podConfig value in the response.

This is the core of the Compatibility Selectors feature — it implements the rule evaluation engine
that processes filterRules with firewall-style first-match-wins semantics.

#### Goals

- **Rule evaluation engine** (shared utility, also used by #847)
- **`hidden` merging** — final `hidden` = admin-set value OR `ui.hide` from first matching rule
- **`restrictions` computation** — `deny` and `denyMessage` from first matching `api.deny` rule
- **`api.hide` filtering** — omit items from response when `api.hide: true` effect matches
- **Response schema update** — add `restrictions` attribute to `ImageConfigValue` and `PodConfigValue`
- **Namespace label resolution** — fetch namespace labels from k8s API when `matchNamespace` conditions exist
- **Cross-option matching** — evaluate `matchImageConfig.selector` against imageConfig labels when
   filtering podConfigs (and vice versa), using the context from the request
- **Swagger regeneration** after model changes
- **Unit tests** covering all rule evaluation scenarios

#### Rule Evaluation Logic

Per the approved design, for each option value:

1. Iterate through `filterRules[]` in order (top-to-bottom)
2. For each rule, check if `scope` matches the value type being evaluated
3. For matching scope, evaluate ALL `match[]` conditions (AND logic):
   - `matchNamespace.selector` — evaluated against namespace labels (fetched from k8s API)
   - `matchImageConfig.selector` — evaluated against imageConfig value's `spawner.labels`
     (uses the selected imageConfig from request context, if provided)
   - `matchPodConfig.selector` — evaluated against podConfig value's `spawner.labels`
     (uses the selected podConfig from request context, if provided)
4. If ALL conditions match → apply `effect`:
   - `ui.hide: true` → merge with `hidden` (OR with admin-set value)
   - `api.hide: true` → omit from response
   - `api.deny: true` → set `restrictions.deny: true` and `restrictions.denyMessage`
[...]

Issue #847, opened 2026-01-27
[TASK] Implement Compatibility Selectors /workspacekinds API endpoint


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

🛑 **Depends on:** 
- #690 (filterRules CRD schema)
- #846 (rule evaluation engine must exist first)

Evolves the `GET /workspacekinds?namespaceFilter=` endpoint to evaluate `spec.filterRules[]` with
`scope: WORKSPACE_KIND` for each WorkspaceKind and include computed `hidden` and `restrictions`
attributes in the response.

This task reuses the rule evaluation engine built in #846.

#### Goals

1. **Evaluate WORKSPACE_KIND-scoped rules** — for each WorkspaceKind, process its filterRules where
   `scope: WORKSPACE_KIND`
2. **`hidden` merging** — final `hidden` = admin-set `spawner.hidden` OR `ui.hide` from first matching rule
3. **`restrictions` computation** — `deny` and `denyMessage` from first matching `api.deny` rule
4. **`api.hide` filtering** — omit entire WorkspaceKind from response when `api.hide: true` matches
5. **Response schema update** — add `restrictions` attribute to the WorkspaceKind list response model
6. **Only evaluate when `namespaceFilter` is provided** — when absent (admin/Joel view), return
   all WorkspaceKinds with `hidden` as admin-set value and `restrictions.deny: false`
7. **Swagger regeneration**
8. **Unit tests**

#### Evaluation Logic

When `namespaceFilter` query parameter is provided:

1. Fetch namespace labels from k8s API for the given namespace
2. For each WorkspaceKind:
   a. Iterate through its `filterRules[]` in order
   b. For rules with `scope: WORKSPACE_KIND`:
      - Evaluate `matchNamespace.selector` against fetched namespace labels
      - (`matchImageConfig` and `matchPodConfig` are not applicable for WORKSPACE_KIND scope)
   c. First matching rule → apply effect (ui.hide, api.hide, api.deny)
   d. No match → retain admin-set values
3. Return filtered list with `hidden` and `restrictions` per WorkspaceKind

When `namespaceFilter` is NOT provided:
- Do NOT evaluate filterRules (this is the admin/Joel listing mode)
- Return all WorkspaceKinds with `hidden` as admin-set value, `restrictions.deny: false`

#### Response Schema Changes

Add to WorkspaceKind list response model:

```go
[...]

PR #1315, opened 2026-08-10
feat: Implement filterRules evaluation for /listvalues API

gangli113:
closes: #846

Evolves the stub `POST /listvalues` endpoint into a full implementation that evaluates `spec.filterRules[]` from the WorkspaceKind CRD to compute `hidden` and `restrictions` for each imageConfig and podConfig value in the response. This is the core of the Compatibility Selectors feature — the firewall-style (first-match-wins) rule evaluation engine.

Implemented the following:

1. Add a standalone `filterrules` package with a shared, dependency-light `Evaluate` engine (first-match-wins semantics), so it can be reused by the upcoming `/workspacekinds` API (#847).
2. Rule evaluation per option value:
   - `hidden` merging — final `hidden` = admin-set value **OR** `ui.hide` from the first matching rule.
   - `restrictions` computation — `deny` + `denyMessage` from the first matching `api.deny` rule.
   - `api.hide` filtering — omit items from the response when an `api.hide: true` effect matches.
3. Match condition evaluation (AND logic) using standard Kubernetes label selectors (`metav1.LabelSelectorAsSelector`):
   - `matchNamespace.selector` — against namespace labels fetched once per request from the k8s API.
   - `matchImageConfig.selector` / `matchPodConfig.selector` — against the option value's `spawner.labels`.
4. Cross-option matching — when `context.imageConfig.id` / `context.podConfig.id` is provided, the selected option's labels are resolved and used to evaluate cross-scope conditions (e.g. selecting an NVIDIA image hides non-GPU podConfigs).
5. Conservative semantics for absent context — when the required labels are not available (e.g. no `context.namespace.name`), the condition is treated as non-matching, so nothing is hidden/denied without full context.
6. Unit tests covering all rule evaluation scenarios: namespace matching, cross-option matching, `api.hide` omission, `api.deny` + `denyMessage`, `hidden` merging, first-match-wins ordering, and the no-rules default (100% statement coverage of the engine).

Notes:

- The response schema already included the `restrictions` attribute (added in a prior PR), so no swagger regeneration was required.
- `WORKSPACE_KIND`-scoped rules are intentionally not evaluated here — they belong to the `/workspacekinds` endpoint (#847). The engine is generic and already supports that scope.

### DIFF
--- a/workspaces/backend/api/workspacekind_podtemplate_options_handler_test.go
+++ b/workspaces/backend/api/workspacekind_podtemplate_options_handler_test.go
@@ -129,7 +129,7 @@ var _ = Describe("WorkspaceKinds Handler", func() {
 				wsk := &kubefloworgv1beta1.WorkspaceKind{}
 				Expect(k8sClient.Get(ctx, workspaceKind1Key, wsk)).To(Succeed())
 				listValuesRequest := &models.ListValuesRequest{Context: listValuesContext}
-				wskModel, err := models.NewPodTemplateOptionsModelFromWorkspaceKind(wsk, listValuesRequest)
+				wskModel, err := models.NewPodTemplateOptionsModelFromWorkspaceKind(wsk, listValuesRequest, nil)
 				Expect(err).NotTo(HaveOccurred())
 				return wskModel
 			}
--- a/workspaces/backend/api/workspacekind_podtemplate_options_handler_test.go
+++ b/workspaces/backend/api/workspacekind_podtemplate_options_handler_test.go
@@ -128,8 +128,18 @@ var _ = Describe("WorkspaceKinds Handler", func() {
 			getExpected := func(listValuesContext models.ListValuesContext) *models.PodTemplateOptions {
 				wsk := &kubefloworgv1beta1.WorkspaceKind{}
 				Expect(k8sClient.Get(ctx, workspaceKind1Key, wsk)).To(Succeed())
+
+				// resolve the namespace labels the same way the handler does, so the expected model
+				// reflects any matchNamespace rules rather than assuming an absent namespace context.
+				var namespaceLabels map[string]string
+				if listValuesContext.Namespace != nil {
+					ns := &corev1.Namespace{}
+					Expect(k8sClient.Get(ctx, types.NamespacedName{Name: listValuesContext.Namespace.Name}, ns)).To(Succeed())
+					namespaceLabels = ns.Labels
+				}
+
 				listValuesRequest := &models.ListValuesRequest{Context: listValuesContext}
-				wskModel, err := models.NewPodTemplateOptionsModelFromWorkspaceKind(wsk, listValuesRequest, nil)
+				wskModel, err := models.NewPodTemplateOptionsModelFromWorkspaceKind(wsk, listValuesRequest, namespaceLabels)
 				Expect(err).NotTo(HaveOccurred())
 				return wskModel
 			}
--- a/workspaces/backend/internal/filterrules/engine.go
+++ b/workspaces/backend/internal/filterrules/engine.go
@@ -1,0 +1,242 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package filterrules implements the shared, firewall-style (first-match-wins)
+// evaluation engine for WorkspaceKind `spec.filterRules[]`. It is intentionally
+// dependency-light (pure in-memory label matching plus the CRD types) so it can
+// be reused by both the `/listvalues` (#846) and `/workspacekinds` (#847) APIs.
+package filterrules
+
+import (
+	kubefloworgv1beta1 "github.com/kubeflow/notebooks/workspaces/controller/api/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+
+	"github.com/kubeflow/notebooks/workspaces/backend/internal/models/common"
+)
+
+// EvalTarget identifies the single value being evaluated and supplies its own labels.
+type EvalTarget struct {
+	// Scope is the type of value being evaluated (IMAGE_CONFIG, POD_CONFIG, or WORKSPACE_KIND).
+	Scope kubefloworgv1beta1.FilterRuleScope
+
+	// Labels are this value's own `spawner.labels`, used for same-scope match conditions.
+	Labels []kubefloworgv1beta1.OptionSpawnerLabel
+}
+
+// EvalContext holds the request-scoped inputs shared across all evaluations.
+//
+// A nil label map means the corresponding context was NOT provided in the request, so any
+// match condition that requires it is treated as non-matching (conservative: don't hide or
+// deny without full context). A non-nil but empty map means the context was provided but
+// carries no labels.
+type EvalContext struct {
+	// NamespaceLabels are the labels of the namespace the workspace would be created in.
+	// nil when `context.namespace.name` was absent from the request.
+	NamespaceLabels map[string]string
+
+	// ImageConfigLabels are the `spawner.labels` of the imageConfig value selected in the
+	// request. nil when `context.imageConfig.id` was absent. Used for cross-option matching
+	// on non-IMAGE_CONFIG scope rules.
+	ImageConfigLabels map[string]string
+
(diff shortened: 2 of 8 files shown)


workspaces/backend/internal/repositories/workspacekinds/repo.go
```diff
@@ -188,15 +188,45 @@ func (r *WorkspaceKindRepository) ListPodTemplateOptionsValues(ctx context.Conte
...
+		}
+		return nil, err
+	}
+
+	// always return a non-nil map so matchNamespace conditions are evaluated (namespace present)
+	if namespace.Labels == nil {
+		return map[string]string{}, nil
+	}
+	return namespace.Labels, nil
+}
```
andyatmiami: `r.client.Get` retrieves the namespace from the controller-runtime informer cache. Returning `namespace.Labels` directly exposes a reference to the cached object's label map. While current consumers only read the map, any future consumer that writes to it would silently corrupt the cached object. Other map extraction points in this PR (e.g., `spawnerLabelsToMap`) correctly create defensive copies.

Also - I think we can streamline this implementation a bit (drawing inspiration from #1279)

```
result := make(map[string]string, len(namespace.Labels))
maps.Copy(result, namespace.Labels)
return result, nil
```

workspaces/backend/api/workspacekind_podtemplate_options_handler_test.go
```diff
@@ -129,7 +129,7 @@ var _ = Describe("WorkspaceKinds Handler", func() {
 				wsk := &kubefloworgv1beta1.WorkspaceKind{}
 				Expect(k8sClient.Get(ctx, workspaceKind1Key, wsk)).To(Succeed())
 				listValuesRequest := &models.ListValuesRequest{Context: listValuesContext}
-				wskModel, err := models.NewPodTemplateOptionsModelFromWorkspaceKind(wsk, listValuesRequest)
+				wskModel, err := models.NewPodTemplateOptionsModelFromWorkspaceKind(wsk, listValuesRequest, nil)
```
andyatmiami: While probably safe today - i worry this `nil` argument reads a bit like a "magic string" and implicitly requires that any text fixture is **NOT** defining any `namespaceLabels` - as they will never be resolvable (as we hardcode `nil` to bypass any real processing)

Not necessarily asking for a change here... as offhand I'm not 100% sure how to better handle this - but at least wanted to call it out for your consideration...

workspaces/backend/internal/filterrules/engine_test.go
```diff
@@ -0,0 +1,526 @@
...
+	Context("when no rules are provided", func() {
+		It("returns the non-restrictive default", func() {
+			result := Evaluate(nil, EvalTarget{
+				Scope:  kubefloworgv1beta1.FilterRuleScopeImageConfig,
+				Labels: spawnerLabels(map[string]string{"gpu": "true"}),
+			}, EvalContext{})
+
+			Expect(result.UIHide).To(BeFalse())
+			Expect(result.APIHide).To(BeFalse())
+			Expect(result.Restrictions).To(Equal(common.DefaultRestrictions()))
```
andyatmiami: `BeComparableTo` might be a little more readable/maintainable/consistent

workspaces/backend/internal/filterrules/engine.go
```diff
@@ -0,0 +1,242 @@
...
+//
+// When targetLabels is nil the required context is absent, so the condition is non-matching
+// regardless of the selector (conservative: don't hide or deny without full context). An
+// invalid selector (which the CRD webhook should already reject) is also non-matching.
+func matchSelector(labelSelector metav1.LabelSelector, targetLabels map[string]string) bool {
+	if targetLabels == nil {
+		return false
+	}
+
+	selector, err := metav1.LabelSelectorAsSelector(&labelSelector)
```
andyatmiami: While under a slightly different context - we did end up adding a label selector cache in #1279 ... I wonder if that idea is also warranted here..

More an open question to consider/go with your gut for now - as this is way less of a "hot path" as compared to the reconcile loop in the referenced PR.

andyatmiami: Instead of a long-lived cache, you could compile all selectors upfront:

```
// compiledRule holds a pre-parsed rule with its selectors already compiled.
type compiledRule struct {
    rule      *kubefloworgv1beta1.FilterRule
    selectors []compiledMatch
}

type compiledMatch struct {
    match    *kubefloworgv1beta1.FilterRuleMatch
    selector labels.Selector // compiled once
}
```

Then split evaluation into two phases:

1. CompileRules(rules) — called once per request, compiles all LabelSelectors into labels.Selectors, skips invalid ones
2. Evaluate(compiledRules, target, evalCtx) — called per value, only does selector.Matches() (cheap)

workspaces/backend/internal/models/workspacekinds/podtemplate/options/types_write.go
```diff
@@ -37,6 +37,18 @@ func (d *ListValuesRequest) Validate(prefix *field.Path) []*field.Error {
...
+func (d *ListValuesRequest) ConfigIDs() (imageConfigID, podConfigID string) {
+	if d.Context.ImageConfig != nil {
+		imageConfigID = d.Context.ImageConfig.Id
+	}
+	if d.Context.PodConfig != nil {
+		podConfigID = d.Context.PodConfig.Id
+	}
+	return imageConfigID, podConfigID
+}
+
```
andyatmiami: I wonder if this could be a private helper and/or inlined in the one call site we have that leverages this (exported) function?

I'm not sure (although I could certainly be wrong) that we'd really have a valid case of this being called in such a manner to warrant this being exportable.

WDYT?

workspaces/backend/internal/filterrules/engine.go
```diff
@@ -0,0 +1,242 @@
...
+			if effect.API.DenyMessage != nil {
+				result.Restrictions.DenyMessage = &common.DenyMessage{
+					Text: effect.API.DenyMessage.Text,
+				}
+			}
+		}
+	}
+
+	return result
+}
```
andyatmiami, line 294: The codebase convention is to return the struct literal directly, extracting any complex logic into helpers:

```
func resultFromEffect(effect kubefloworgv1beta1.FilterRuleEffect) EvalResult {
      return EvalResult{
              UIHide:       resolveUIHide(effect),
              APIHide:      resolveAPIHide(effect),
              Restrictions: resolveRestrictions(effect),
      }
}
```
This way the struct shape is immediately visible — you can see at a glance what an EvalResult is made of, and each helper is a self-contained pure function. The mutation-style version obscures the structure behind a sequence of conditional assignments, and makes it harder to reason about which fields
are set under which conditions.

ℹ️ Was just using the code snippet above as an example.. in this context we can probably inline some logic and/or have helper logic defined before the return (i.e. maybe an unexported function is NOT necessary - will leave for you to decide)

andyatmiami (changes requested):
@gangli113 

Another really exciting and solid contribution... mostly syntactical/style feedback - and an open question or two you can think on and decide if you want to _take action_ .

let me know if you have any questions - we are **close** to be able to getting this in

👍

workspaces/backend/api/workspacekind_podtemplate_options_handler_test.go
gangli113: Added code to get namespaceLabels, and passed the labels into the function

workspaces/backend/internal/filterrules/engine.go
gangli113: Good idea. I just followed you suggestion split evaluation into two phase. CompileRules and Evaluate

compileRules is called by BuildEvalContext() once per request.

workspaces/backend/internal/filterrules/engine.go
gangli113: refactored resultFromEffect, and it's much cleaner now. Thank you for the suggestion.

workspaces/backend/internal/models/workspacekinds/podtemplate/options/types_write.go
gangli113: I made configIDs to be a private helper, it's pretty straight forward since types_write.go and funcs.go belong to the same package, `option`.

workspaces/backend/internal/repositories/workspacekinds/repo.go
gangli113: I added copy map at the end of the function. 

Quick question: Should `client.get()` perform deepcopy by default? I was looking at [this article](https://kubernetes.io/blog/2026/07/29/controller-runtime-cache-explained/#mistake-2-deepcopy-and-who-owns-the-memory). Wondering if it applies notebooks' setup as well

workspaces/backend/internal/filterrules/engine_test.go
gangli113: Done. use BeComparableTo for all result assertions

andyatmiami (approved):
/lgtm
/approve

---

@gangli113 - excellent contribution here and I appreciate how responsive you were to my review comments.

Excited to have this functionality available in the codebase 💯 Look forward to seeing what other areas you can jump into and continue to make profound improvements to the project!

testing/verification notes:
- [PR-1315-TEST-RESULTS.md](https://github.com/user-attachments/files/31243041/PR-1315-TEST-RESULTS.md)

Issue #1209, opened 2026-06-21
[TASK] Filter Image and Pod Config options by `hidden` and display `restrictions`


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

🛑 **Depends on:** 
- #846 (Backend must return `hidden` and `restrictions` on option values)

Updates the Image Config and Pod Config wizard steps to:
- Filter out options where `hidden: true`
- Show options where `restrictions.deny: true` as disabled with the `denyMessage.text` displayed
- Handle the case where the currently selected option becomes hidden or denied after a context change
  (e.g., user goes back and picks a different image, which changes which podConfigs are visible)

#### Goals

- Filter `imageConfig.values` and `podConfig.values` by `hidden`
- Display restricted options with disabled state and explanation
- Handle "default option is hidden/denied" gracefully (select next available, or show message)
- Handle "currently selected option is no longer visible" when context changes
- Works for both Create and Edit Workspace flows
- Unit tests and Cypress tests with rule-evaluated mock responses

### Acceptance Criteria

- [ ] Image config options with `hidden: true` are not shown
- [ ] Pod config options with `hidden: true` are not shown
- [ ] Options with `restrictions.deny: true` are shown but disabled with explanation text
- [ ] When the default option is hidden, the next available option is pre-selected (or no default)
- [ ] When the user changes image selection and re-fetches /listvalues, the pod config list updates
- [ ] Edit Workspace flow correctly displays restrictions for current option values
- [ ] Cypress tests verify filtering with mocked rule-evaluated responses

Issue #735, opened 2025-11-10
[FEATURE] Frontend for compatibility selectors


paulovmr:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Motivation

As a data scientist, Bella should see, by default, only compatible resources that she has access to while creating or editing a workspace. 

### Acceptance Criteria

- [ ] See only accessible Workspace Kinds in the workspace creation Wokspace Kind step
- [ ] See only accessible and compatible images in the workspace creation image step
- [ ] See only accessible and compatible pod configs in the workspace creation pod config step

thesuperzapper:
There is no task yet, but the frontend will need to intergrate correctly with backend LIST apis which will be updated with `restrictions` and `hidden` sections.

PR #1325, opened 2026-08-13
fix: filter hidden options and display restrictions popover

Snehadas2005:
Fixes issue where restricted or hidden options were not properly evaluated or displayed in the Workspace creation/edit wizard.
This PR implements rule-based option filtering and restriction handling for both **Image Config** and **Pod Config** selection steps:
1. Filters out options where `hidden: true` from the available selection lists.
2. Renders options with `restrictions.deny: true` as disabled (`isDisabled`) and non-selectable, displaying a `BanIcon` with a hover popover containing the restriction message (`denyMessage.text`).
3. Resolves usable default options upon kind selection; if a default option is hidden or denied, the form pre-selects the next available usable option.
4. Preserves currently selected pod configs during context changes if they exist and are not hidden (even if denied), so users can see why their current selection is restricted.

**Key Changes:**
- **`WorkspaceFormOptionCard.tsx`**: Updated option card to check `restrictions.deny`, apply disabled styling (`workspace-option-card--restricted`), block card click interactions, and render `RestrictedIconWithPopover`.
- **`RestrictedIconWithPopover.tsx`**: Added PatternFly `Popover` wrapper with `BanIcon` to show restriction explanation text on hover.
- **`WorkspaceFormImageList.tsx` & `WorkspaceFormPodConfigList.tsx`**: Added defensive check in option click/change handlers to prevent selection of denied options.
- **`WorkspaceForm.tsx`**: Added `resolveUsableDefault` helper to resolve usable (non-hidden, non-denied) default options for `imageConfig` and `podConfig`. Updated podConfig re-validation effect to leave denied-but-present options selected for clear UI feedback.
- **`testBuilders.ts`**: Added `buildMockImageWithRestrictions` test helper for unit and Cypress tests.

**Screenshots / UI Verification:**

Screenshot 2026-08-13 235057

**How to Test:**
1. Run local dev server: `npm run start:dev`
2. Navigate to Create Workspace form -> Image & Pod Config steps.
3. Verify options with `hidden: true` do not appear.
4. Verify options with `restrictions.deny: true` are disabled and hovering over the icon displays the restriction text.
5. Run TypeScript check & unit tests:
   ```bash
   npm run test:type-check
   npm run test:unit

closes: #1209
related: #735

### DIFF
--- a/workspaces/frontend/src/__tests__/cypress/cypress/utils/testBuilders.ts
+++ b/workspaces/frontend/src/__tests__/cypress/cypress/utils/testBuilders.ts
@@ -182,3 +182,20 @@ export const interceptListValues = (
     })
     .as(`listValues-${kind.name}`);
 };
+
+export const buildMockImageWithRestrictions = (
+  id: string,
+  displayName: string,
+  deny: boolean,
+  denyMessageText?: string,
+): OptionsImageConfigValue => ({
+  id,
+  displayName,
+  description: `Image: ${displayName}`,
+  hidden: false,
+  redirect: undefined,
+  clusterMetrics: undefined,
+  restrictions: deny
+    ? { deny: true, denyMessage: { text: denyMessageText ?? '' } }
+    : { deny: false },
+});
--- a/workspaces/frontend/src/app/components/RestrictedIconWithPopover.tsx
+++ b/workspaces/frontend/src/app/components/RestrictedIconWithPopover.tsx
@@ -1,0 +1,17 @@
+import * as React from 'react';
+import { Popover } from '@patternfly/react-core/dist/esm/components/Popover';
+import { BanIcon } from '@patternfly/react-icons/dist/esm/icons/ban-icon';
+
+interface RestrictedIconWithPopoverProps {
+  id: string;
+  message: string;
+}
+
+export const RestrictedIconWithPopover: React.FC<RestrictedIconWithPopoverProps> = ({
+  id,
+  message,
+}) => (
+  <Popover id={id} bodyContent={message} triggerAction="hover">
+    <BanIcon aria-label="Restricted option" className="workspace-option-card__restricted-icon" />
+  </Popover>
+);
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
@@ -119,6 +119,21 @@ const WorkspaceForm: React.FC = () => {
     // Store original values for diff comparison
     setOriginalData(initialFormData);
   }, [initialFormData, initialFormDataLoaded, mode, replaceData]);
+  function resolveUsableDefault<
+    T extends { id: string; hidden?: boolean; restrictions?: { deny?: boolean } },
+  >(values: T[] | undefined, defaultId: string | undefined): string | undefined {
+    if (!values) {
+      return undefined;
+    }
+    const defaultOption = values.find((v) => v.id === defaultId);
+    if (defaultOption && !defaultOption.restrictions?.deny) {
+      return defaultId;
+    }
+    const isUsable = (opt?: T) => !opt?.restrictions?.deny;
+    const fallback = values.find((v) => isUsable(v));
+
+    return fallback?.id;
+  }
 
   // Apply default imageConfig and podConfig from listValues when a kind is first selected.
   // Only sets defaults when the values are unset (undefined), so user selections are never overwritten.
@@ -127,10 +142,22 @@ const WorkspaceForm: React.FC = () => {
       return;
     }
     if (!data.imageConfig && allValuesData.imageConfig.default) {
-      setData('imageConfig', allValuesData.imageConfig.default);
+      const resolved = resolveUsableDefault(
+        allValuesData.imageConfig.values,
+        allValuesData.imageConfig.default,
+      );
+      if (resolved) {
+        setData('imageConfig', resolved);
+      }
     }
     if (!data.podConfig && allValuesData.podConfig.default) {
-      setData('podConfig', allValuesData.podConfig.default);
+      const resolved = resolveUsableDefault(
+        allValuesData.podConfig.values,
+        allValuesData.podConfig.default,
+      );
+      if (resolved) {
+        setData('podConfig', resolved);
+      }
     }
   }, [allValuesData, allValuesLoaded, data.kind, data.imageConfig, data.podConfig, setData]);
 
@@ -140,7 +167,9 @@ const WorkspaceForm: React.FC = () => {
       return;
     }
     const podConfigOptions = filteredValuesData.podConfig.values ?? [];
-    const isStillValid = podConfigOptions.some((pc) => pc.id === data.podConfig);
+    const current = podConfigOptions.find((pc) => pc.id === data.podConfig);
+    // denied-but-present is left alone on purpose; hidden options are cleared
+    const isStillValid = !!current;
     if (!isStillValid) {
       setData('podConfig', undefined);
     }
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/image/WorkspaceFormImageList.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/image/WorkspaceFormImageList.tsx
@@ -55,6 +55,9 @@ export const WorkspaceFormImageList: React.FunctionComponent<WorkspaceFormImageL
       const newSelectedWorkspaceImage = filteredWorkspaceImages.find(
         (image) => image.displayName === event.currentTarget.name,
       );
+      if (newSelectedWorkspaceImage?.restrictions.deny) {
+        return;
+      }
       onSelect(newSelectedWorkspaceImage);
     },
     [filteredWorkspaceImages, onSelect],
@@ -62,12 +65,12 @@ export const WorkspaceFormImageList: React.FunctionComponent<WorkspaceFormImageL
 
   const handleCardClick = useCallback(
     (image: OptionsImageConfigValue) => {
-      if (image.id !== selectedImage?.id) {
+      if (image.restrictions.deny) {
         return;
       }
       onSelect(image);
     },
-    [selectedImage, onSelect],
+    [onSelect],
   );
 
   return (
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/podConfig/WorkspaceFormPodConfigList.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/podConfig/WorkspaceFormPodConfigList.tsx
@@ -58,12 +58,12 @@ export const WorkspaceFormPodConfigList: React.FunctionComponent<
 
   const handleCardClick = useCallback(
     (podConfig: OptionsPodConfigValue) => {
-      if (podConfig.id !== selectedPodConfig?.id) {
+      if (podConfig.restrictions.deny) {
         return;
       }
       onSelect(podConfig);
     },
-    [selectedPodConfig?.id, onSelect],
+    [onSelect],
   );
 
   return (
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/shared/WorkspaceFormOptionCard.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/shared/WorkspaceFormOptionCard.tsx
@@ -7,8 +7,10 @@ import {
 } from '@patternfly/react-core/dist/esm/components/Card';
 import { Label } from '@patternfly/react-core/dist/esm/components/Label';
 import { Flex, FlexItem } from '@patternfly/react-core/dist/esm/layouts/Flex';
+import { css } from '@patternfly/react-styles';
 import { HiddenIconWithPopover } from '~/app/components/HiddenIconWithPopover';
 import { RedirectIconWithPopover } from '~/app/components/RedirectIconWithPopover';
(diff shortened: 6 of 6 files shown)


christian-heusel (changes requested):
Please fix the CI errors 😊  Also the frontent portion of the code contains husky pre-commit hooks, maybe enable that so that this verification runs automatically on commit (locally) 🤗 

Edit: also it seems like you need to rebase 🤗

Snehadas2005:
Thank you for the reminder, @christian-heusel! It would be very helpful if you could review the codebase and identify any necessary changes.

workspaces/frontend/package.json
```diff
@@ -72,7 +72,7 @@
     "chai-subset": "^1.6.0",
     "concurrently": "^9.1.0",
     "copy-webpack-plugin": "^13.0.0",
-    "core-js": "^3.40.0",
+    "core-js": "^3.40.0",
```
christian-heusel: Please remove this extra whitespace change 😅

workspaces/frontend/src/__tests__/cypress/cypress/support/commands/api.ts
```diff
@@ -95,6 +90,11 @@ declare global {
           },
           response: string | ApiErrorEnvelope,
         ) => Cypress.Chainable<null>) &
+        ((
+          type: 'POST /api/:apiVersion/workspaces/:namespace/:workspaceName/actions/pause',
+          options: { path: { apiVersion: string; namespace: string; workspaceName: string } },
+          response: ApiWorkspaceActionPauseEnvelope | ApiErrorEnvelope,
+        ) => Cypress.Chainable<null>) &
```
christian-heusel: I think you moved this on accident without changing it 🤔 Given that it is unchanged, please remove that diff from the PR 🤗

workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
```diff
@@ -117,6 +117,21 @@ const WorkspaceForm: React.FC = () => {
...
+  function resolveUsableDefault<
+    T extends { id: string; hidden?: boolean; restrictions?: { deny?: boolean } },
+  >(values: T[] | undefined, defaultId: string | undefined): string | undefined {
+    if (!values) {
+      return undefined;
+    }
+    const defaultOption = values.find((v) => v.id === defaultId);
+    if (defaultOption && !defaultOption.restrictions?.deny) {
+      return defaultId;
+    }
```
christian-heusel, line 131: We should also check here whether the option is hidden (and not only for `.deny`)
Maybe you could just move the `isUsable` up a bit and check for `defaultOption && isUsable`? 🤔

workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
```diff
@@ -138,7 +165,8 @@ const WorkspaceForm: React.FC = () => {
       return;
     }
     const podConfigOptions = filteredValuesData.podConfig.values ?? [];
-    const isStillValid = podConfigOptions.some((pc) => pc.id === data.podConfig);
+    const current = podConfigOptions.find((pc) => pc.id === data.podConfig);
+    const isStillValid = !!current; // denied-but-present is left alone on purpose
```
christian-heusel: Should we also check for `hidden` here? 🤔 

<details>
<summary> Show more detailed finding (backed by 🤖, so treat with caution):</summary>

The `podConfig` re-validation effect only checks that the current selection still exists in the list, not whether it's become hidden. A `podConfig` that flips to `hidden: true` (but isn't denied) stays selected, contradicting the PR description's rule to clear hidden (even non-denied) selections.

The bug: `isStillValid` only checks that the id is still present in the returned list (`!!current`). It never looks at `current.hidden`. A `podConfig` option that becomes incompatible with the newly selected image but is represented as hidden: true (rather than removed from the array or `restrictions.deny: true`) still passes `isStillValid`, so the selection is silently kept.

**Concrete failure scenario:**
1. User selects image A, then picks `podConfig` "Large" — visible, not hidden, not denied for image A.
2. User goes back and switches the image to B.
3. filteredValuesData refetches with imageId: B. The backend marks "Large" as hidden: true for this image (still present in the array, not denyd).
4. The effect finds current (id still in the list) → `isStillValid = true` → `data.podConfig` is left as "Large".
5. The pod-config picker (which presumably filters out hidden options from the visible list, same as the image picker) no longer shows "Large" as selectable/visible, but the form state still carries it forward to Summary and submission — the user has a selection they can't see, can't consciously confirm, and can't change without noticing something's off.

This contradicts the PR's own stated rule (per the review) that a selection should be preserved only "if they exist and are not hidden (even if denied)" — i.e., hidden should clear it, but a merely-denied-and-visible option should not.

Fix: add the hidden check, keeping the deny exemption the comment describes:

```ts
const isStillValid = !!current && !current.hidden; // denied-but-present is left alone on purpose
```

</details>

workspaces/frontend/src/app/hooks/useWorkspacePodTemplateDetails.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    const envelope = await api.workspaces.getWorkspacePodTemplateDetails(namespace, name);
+    return envelope.data;
+  }, [api, apiAvailable, namespace, name]);
+
+  return useFetchState(call, null);
+};
+
+export default useWorkspacePodTemplateDetails;
```
christian-heusel: This duplicates the existing hook in `useWorkspaceDetails.ts` (🤖 -finding):

Both hooks wrap the exact same backend call, `api.workspaces.getWorkspacePodTemplateDetails(namespace, name)`:

- `workspaces/frontend/src/app/hooks/useWorkspaceDetails.ts`
- `workspaces/frontend/src/app/hooks/useWorkspacePodTemplateDetails.ts`

`WorkspaceDetails.tsx` (and its test) were switched over to `useWorkspacePodTemplateDetails` in this PR, but `useWorkspaceDetails` was left in the tree instead of being reused/renamed. A repo-wide search shows it now has zero callers outside its own file and its own spec test (`useWorkspaceDetails.spec.tsx`) hence it's now orphaned.

The two hooks differ only slightly (return-type wrapper: `DetailsWorkspaceDetails | null` vs `ApiWorkspaceDetailsEnvelope['data'] | null`, which are structurally the same; missing-args behavior: rejects with `NotReadyError` vs returns `null`; named vs default export), so this reads like a copy made to fit the new call site rather than an intentional second hook.

**Suggestion:** delete `useWorkspaceDetails.ts` + `useWorkspaceDetails.spec.tsx` (if the null-on-missing-args behavior of `useWorkspacePodTemplateDetails` is acceptable everywhere), or fold the two into a single hook and update the one call site. Otherwise this leaves two near-identically-named hooks for the same data in the codebase, which will confuse future readers about which one is actually live.

workspaces/frontend/src/__tests__/cypress/cypress/support/commands/api.ts
```diff
@@ -169,6 +169,11 @@ declare global {
           type: 'GET /api/:apiVersion/secrets/:namespace/:secretName',
           options: { path: { apiVersion: string; namespace: string; secretName: string } },
           response: ApiSecretEnvelope | ApiErrorEnvelope,
+        ) => Cypress.Chainable<null>) &
+        ((
+          type: 'GET /api/:apiVersion/workspaces/:namespace/:workspaceName/podtemplate/details',
+          options: { path: { apiVersion: string; namespace: string; workspaceName: string } },
+          response: ApiWorkspaceDetailsEnvelope | ApiErrorEnvelope,
         ) => Cypress.Chainable<null>);
```
christian-heusel: `GET /api/:apiVersion/workspaces/:namespace/:workspaceName/podtemplate/details` is declared twice in `workspaces/frontend/src/__tests__/cypress/cypress/support/commands/api.ts`.

The overload already exists :

https://github.com/kubeflow/notebooks/blob/07a611c3884ab205544bd3815776ce6399b36608/workspaces/frontend/src/__tests__/cypress/cypress/support/commands/api.ts#L84-L87

This PR adds a second, byte-identical copy of the same overload near the end of the file (around line 174), right after the `secrets/:namespace/:secretName` entry:

```ts
((
  type: 'GET /api/:apiVersion/secrets/:namespace/:secretName',
  options: { path: { apiVersion: string; namespace: string; secretName: string } },
  response: ApiSecretEnvelope | ApiErrorEnvelope,
) => Cypress.Chainable<null>) &
((
  type: 'GET /api/:apiVersion/workspaces/:namespace/:workspaceName/podtemplate/details',
  options: { path: { apiVersion: string; namespace: string; workspaceName: string } },
  response: ApiWorkspaceDetailsEnvelope | ApiErrorEnvelope,
) => Cypress.Chainable<null>);
```

Both declarations are identical (same type string, same options shape, same response type), so the second one is dead weight in the `cy.mockApi()` overload chain — it can never be selected over the first and only adds noise for anyone reading or extending this file.

**Suggestion:** drop the second copy at the end of the file; the existing declaration already covers this route
```suggestion
```

christian-heusel (changes requested):
Hey, thanks a lot for working on this change! 🤗 

I found a few things that require changing, although I need to defer to the actual frontend maintainers for a proper review though 😅

christian-heusel (changes requested):
This PR also seems to mess with the sizing in the create workflow, i.e. when selecting the GPU podOption this now resizes things:

**Before (`notebooks-v2` checkout at 07a611c3884ab205544bd3815776ce6399b36608):**

Image

**After:**

Image

Since this expansion only happens for selected items this is rather surprising for the user 🤔

Snehadas2005:
Thank you for the review, @christian-heusel! I have updated the PR and addressed all the cleanup points:
* **Cleanup & Formatting:** Removed the accidental whitespace in `package.json`, reverted the extra `POST /pause` diff in `api.ts`, and cleaned up duplicate type overloads.
* **Refactoring & Visual Fix:** Removed the orphaned `useWorkspaceDetails.ts` hook (and spec), and restored `isCompact` on `<Card>` in `WorkspaceFormOptionCard.tsx` to fix the GPU card selection expansion.
* **Form Logic & Test Spec Contract:** While testing `resolveUsableDefault` and the `podConfig` revalidation effect, I verified against the Cypress test suite (`optionCardDisplay.cy.ts` and `createWorkspace.cy.ts`). The test contracts explicitly expect `hidden` options to remain selectable and eligible as defaults (showing the hidden badge/confirmation modal UX), whereas `.deny` acts as the hard block. So I kept the revalidation and default-resolution gated on `.deny` to preserve the intended modal workflow.

Everything is green locally (`tsc`, `eslint`, and all Jest/Cypress specs passing). Ready for review whenever you have a moment!

/cc @thaorell for frontend review whenever you get a chance.

workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
Snehadas2005: Thank you for your feedback on this! I intentionally kept it set to deny only, as outlined in the Cypress contracts for `optionCardDisplay.cy.ts` and `createWorkspace.cy.ts`. The expectation is that hidden options remain selectable and eligible by default, as they are indicated by the `hidden` badge and associated confirmation modal. Introducing a hidden check would disrupt this workflow. However, I am open to adapting if you see merit in revising the test contract. Happy to have your thoughts on this one. 😊

workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
```diff
@@ -119,6 +119,21 @@ const WorkspaceForm: React.FC = () => {
     // Store original values for diff comparison
     setOriginalData(initialFormData);
   }, [initialFormData, initialFormDataLoaded, mode, replaceData]);
+  function resolveUsableDefault<
```
thaorell, line 122: `resolveUsableDefault` only skips the default when `restrictions.deny` is true — it never checks `hidden`. This contradicts both the PR description ("if a default option is hidden or denied, the form pre-selects the next available usable option") and issue #1209's acceptance criterion that a hidden default should fall through to the next available option. As written, a hidden default is applied directly to `data.imageConfig`/`data.podConfig`, which the `optionCardDisplay.cy.ts` test also asserts as the expected behavior — so the code, the PR description, and the test suite currently disagree on what "usable default" means.

workspaces/frontend/src/__tests__/cypress/cypress/utils/testBuilders.ts
```diff
@@ -182,3 +182,20 @@ export const interceptListValues = (
     })
     .as(`listValues-${kind.name}`);
 };
+
```
thaorell, line 185: this function is not referenced anywhere. consider removing it

workspaces/frontend/src/app/components/RestrictedIconWithPopover.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
```
thaorell, line 1: Consider removing this component and adhering to the precedence set by `src/app/pages/Workspaces/Form/kind/WorkspaceFormKindList.tsx` for how to display restriction and hidden options, aka using <Tooltip/> (line 188 in `src/app/pages/Workspaces/Form/kind/WorkspaceFormKindList.tsx` )

thaorell (changes requested):

PR #1327, opened 2026-08-14
chore: Workspace table view adjustments

christian-heusel:
## Summary

This PR reworks the workspaces table (`WorkspaceTable.tsx`) based on feedback
from a recent walkthrough that we did in a Notebooks WG leads session with the
goal to have the UI be easier to navigate for the Data Scientist persona.

The PR covers the following broad changes:

- quicker navigation to the details view by clicking a workspace's name ➡️ Split out to #1331 
- an adjusted (perhaps more sensible) column order
- responsive sizing that degrades more gracefully on narrow viewports
- a redesigned **Connect** control that is now a split button ➡️ Split out to #1332 

_Note_: This is not so much about the actual implementation but more intended as
a conversation starter on what we like or don't like about these changes.

The plan is to then split out/adapt changes as needed so that we can include
what we would like to keep in the context of the beta.

# Screenshots

## Wide viewport

grafik

## Narrow

grafik

### DIFF
--- a/workspaces/frontend/src/app/app.css
+++ b/workspaces/frontend/src/app/app.css
@@ -37,6 +37,14 @@
   cursor: pointer;
 }
 
+/* Workspace table row name: looks like plain text until hovered/focused, then reveals link styling.
+   Selector specificity matches PatternFly's own `.pf-v6-c-button` base rule (which also sets these
+   custom properties) so this override wins regardless of CSS injection order. */
+.pf-v6-c-button.workspace-table__name-link {
+  --pf-v6-c-button--m-link--m-inline--Color: var(--pf-t--global--text--color--regular);
+  --pf-v6-c-button--m-link--m-inline--TextDecorationLine: none;
+}
+
 .summary-redirect-icon-button {
   cursor: pointer;
   display: inline-flex;
--- a/workspaces/frontend/src/app/app.css
+++ b/workspaces/frontend/src/app/app.css
@@ -45,6 +45,28 @@
   --pf-v6-c-button--m-link--m-inline--TextDecorationLine: none;
 }
 
+/* Name is the only workspace table column whose content is unbounded user input, so it is the only
+   one that can grow without limit. Capping it means a pathological name truncates instead of
+   pushing the whole table into a horizontal scroll.
+
+   The cap sits on this span — an element the app owns and renders itself — rather than on the cell
+   or the button. Because `40ch` is an absolute length, the span shrink-wraps and bounds the cell's
+   `min-width: fit-content` on its own, so the constraint does not have to be threaded through the
+   boxes in between. Styling our own element also keeps us clear of PatternFly's internals: its
+   Button renders children inside a `.pf-v6-c-button__text` span that takes no className prop, and
+   reaching for that class directly would break on any release that changes it.
+
+   These declarations are `pf-v6-u-text-truncate` minus its `max-width: 100%`, which cannot be used
+   here because it would override the cap. `white-space` is required because PatternFly's base
+   button rule sets `normal`, overriding the `nowrap` the cell inherits from `fitContent`. */
+.workspace-table__name-text {
+  display: block;
+  max-width: 40ch;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
 .summary-redirect-icon-button {
   cursor: pointer;
   display: inline-flex;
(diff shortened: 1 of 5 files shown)


workspaces/frontend/src/app/app.css
```diff
@@ -37,6 +37,36 @@
   cursor: pointer;
 }
 
+/* Workspace table row name: looks like plain text until hovered/focused, then reveals link styling.
+   Selector specificity matches PatternFly's own `.pf-v6-c-button` base rule (which also sets these
+   custom properties) so this override wins regardless of CSS injection order. */
+.pf-v6-c-button.workspace-table__name-link {
+  --pf-v6-c-button--m-link--m-inline--Color: var(--pf-t--global--text--color--regular);
+  --pf-v6-c-button--m-link--m-inline--TextDecorationLine: none;
+}
```
christian-heusel, line 46: This is most likely not how this should be done, any ideas for alternatives?

christian-heusel:
/assign @thesuperzapper 
for feedback
/cc @Snehadas2005 
for implementation ideas and review

workspaces/frontend/src/app/components/WorkspaceTable.tsx
```diff
@@ -349,21 +360,19 @@ const WorkspaceTable = React.forwardRef<WorkspaceTableRef, WorkspaceTableProps>(
...
                       <Th
                         key={`workspace-table-column-${columnKey}`}
-                        width={wsTableColumns[columnKey].width}
                         sort={specialProps.hasContent ? getSortParams(columnKey) : undefined}
                         aria-label={specialProps.hasContent ? columnKey : undefined}
-                        modifier={modifier}
+                        modifier="fitContent"
+                        visibility={columnVisibility[columnKey]}
                         screenReaderText={specialProps.screenReaderText}
                       >
```
Snehadas2005, line 377: Two things here that look like rebase leftovers rather than intentional:
1. `modifier` is set twice — `modifier={modifier}` from the old per-column logic, then `modifier="fitContent"` right after. Only the second one wins at runtime, and if `react/jsx-no-duplicate-props` is enabled this should fail lint. Can we drop the first one (and the now-unused `getColumnModifier` call above it) since `columnVisibility`/`fitContent` seems to be the new approach? 
2. `width={wsTableColumns[columnKey].width}` — `filterableDataHelper.ts` drops `width` from `DataFieldDefinition` in this same PR, so this reference shouldn't type-check. Does `npm run type-check` pass locally on this branch? Want to make sure I'm not misreading a diff artifact.

workspaces/frontend/src/app/pages/Workspaces/WorkspaceConnectAction.tsx
```diff
@@ -48,17 +68,26 @@ export const WorkspaceConnectAction: React.FunctionComponent<WorkspaceConnectAct
...
           isExpanded={open}
-          isDisabled={workspace.state !== V1Beta1WorkspaceState.WorkspaceStateRunning}
+          isDisabled={isDisabled}
           aria-label="Select connection endpoint"
-        >
-          Connect
-        </MenuToggle>
+          splitButtonItems={[
+            <MenuToggleAction
+              id={`${workspace.name}-connect`}
```
Snehadas2005, line 75: `id` here is derived from `workspace.name` alone, but names are unique within a namespace, not cluster-wide. If the table ever shows workspaces across namespaces (e.g. an all-namespaces view), two rows could end up with the same DOM `id`. Might be worth `${workspace.namespace}-${workspace.name}-connect` to be safe — not blocking for this draft, just flagging before it gets baked in.

workspaces/frontend/src/app/app.css
Snehadas2005: I believe that, rather than matching PatternFly's selector specificity (which is fragile if PF changes that base rule's specificity in a future release), what about setting the same custom properties inline via the `style` prop on the `<Button>` itself?

```jsx
<Button
   variant="link"
   isInline
   className="workspace-table__name-link"
   style={{
     '--pf-v6-c-button--m-link--m-inline--Color': 'var(--pf-t--global--text--color--regular)',
     '--pf-v6-c-button--m-link--m-inline--TextDecorationLine': 'none',
   }}
   ...
 ```
Inline styles win over external rules regardless of selector specificity, so it sidesteps the whole "match PF's specificity" problem, and it stays scoped to this one instance instead of a global rule any other `.pf-v6-c-button` could accidentally pick up.

Snehadas2005 (commented):
Hi @christian-heusel! I added some comments; I hope these will be helpful and will help you further with the implementation. 😊

christian-heusel:
Most relevant parts have been split out into other PRs and already merged, so this Proof of concept can be closed 🤗 
cc @thisis-Shitanshu in case you want to have a look at the table sizing issue 🤗

thisis-Shitanshu:
Thanks @christian-heusel ! I’ll take a look. First, by reproducing the current behavior across different viewport sizes and compare it with the approach from this PoC 🙂

PR #1329, opened 2026-08-15
feat: show imageConfig/podConfig description as tooltip in workspace table

christian-heusel:
The workspace table lists each workspace's image and pod config only by their `displayName` (e.g. "Small CPU"). The underlying `WorkspaceKind` options already carry a human-written `description` explaining what the option actually provides (resources, intended use, etc.), but it wasn't surfaced anywhere in the table view — users had to open the workspace kind or the create/edit form to see it.

This PR adds a hover tooltip on the image and pod config cells in the workspace table, showing that option's `description`, so users can understand what an option means without leaving the table.


thaorell (approved):
/lgtm

paulovmr (approved):
/lgtm
/approve

Issue #1295, opened 2026-07-31
[TASK] [Frontend] Display point-in-time resource usage widget on Workspace Detail Resources tab


richabanker:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

This task updates the Workspace Detail view's Resources tab in the Notebooks v2 frontend to asynchronously fetch and display live container CPU and Memory utilization alongside configured requests and limits. It ensures the detail page remains responsive without blocking on metrics requests and provides clear UX messaging for degraded or pending metrics states.

### Acceptance Criteria

 - [ ] The Resources tab asynchronously calls GET /api/v1/workspaces/:namespace/:name/resources/usage after initial rendering and shows a loading state while in flight
 - [ ] Current CPU and Memory usage are displayed relative to configured requests and limits per container (supporting multi-container pods, e.g., "250m / 500m requested (1 limit)")
 - [ ] When metricsFromMetricsServer is omitted for a container (pod just started), the UI displays a transient pending state (e.g., "Collecting...")
 - [ ] Degraded API error codes are mapped to human-readable UI messages (METRICS_API_NOT_AVAILABLE indicates Metrics Server is not installed; WORKSPACE_NOT_RUNNING shows usage as N/A while paused)
 - [ ] Static workspace configuration (volumes, storage mounts) renders and remains fully functional even when metrics data is unavailable
 - [ ] End-to-end tests cover successful usage rendering, multi-container pods, pending states, and degraded error messaging

PR #1334, opened 2026-08-17
feat: display cluster resource utilization

thaorell:
closes: #1295

- Add resource utilization cards to Resources tab in Workspace Details
- Add `useWorkspaceResourceUsage` hook to handle resource data fetching
- Add tests
- [FUTURE WORK] Add trend line charts to show resources
Screenshot 2026-08-17 at 4 36 07 PM


christian-heusel:
@thaorell could you rebase your changes on top of `notebooks-v2` now that https://github.com/kubeflow/notebooks/pull/1241 is merged? 🤗

christian-heusel:
@thaorell could you rebase your changes? 🤗 From what I can see this is otherwise ready to go, so it would be nice to get it in shape to 🚢 😄

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceResourceCards.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  const containerData = activeContainer ? containers[activeContainer] : undefined;
+
+  const resourceKeys = useMemo(
+    () => (containerData ? getResourceKeys(containerData) : []),
+    [containerData],
+  );
+
+  return (
+    <DetailsLoadingState error={error} loaded={loaded}>
```
paulovmr, line 135: All resource-usage failures are passed to DetailsLoadingState, which renders only “Failed to load details.” A paused workspace therefore loses its resource section instead of showing usage as N/A, and the API error code/message is discarded; to complete the acceptance criteria, the expected degraded responses should be mapped to their specified UI states and add assertions for those messages.

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceResourceCards.tsx
```diff
@@ -0,0 +1,183 @@
...
+                  },
+                  ...containerNames.map((name) => ({
+                    content: name,
+                    value: name,
+                    selected: name === activeContainer,
+                  })),
+                ]}
+                onSelect={(_ev, selection) => setSelectedContainer(String(selection))}
+                toggleProps={{
+                  'aria-labelledby': 'resource-container-label',
```
paulovmr: The select toggle references `resource-container-label`, but no element with that id is rendered. This leaves the interactive container selector without an accessible name; wrap it in the existing labelled-control pattern or add a visible label with the matching id.

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceResourceCards.tsx
```diff
@@ -0,0 +1,183 @@
...
+              <span className="pf-v6-c-description-list__term">Resource Utilization</span>
+            </FlexItem>
+            <FlexItem>
+              <SimpleSelect
+                initialOptions={[
+                  {
+                    content: 'Select a container',
+                    value: '',
+                    isDisabled: true,
+                    selected: !selectedContainer,
```
paulovmr: When resource data loads, both the placeholder and the first container are marked selected because `selectedContainer` remains undefined. SimpleSelect picks the first selected option, so the cards display the first container while the control says “Select a container”; select the first returned container or only select the placeholder when there are no containers.

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceResourceCards.tsx
```diff
@@ -0,0 +1,183 @@
...
+const ResourceCard: React.FC<ResourceCardProps> = ({ resourceKey, container }) => {
+  const requestValue = container.resources.requests?.[resourceKey];
+  const limitValue = container.resources.limits?.[resourceKey];
+  const metrics = container.metricsFromMetricsServer;
+  const usageValue = metrics?.usage[resourceKey as keyof typeof metrics.usage] ?? undefined;
+
+  return (
+    <Card isCompact data-testid={`resource-card-${resourceKey}`}>
+      <CardTitle>
+        <Content component={ContentVariants.h3}>{getDisplayName(resourceKey)}</Content>
```
paulovmr, line 74: The resource keys are cast to ResourceType, but `nvidia.com/gpu` is not a key in RESOURCE_DISPLAY_NAMES. The GPU test therefore renders a card with an empty heading, which is neither understandable nor accessible; use a display-name map for extended resource keys or fall back to the raw key, and assert the heading in the test.

workspaces/frontend/src/__tests__/cypress/cypress/pages/workspaces/workspaces.ts
```diff
@@ -388,6 +388,53 @@ class WorkspaceDetailsDrawer {
...
+  assertResourcesTabContentContainsText(text: string) {
+    this.findResourcesTabContent().should('contain.text', text);
+  }
+
+  assertResourceErrorAlert(testId: string) {
+    this.findResourcesTabContent().findByTestId(testId).should('exist');
+  }
+
+  selectResourceContainer(optionName: string) {
+    this.findResourcesTabContent().find('#resource-container-select').click();
```
paulovmr: This page object locates the container select by its DOM id even though it is not a form control. Add a `data-testid` to the toggle and use `findByTestId` so this selector remains stable under implementation changes.

paulovmr (changes requested):

workspaces/frontend/src/app/pages/Workspaces/Details/WorkspaceResourceCards.tsx
paulovmr: The fix only special-cases an error when the workspace state is paused; every other resource-usage error still falls through to the generic “Failed to load details” state. Please consider preserving and inspecting the API error code so METRICS_API_NOT_AVAILABLE can explain that Metrics Server is unavailable, while WORKSPACE_NOT_RUNNING renders the specified N/A state.

paulovmr (changes requested):

paulovmr (approved):
/lgtm
/approve

PR #1311, opened 2026-08-06
feat: give each Workspace its own ServiceAccount

Jefftree:
Workspaces can't be granted RBAC individually while they all share one static `default-editor` ServiceAccount, so the controller now creates and owns a ServiceAccount per Workspace, and WorkspaceKind gains `spec.podTemplate.serviceAccount.clusterRoles` to bind ClusterRoles to it through namespaced RoleBindings.

This grants the controller the `bind` verb on `clusterroles`, which is worth a close look. Kubernetes refuses a RoleBinding create unless the creator holds every permission in the referenced role or holds `bind` on it, so binding administrator-chosen ClusterRoles is impossible without it.

related: #1257

PR #1337, opened 2026-08-19
feat: let a Workspace bind Roles to its ServiceAccount, gated by a SubjectAccessReview

Jefftree:
## What this PR does

Adds `Workspace.spec.podTemplate.serviceAccount.roles[]`. Each entry becomes a namespaced RoleBinding from the Workspace's controller-owned ServiceAccount to a Role of that name. Changing the list is gated by a SubjectAccessReview in the Workspace webhook: `create` rolebindings to add an entry, `delete` to remove one.

Implements rows 1 and 2 of the authorization table in #1257.

Kubernetes prevents privilege escalation through RoleBinding in the REST storage layer, and skips that check when the creator holds `bind` on the referenced role. The controller must hold `bind` cluster-wide, since it binds roles whose permissions it does not itself hold, so the native check never fires for our bindings. The SAR is what replaces it.

## Decisions

- **The SAR checks "can you create rolebindings here", not "can you bind this Role".** Creating the binding by hand would also require holding the Role's rules or holding `bind` on it. The first needs the API server's rule resolver, which a webhook cannot reach. The second is just another SAR, but neither the default `admin` nor `kubeflow-admin` holds `bind`, so requiring it would lock out the intended user until an admin grants it.
- **Removing an entry requires `delete rolebindings`,** per the issue. Note this gates de-escalation, and deleting the Workspace drops the bindings with no check at all, so a caller with `create` but not `delete` can add a Role and never remove it.
- **No `kind` field on `roles[]`,** per the issue. The reconciler is already keyed on `(kind, name)`, so the API is the only thing that would change if we ever want per-Workspace ClusterRoles.
- **The RoleBinding name hashes `(workspace, kind, role)`.** Without the kind, a Role and a ClusterRole of the same name collide onto one RoleBinding and the reconciler flaps between two `roleRef` values.
- **The backend preserves `spec.podTemplate.serviceAccount` across PUT.** `ApplyWorkspaceUpdateModelToWorkspace` rebuilds `spec.podTemplate` wholesale, so without this the field is wiped on every backend update. Inverting that function to assign field by field would make preserve the default, but that is a bigger change than belongs here.

## Known gaps

- The backend acts as its own ServiceAccount and does not impersonate. Before it can expose this field it must either impersonate the user or run its own SAR, otherwise the webhook check evaluates the backend rather than the caller.
- Row 4 of the issue, the `can-use=true` label gate on Roles, lands with the backend API surface. The issue describes it as consistent with StorageClasses and Secrets, but this codebase gates StorageClasses with `can-use` and Secrets with `can-mount`, so that needs settling.
- `reconcileRoleBindings` runs last in `Reconcile`, after eleven paths that can return early, so a Workspace stuck in `Error` keeps its bindings after a Role is removed from the spec. Pre-existing, worth a separate fix.

## Testing

Controller and backend `make test` and `make lint` pass, no generated drift. New coverage: webhook allow and deny paths for add and remove against real non-admin identities, RoleBinding creation and deletion per entry, Role and ClusterRole name separation, and one test that creates a real Role, binds it, and confirms with a SubjectAccessReview that the ServiceAccount gains and then loses the permission.


workspaces/controller/internal/webhook/workspace_webhook.go
```diff
@@ -205,6 +222,115 @@ func (v *WorkspaceValidator) validateWorkspaceKind(ctx context.Context, workspac
...
+//
+// This is the whole security boundary. The controller holds "bind" cluster-wide, so the RBAC
+// escalation prevention which normally runs on RoleBinding create never applies to our bindings.
+//
+// NOTE: we check neither that the caller holds the Role's rules, nor that they hold "bind" on it,
+//
+//	both of which the API server would require. The first needs its rule resolver, which a
+//	webhook cannot reach. The second is just another SAR, but requiring it would lock out the
+//	default "admin" and "kubeflow-admin" roles, which hold neither.
+//	See https://github.com/kubeflow/notebooks/issues/1257.
```
thesuperzapper: We are we talking about how we dont solve this problem, we should explain the actual method used?

And probably put that explanation of how we solve it directly above the code which does that.

workspaces/controller/internal/webhook/workspace_webhook.go
```diff
@@ -205,6 +222,115 @@ func (v *WorkspaceValidator) validateWorkspaceKind(ctx context.Context, workspac
...
+			removed = append(removed, name)
+		}
+	}
+	if len(added) == 0 && len(removed) == 0 {
+		return nil
+	}
+
+	// NOTE: the userInfo comes from the AdmissionRequest, not the webhook's own identity.
+	//       Requests which do not carry one (that is, calls which did not come through admission)
+	//       are rejected rather than allowed, so a missing request can never mean "allowed".
```
thesuperzapper, line 267: I wonder if we should link the issue upstream for a bit of context (no pun intended):
- https://github.com/kubernetes-sigs/controller-runtime/issues/1600#issuecomment-4897745096

workspaces/controller/internal/webhook/workspace_webhook.go
```diff
@@ -205,6 +222,115 @@ func (v *WorkspaceValidator) validateWorkspaceKind(ctx context.Context, workspac
...
+	if reason == "" {
+		reason = "no RBAC policy matched"
+	}
+	return field.Forbidden(
+		rolesPath,
+		fmt.Sprintf("changing %v requires permission to %s RoleBindings in namespace %q, which user %q does not have (%s)",
+			roleNames, verb, namespace, req.UserInfo.Username, reason),
+	)
+}
+
```
thesuperzapper: This should be replaced with the caching authorizer from the kubernetes apiserver go package, similar to how we are doing it in the workspaces backend.

https://github.com/kubeflow/notebooks/blob/455b3339e9e5f5ff610d6d3e04a7d68fa9ad5c29/workspaces/backend/internal/auth/authorization.go#L41-L74

thesuperzapper (changes requested):

workspaces/controller/api/v1beta1/workspace_types.go
```diff
@@ -63,10 +63,47 @@ type WorkspacePodTemplate struct {
...
+}
+
+// WorkspaceRole identifies a Role to bind to a Workspace's ServiceAccount via a RoleBinding.
+type WorkspaceRole struct {
+	// the name of the Role to bind to the Workspace ServiceAccount
+	//  - note, Role names are path segment names, so unlike most Kubernetes
+	//    resource names they may contain uppercase letters and ":"
+	// +kubebuilder:validation:MinLength:=1
+	// +kubebuilder:validation:MaxLength:=253
+	// +kubebuilder:validation:XValidation:message="must not be '.' or '..', and must not contain '/' or '%'",rule="self != '.' && self != '..' && !self.contains('/') && !self.contains('%')"
```
thesuperzapper: Using `XValidation` uses up CEL validation reource tokens https://kubernetes.io/docs/reference/using-api/cel/#resource-constraints

We need to think about if its better for us to use our own webhook on this one, with the upstream apimachinery `IsPathSegmentPrefix` function? (I am open to leaving it, just want to know your thoughts).

Also, are we sure that the max length is 253, have we tested this on real clusters?

workspaces/controller/internal/webhook/workspace_webhook.go
Jefftree: done, moved the explanation onto `authorizeRoleBindingVerb()` and left the alternatives we didn't take in https://github.com/kubeflow/notebooks/issues/1257.

workspaces/controller/internal/webhook/workspace_webhook.go
Jefftree: Added.

workspaces/controller/internal/webhook/workspace_webhook.go
Jefftree: done. Had to copy `NewRequestAuthorizer` rather than import it, the backend's is under its own module's `internal/`.

workspaces/controller/api/v1beta1/workspace_types.go
Jefftree: yeah, `MaxItems` and `MaxLength` were both just there to keep the CEL cost estimate under the limit. @christian-heusel pointed out the same thing on https://github.com/kubeflow/notebooks/pull/1311 and I forgot to carry it over here, so it's `Pattern` now like `WorkspaceKindClusterRole.Name` and both are gone.

Ah my bad, it looks like path segment names don't have a length restriction, removed that check.

PR #1341, opened 2026-08-21
fix: break infinite status reconciliation loop in notebook controller

Raakshass:
## Summary

Fixes the infinite reconciliation loop in the notebook controller where `updateNotebookStatus` issues a `Status().Update()` PUT on every reconcile even when nothing has changed.

## Root Cause

Three-link causal chain:

1. [`PodCondToNotebookCond` (line 330)](https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/notebook-controller/controllers/notebook_controller.go#L330) — kubelet never sets `LastProbeTime` on pod conditions, so the else branch injects `metav1.Now()` on every reconcile, producing a different timestamp each time.

2. [`updateNotebookStatus` (line 238-240)](https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/notebook-controller/controllers/notebook_controller.go#L238-L240) — unconditional `r.Status().Update(ctx, nb)` with no comparison guard.

3. [`SetupWithManager` (line 713-714)](https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/notebook-controller/controllers/notebook_controller.go#L713-L714) — `For(&v1beta1.Notebook{})` with no `GenerationChangedPredicate`, so the status write bumps `resourceVersion`, emits a watch event, and re-enqueues.

## Changes

| File | Change |
|------|--------|
| `controllers/notebook_controller.go` | Reorder `PodCondToNotebookCond` to resolve `LastTransitionTime` first, fall back `LastProbeTime` to `condition.LastTransitionTime` instead of `metav1.Now()`, add `reflect.DeepEqual` guard in `updateNotebookStatus` |

## Design Decisions

- **`LastProbeTime` falls back to `LastTransitionTime`, not zero** — the [CRD schema](https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/notebook-controller/config/crd/bases/kubeflow.org_notebooks.yaml#L3133-L3135) declares `lastProbeTime` as `type: string, format: date-time` without `nullable: true`, and `omitempty` on [`metav1.Time` (a struct)](https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/notebook-controller/api/v1beta1/notebook_types.go#L53) does not suppress the zero value.
- **Block reorder** — `LastTransitionTime` resolved first so `LastProbeTime` can reference it without a second `metav1.Now()` call.
- **Both changes needed** — `DeepEqual` guard alone does not break the loop because `metav1.Now()` makes it return false every time. The timestamp fix makes the status deterministic; the guard eliminates the residual no-op PUT.
- **No new imports** — `reflect` already used at [line 257](https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/notebook-controller/controllers/notebook_controller.go#L257).

## Test Results

All 5 `TestCreateNotebookStatus` subtests pass. `go vet` and `go build` clean.

## Follow-up

- Add test case with `LastProbeTime` omitted asserting fallback equals `LastTransitionTime`
- `CopyStatefulSetFields` secondary loop is a separate concern

## Contributor Checklist

- [x] DCO sign-off
- [x] All existing tests pass
- [x] `go vet` clean
- [x] `go build` clean
- [x] Single file, minimal diff

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -235,6 +235,10 @@ func updateNotebookStatus(r *NotebookReconciler, nb *v1beta1.Notebook,
 		return err
 	}
 
+	if reflect.DeepEqual(nb.Status, status) {
+		return nil
+	}
+
 	log.Info("Updating Notebook CR Status", "status", status)
 	nb.Status = status
 	return r.Status().Update(ctx, nb)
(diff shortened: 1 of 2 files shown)


Raakshass:
@juliusvonkohout

juliusvonkohout:
closed in favor of https://github.com/kubeflow/notebooks/pull/1343

PR #1343, opened 2026-08-21
fix: infinite status reconciliation loop in notebook controller

juliusvonkohout:
Alternative to https://github.com/kubeflow/notebooks/pull/1341

The reconcilliation loop creates a lot of logging costs and api server load if you have for example 250 notebooks on a cluster.

```
│ 1.7873218948448386e+09    INFO    controllers.Notebook    Reconciliation loop started    {"notebook": "my-namespace/my-notebook"}                    │
│ 1.787321894844961e+09    INFO    controllers.Notebook    Updating StatefulSet    {"notebook": "my-namespace/my-notebook", "namespace": "xdiv-finance │
│ -ior-prediction-dev", "name": "daniel"}                                                                                                                            │
│ 1.7873218949141026e+09    INFO    controllers.Notebook    Initializing Notebook CR Status    {"notebook": "my-namespace/my-notebook"}                │
│ 1.7873218949141288e+09    INFO    controllers.Notebook    Calculating Notebook's  containerState    {"notebook": "my-namespace/my-notebook"}         │
│ 1.7873218949141347e+09    INFO    controllers.Notebook    Updating Notebook CR state:     {"notebook": "my-namespace/my-notebook", "state": {"runnin │
│ g":{"startedAt":"2026-08-18T13:11:00Z"}}}                                                                                                                          │
│ 1.7873218949141548e+09    INFO    controllers.Notebook    Calculating Notebook's Conditions    {"notebook": "my-namespace/my-notebook"}              │
│ 1.787321894914165e+09    INFO    controllers.Notebook    Updating Notebook CR Status    {"notebook": "my-namespace/my-notebook", "status": {"conditi │
│ ons":[{"type":"PodReadyToStartContainers","status":"True","lastProbeTime":"2026-08-21T14:18:14Z","lastTransitionTime":"2026-08-18T13:10:58Z"},{"type":"Initialized │
│ ","status":"True","lastProbeTime":"2026-08-21T14:18:14Z","lastTransitionTime":"2026-08-18T13:11:00Z"},{"type":"Ready","status":"True","lastProbeTime":"2026-08-21T │
│ 14:18:14Z","lastTransitionTime":"2026-08-18T13:11:07Z"},{"type":"ContainersReady","status":"True","lastProbeTime":"2026-08-21T14:18:14Z","lastTransitionTime":"202 │
│ 6-08-18T13:11:07Z"},{"type":"PodScheduled","status":"True","lastProbeTime":"2026-08-21T14:18:14Z","lastTransitionTime":"2026-08-18T13:10:37Z"}],"readyReplicas":1, │
│ "containerState":{"running":{"startedAt":"2026-08-18T13:11:00Z"}}}}   
```
Imagine this continously at full speed without interruption increasing the revisions.
Assisted-by: GitHub Copilot

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -137,6 +137,9 @@ func (r *NotebookReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	// Reconcile StatefulSet
 	ss := generateStatefulSet(instance)
+	if err := reconcilehelper.SetStatefulSetTemplateHash(ss); err != nil {
+		return ctrl.Result{}, err
+	}
 	if err := ctrl.SetControllerReference(instance, ss, r.Scheme); err != nil {
 		return ctrl.Result{}, err
 	}
@@ -235,6 +238,10 @@ func updateNotebookStatus(r *NotebookReconciler, nb *v1beta1.Notebook,
 		return err
 	}
 
+	if reflect.DeepEqual(nb.Status, status) {
+		return nil
+	}
+
 	log.Info("Updating Notebook CR Status", "status", status)
 	nb.Status = status
 	return r.Status().Update(ctx, nb)
@@ -269,13 +276,11 @@ func createNotebookStatus(r *NotebookReconciler, nb *v1beta1.Notebook,
 			continue
 		}
 
-		if pod.Status.ContainerStatuses[i].State == nb.Status.ContainerState {
-			continue
-		}
-
 		// Update Notebook CR's status.ContainerState
 		cs := pod.Status.ContainerStatuses[i].State
-		log.Info("Updating Notebook CR state: ", "state", cs)
+		if cs != nb.Status.ContainerState {
+			log.Info("Updating Notebook CR state: ", "state", cs)
+		}
 
 		status.ContainerState = cs
 		notebookContainerFound = true
@@ -321,22 +326,20 @@ func PodCondToNotebookCond(podc corev1.PodCondition) v1beta1.NotebookCondition {
 		condition.Reason = podc.Reason
 	}
 
-	// check if podc.LastProbeTime is null. If so initialize
-	// the field with metav1.Now()
-	check := podc.LastProbeTime.Time.Equal(time.Time{})
-	if !check {
-		condition.LastProbeTime = podc.LastProbeTime
-	} else {
-		condition.LastProbeTime = metav1.Now()
-	}
-
 	// check if podc.LastTransitionTime is null. If so initialize
 	// the field with metav1.Now()
-	check = podc.LastTransitionTime.Time.Equal(time.Time{})
+	check := podc.LastTransitionTime.Time.Equal(time.Time{})
 	if !check {
 		condition.LastTransitionTime = podc.LastTransitionTime
 	} else {
 		condition.LastTransitionTime = metav1.Now()
+	}
+
+	check = podc.LastProbeTime.Time.Equal(time.Time{})
+	if !check {
+		condition.LastProbeTime = podc.LastProbeTime
+	} else {
+		condition.LastProbeTime = condition.LastTransitionTime
 	}
 
 	return condition
--- a/components/notebook-controller/controllers/notebook_controller_test.go
+++ b/components/notebook-controller/controllers/notebook_controller_test.go
@@ -180,6 +180,35 @@ func TestCreateNotebookStatus(t *testing.T) {
 			},
 		},
 		{
+			name: "NotebookContainerStateUnchanged",
+			currentNb: nbv1beta1.Notebook{
+				ObjectMeta: v1.ObjectMeta{Name: "test"},
+				Status: nbv1beta1.NotebookStatus{
+					ContainerState: corev1.ContainerState{
+						Running: &corev1.ContainerStateRunning{},
+					},
+				},
+			},
+			pod: corev1.Pod{
+				Status: corev1.PodStatus{
+					ContainerStatuses: []corev1.ContainerStatus{
+						{
+							Name: "test",
+							State: corev1.ContainerState{
+								Running: &corev1.ContainerStateRunning{},
+							},
+						},
+					},
+				},
+			},
+			expectedNbStatus: nbv1beta1.NotebookStatus{
+				Conditions: []nbv1beta1.NotebookCondition{},
+				ContainerState: corev1.ContainerState{
+					Running: &corev1.ContainerStateRunning{},
+				},
+			},
+		},
+		{
 			name: "mirroringPodConditions",
 			pod: corev1.Pod{
 				ObjectMeta: v1.ObjectMeta{
@@ -226,6 +255,31 @@ func TestCreateNotebookStatus(t *testing.T) {
 					},
 				},
 				ReadyReplicas:  int32(1),
+				ContainerState: corev1.ContainerState{},
+			},
+		},
+		{
+			name: "mirroringPodConditionWithoutProbeTime",
+			pod: corev1.Pod{
+				Status: corev1.PodStatus{
+					Conditions: []corev1.PodCondition{
+						{
+							Type:               "Ready",
+							Status:             "True",
+							LastTransitionTime: v1.Date(2022, time.Month(8), 30, 1, 10, 30, 0, time.UTC),
+						},
+					},
+				},
+			},
+			expectedNbStatus: nbv1beta1.NotebookStatus{
+				Conditions: []nbv1beta1.NotebookCondition{
+					{
+						Type:               "Ready",
+						Status:             "True",
+						LastProbeTime:      v1.Date(2022, time.Month(8), 30, 1, 10, 30, 0, time.UTC),
+						LastTransitionTime: v1.Date(2022, time.Month(8), 30, 1, 10, 30, 0, time.UTC),
+					},
+				},
 				ContainerState: corev1.ContainerState{},
 			},
 		},
@@ -287,6 +341,21 @@ func TestCreateNotebookStatus(t *testing.T) {
 		})
 	}
 
+}
+
+func TestUpdateNotebookStatusSkipsUnchangedStatus(t *testing.T) {
+	reconciler := createMockReconciler()
+	reconciler.Client = fake.NewFakeClientWithScheme(scheme.Scheme)
+	notebook := &nbv1beta1.Notebook{
+		Status: nbv1beta1.NotebookStatus{
+			Conditions:     []nbv1beta1.NotebookCondition{},
+			ContainerState: corev1.ContainerState{},
+		},
+	}
+
+	if err := updateNotebookStatus(reconciler, notebook, &appsv1.StatefulSet{}, &corev1.Pod{}, ctrl.Request{}); err != nil {
+		t.Fatalf("expected unchanged status to skip API update: %v", err)
+	}
 }
 
 func createMockReconciler() *NotebookReconciler {
--- a/components/notebook-controller/controllers/notebook_controller_test.go
+++ b/components/notebook-controller/controllers/notebook_controller_test.go
@@ -259,7 +259,7 @@ func TestCreateNotebookStatus(t *testing.T) {
 			},
 		},
 		{
-			name: "mirroringPodConditionWithoutProbeTime",
+			name: "LastProbeTimeDefaultsToLastTransitionTime",
 			pod: corev1.Pod{
 				Status: corev1.PodStatus{
 					Conditions: []corev1.PodCondition{
--- a/components/notebook-controller/reconcilehelper/util.go
+++ b/components/notebook-controller/reconcilehelper/util.go
@@ -2,12 +2,16 @@ package reconcile
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
 	"reflect"
 
 	"github.com/go-logr/logr"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
@@ -102,33 +106,41 @@ func VirtualService(ctx context.Context, r client.Client, virtualServiceName, na
 
 // Reference: https://github.com/pwittrock/kubebuilder-workshop/blob/master/pkg/util/util.go
 
+const StatefulSetTemplateHashAnnotation = "notebooks.kubeflow.org/statefulset-template-hash"
+
+func SetStatefulSetTemplateHash(statefulSet *appsv1.StatefulSet) error {
+	serializedTemplate, err := json.Marshal(statefulSet.Spec.Template)
+	if err != nil {
+		return fmt.Errorf("serialize StatefulSet pod template: %w", err)
+	}
+
+	if statefulSet.Annotations == nil {
+		statefulSet.Annotations = make(map[string]string)
+	}
+	statefulSet.Annotations[StatefulSetTemplateHashAnnotation] = fmt.Sprintf("%x", sha256.Sum256(serializedTemplate))
+	return nil
+}
+
 // CopyStatefulSetFields copies the owned fields from one StatefulSet to another
 // Returns true if the fields copied from don't match to.
 func CopyStatefulSetFields(from, to *appsv1.StatefulSet) bool {
 	requireUpdate := false
-	for k, v := range to.Labels {
-		if from.Labels[k] != v {
-			requireUpdate = true
-		}
-	}
-	to.Labels = from.Labels
-
-	for k, v := range to.Annotations {
-		if from.Annotations[k] != v {
-			requireUpdate = true
-		}
-	}
-	to.Annotations = from.Annotations
 
 	if *from.Spec.Replicas != *to.Spec.Replicas {
 		*to.Spec.Replicas = *from.Spec.Replicas
 		requireUpdate = true
 	}
 
-	if !reflect.DeepEqual(to.Spec.Template.Spec, from.Spec.Template.Spec) {
+	desiredTemplateHash := from.Annotations[StatefulSetTemplateHashAnnotation]
+	if desiredTemplateHash != to.Annotations[StatefulSetTemplateHashAnnotation] ||
+		!apiequality.Semantic.DeepDerivative(from.Spec.Template, to.Spec.Template) {
+		to.Spec.Template = from.Spec.Template
+		if to.Annotations == nil {
+			to.Annotations = make(map[string]string)
+		}
+		to.Annotations[StatefulSetTemplateHashAnnotation] = desiredTemplateHash
 		requireUpdate = true
 	}
-	to.Spec.Template.Spec = from.Spec.Template.Spec
 
 	return requireUpdate
 }
--- a/components/notebook-controller/reconcilehelper/util_test.go
+++ b/components/notebook-controller/reconcilehelper/util_test.go
@@ -1,0 +1,180 @@
+package reconcile
+
+import (
+	"reflect"
+	"testing"
+
+	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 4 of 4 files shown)


juliusvonkohout:
docker.io/jvonkoho/notebook-controller:pr-1343 if you want to test it. It works well in practice.

christian-heusel (changes requested):
This pull request is currently lacking a few quite basic things, please add them and we  can proceed with the review itself:

1. Motivation: What is the bug that is being fixed here?
2. Reproducer: How can the bug be reproduced by a developer?
3. Implementation: How is the submitted changeset fixing the issue?

Also I suspect you used AI in the creation of this change, please add appropriate trailers in order comply with the Kubeflow AI policy 😊

juliusvonkohout:
> This pull request is currently lacking a few quite basic things, please add them and we can proceed with the review itself:
> 
>     1. Motivation: What is the bug that is being fixed here?

Alternative to https://github.com/kubeflow/notebooks/pull/1341

The reconciliation loop creates a lot of logging costs and api server load if you have for example 250 notebooks on a cluster.

```
│ 1.7873218948448386e+09    INFO    controllers.Notebook    Reconciliation loop started    {"notebook": "my-namespace/my-notebook"}                    │
│ 1.787321894844961e+09    INFO    controllers.Notebook    Updating StatefulSet    {"notebook": "my-namespace/my-notebook", "namespace": "xdiv-finance │
│ -ior-prediction-dev", "name": "daniel"}                                                                                                                            │
│ 1.7873218949141026e+09    INFO    controllers.Notebook    Initializing Notebook CR Status    {"notebook": "my-namespace/my-notebook"}                │
│ 1.7873218949141288e+09    INFO    controllers.Notebook    Calculating Notebook's  containerState    {"notebook": "my-namespace/my-notebook"}         │
│ 1.7873218949141347e+09    INFO    controllers.Notebook    Updating Notebook CR state:     {"notebook": "my-namespace/my-notebook", "state": {"runnin │
│ g":{"startedAt":"2026-08-18T13:11:00Z"}}}                                                                                                                          │
│ 1.7873218949141548e+09    INFO    controllers.Notebook    Calculating Notebook's Conditions    {"notebook": "my-namespace/my-notebook"}              │
│ 1.787321894914165e+09    INFO    controllers.Notebook    Updating Notebook CR Status    {"notebook": "my-namespace/my-notebook", "status": {"conditi │
│ ons":[{"type":"PodReadyToStartContainers","status":"True","lastProbeTime":"2026-08-21T14:18:14Z","lastTransitionTime":"2026-08-18T13:10:58Z"},{"type":"Initialized │
│ ","status":"True","lastProbeTime":"2026-08-21T14:18:14Z","lastTransitionTime":"2026-08-18T13:11:00Z"},{"type":"Ready","status":"True","lastProbeTime":"2026-08-21T │
│ 14:18:14Z","lastTransitionTime":"2026-08-18T13:11:07Z"},{"type":"ContainersReady","status":"True","lastProbeTime":"2026-08-21T14:18:14Z","lastTransitionTime":"202 │
│ 6-08-18T13:11:07Z"},{"type":"PodScheduled","status":"True","lastProbeTime":"2026-08-21T14:18:14Z","lastTransitionTime":"2026-08-18T13:10:37Z"}],"readyReplicas":1, │
│ "containerState":{"running":{"startedAt":"2026-08-18T13:11:00Z"}}}}   
```
Imagine this continously at full speed without interruption increasing the revisions.

Description form @Raakshass 


Fixes the infinite reconciliation loop in the notebook controller where `updateNotebookStatus` issues a `Status().Update()` PUT on every reconcile even when nothing has changed.

## Root Cause

Three-link causal chain:

1. [`PodCondToNotebookCond` (line 330)](https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/notebook-controller/controllers/notebook_controller.go#L330) — kubelet never sets `LastProbeTime` on pod conditions, so the else branch injects `metav1.Now()` on every reconcile, producing a different timestamp each time.

2. [`updateNotebookStatus` (line 238-240)](https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/notebook-controller/controllers/notebook_controller.go#L238-L240) — unconditional `r.Status().Update(ctx, nb)` with no comparison guard.

3. [`SetupWithManager` (line 713-714)](https://github.com/kubeflow/notebooks/blob/notebooks-v1/components/notebook-controller/controllers/notebook_controller.go#L713-L714) — `For(&v1beta1.Notebook{})` with no `GenerationChangedPredicate`, so the status write bumps `resourceVersion`, emits a watch event, and re-enqueues.

> 
>     2. Reproducer: How can the bug be reproduced by a developer?

Just create many e.g. 100 notebooks and watch the reconciliation loop indefinitely.

> 
>     3. Implementation: How is the submitted changeset fixing the issue?


Alternative to #1341.

## 1. Motivation

The notebook controller enters an effectively infinite reconciliation loop even when neither the Notebook nor its workload has changed.

Kubelet commonly leaves `PodCondition.LastProbeTime` unset. The controller previously replaced this empty value with `metav1.Now()` during every reconciliation. This made the computed Notebook status different every time, causing an unconditional `Status().Update()`. That update changed the Notebook's `resourceVersion`, triggered another watch event, and immediately re-enqueued it.

The StatefulSet comparison could also interpret API-defaulted fields as drift, causing unnecessary updates and increasing controller revisions.

On clusters with hundreds of Notebooks, this generates substantial API server load and continuous logging costs.

## 2. Reproducer

1. Deploy the existing notebook controller.
2. Create one or more Notebooks; creating approximately 100 makes the impact particularly visible.
3. Wait until the Notebooks and their StatefulSets are stable.
4. Follow the notebook-controller logs and watch the Notebook resources and StatefulSet revisions.

Despite no user changes, the controller continuously reconciles the same Notebooks, writes their status, updates StatefulSets, and increases resource versions and revisions.

## 3. Implementation

This change makes reconciliation converge:

- `LastProbeTime` now falls back to the stable `LastTransitionTime` instead of using the current time on every reconciliation.
- The computed Notebook status is compared with the existing status. `Status().Update()` is skipped when they are equal.
- Desired StatefulSet pod templates receive a SHA-256 hash annotation.
- StatefulSet templates are compared using Kubernetes semantic derivative comparison, preventing API-defaulted or additional fields from causing false drift.
- The template and its hash are copied only when the desired template actually changes. Existing StatefulSets without the annotation receive one migration update and then converge.
- Tests cover unchanged status, condition timestamps, Kubernetes-defaulted fields, desired template changes and removals, drift detection, and hash migration.

> 
> 
> Also I suspect you used AI in the creation of this change, please add appropriate trailers in order comply with the Kubeflow AI policy 😊

Assisted-by: GitHub Copilot

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -321,22 +326,20 @@ func PodCondToNotebookCond(podc corev1.PodCondition) v1beta1.NotebookCondition {
...
 	// the field with metav1.Now()
-	check := podc.LastProbeTime.Time.Equal(time.Time{})
+	check := podc.LastTransitionTime.Time.Equal(time.Time{})
 	if !check {
-		condition.LastProbeTime = podc.LastProbeTime
+		condition.LastTransitionTime = podc.LastTransitionTime
 	} else {
-		condition.LastProbeTime = metav1.Now()
+		condition.LastTransitionTime = metav1.Now()
 	}
```
thesuperzapper: The issue which is causing the problem raised (update of status when there should not be) is much sipler than it seems to fix.

The problem was introduced in https://github.com/kubeflow/notebooks/commit/02e2fa4c65250959ca51b827e2283d6b8c77a746#diff-a436986c5e93996f9035eee6798d064ab8659957e8443eb44b3a4d3db289bde4R380-R395, where is seems the author misunderstood what the `PodCondToNotebookCond` function did.

It should simply map from `PodCondition` to `NotebookCondition`, which should be a 1-1 mapping, thus there should never be a case where we update the `LastTransitionTime` or `LastProbeTime` because it MUST be exactly the same as the Pod condition.

The correct fix is to simply set `LastTransitionTime` and `LastProbeTime` to the exact values from the `PodCondition`.

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -235,6 +238,10 @@ func updateNotebookStatus(r *NotebookReconciler, nb *v1beta1.Notebook,
 		return err
 	}
 
+	if reflect.DeepEqual(nb.Status, status) {
+		return nil
+	}
+
```
thesuperzapper: This is proably still fine, but we should cofirm if deepEaul actually works in this case, rather than semantic, or some other check due to default values not being seralised.

We probably just can remove this to get the PR moving however.

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -137,6 +137,9 @@ func (r *NotebookReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	// Reconcile StatefulSet
 	ss := generateStatefulSet(instance)
+	if err := reconcilehelper.SetStatefulSetTemplateHash(ss); err != nil {
+		return ctrl.Result{}, err
+	}
```
thesuperzapper: Lets remove this, this is not needed to fix the stated issue, and will likley introduce other issues.

thesuperzapper (changes requested):

components/notebook-controller/controllers/notebook_controller.go
juliusvonkohout: Agreed. I have removed the StatefulSet hashing changes and associated tests from this fix. The production change is now limited to exact timestamp copying.

components/notebook-controller/controllers/notebook_controller.go
juliusvonkohout: I have removed the equality guard. It is not needed to fix the timestamp-driven feedback loop. An API-backed regression test on Kubernetes 1.31 confirms that resource versions stabilize without it and subsequent Pod status changes still propagate.

components/notebook-controller/controllers/notebook_controller.go
juliusvonkohout: Agreed. I have removed both timestamp fallbacks and now copy LastProbeTime and LastTransitionTime exactly, including zero values.

christian-heusel:
As we discussed in Slack I have now checked the issue again that the PR is fixing and was able to reproduce the issue 🐛 

I left the information from the agent-based debugging session below, since there is also some interesting information on why a certain amount of notebooks is needed for this to actually trigger the self-sustained loop:

---

## Root cause

`components/notebook-controller/controllers/notebook_controller.go` @ tag `v1.11.0`:

1. **Timestamp injection**: `PodCondToNotebookCond` stamps `metav1.Now()` whenever a pod
   condition's `LastProbeTime` is unset (which kubelet leaves unset for pods without a
   readiness probe):
   ```go
   check := podc.LastProbeTime.Time.Equal(time.Time{})
   if !check { condition.LastProbeTime = podc.LastProbeTime } else { condition.LastProbeTime = metav1.Now() }
   ```
2. **Unconditional status write**: `updateNotebookStatus` never compares computed vs.
   stored status:
   ```go
   nb.Status = status
   return r.Status().Update(ctx, nb)
   ```
   `createNotebookStatus` rebuilds `status.Conditions` from the pod every reconcile, so the
   Ready condition gets a fresh `metav1.Now()` each time.
3. **No `GenerationChangedPredicate`**: `SetupWithManager` uses a bare
   `For(&v1beta1.Notebook{})`, so a status write bumps `resourceVersion`, emits a watch
   event, and re-enqueues the Notebook.

### Reason for self-sustained looping

`metav1.Time` has 1-second resolution. Chain of events:

- A reconcile writes `lastProbeTime = Now()` (truncated to the second) and re-enqueues
  itself via the `For()` watch.
- If the re-enqueued reconcile runs **within the same second**, the recomputed `Now()` is
  byte-identical → `Status().Update()` is a **no-op** at the apiserver → no watch event →
  **the loop dies.**
- If it runs **in a new second**, the timestamp differs → genuine write → re-enqueue →
  **the loop sustains.**

With one reconcile worker, ~100 Notebooks push the per-Notebook revisit interval above 1s,
so the loop self-sustains with **zero external input**. That is why "~100 makes the impact
particularly visible."

<details>

<summary>Click to see the full repro steps and testing environment</summary>

# Reproducing the notebook-controller reconcile loop (PR #1343)

Reproduction of the "infinite reconciliation loop" fixed by
[kubeflow/notebooks#1343](https://github.com/kubeflow/notebooks/pull/1343), run against a
local Kind cluster deploying Kubeflow with notebook-controller **v1.11.0**
(`ghcr.io/kubeflow/notebooks/notebook-controller:v1.11.0`, the pre-fix release).

**TL;DR** — The bug reproduces, but **only at scale (~100 Notebooks)**, exactly as the PR
notes. With ~10 Notebooks the controller goes completely silent. The reason is a detail
worth adding to the PR's write-up: `metav1.Time` serializes at **1-second** resolution, so
the self-loop only closes when a Notebook's self-triggered reconcile is revisited **more
than one second later** — which happens once the single reconcile worker is saturated by
~100 Notebooks, but not at small scale.

---

## Root cause (confirmed against the v1.11.0 source)

`components/notebook-controller/controllers/notebook_controller.go` @ tag `v1.11.0`:

1. **Timestamp injection** — `PodCondToNotebookCond` stamps `metav1.Now()` whenever a pod
   condition's `LastProbeTime` is unset (which kubelet leaves unset for pods without a
   readiness probe):
   ```go
   check := podc.LastProbeTime.Time.Equal(time.Time{})
   if !check { condition.LastProbeTime = podc.LastProbeTime } else { condition.LastProbeTime = metav1.Now() }
   ```
2. **Unconditional status write** — `updateNotebookStatus` never compares computed vs.
   stored status:
   ```go
   nb.Status = status
   return r.Status().Update(ctx, nb)
   ```
   `createNotebookStatus` rebuilds `status.Conditions` from the pod every reconcile, so the
   Ready condition gets a fresh `metav1.Now()` each time.
3. **No `GenerationChangedPredicate`** — `SetupWithManager` uses a bare
   `For(&v1beta1.Notebook{})`, so a status write bumps `resourceVersion`, emits a watch
   event, and re-enqueues the Notebook.

There is also a wasteful StatefulSet path: `CopyStatefulSetFields` uses
`reflect.DeepEqual(generated, stored)`, which is *always* true because the stored STS is
API-defaulted and the generated one is not, so `r.Update(sts)` is called on every
reconcile (see "Nuance" below).

### The scale threshold (addition to the PR's reproduction)

`metav1.Time` has 1-second resolution. Chain of events:

- A reconcile writes `lastProbeTime = Now()` (truncated to the second) and re-enqueues
  itself via the `For()` watch.
- If the re-enqueued reconcile runs **within the same second**, the recomputed `Now()` is
  byte-identical → `Status().Update()` is a **no-op** at the apiserver → no watch event →
  **the loop dies.**
- If it runs **in a new second**, the timestamp differs → genuine write → re-enqueue →
  **the loop sustains.**

With one reconcile worker, ~100 Notebooks push the per-Notebook revisit interval above 1s,
so the loop self-sustains with **zero external input**. That is why "~100 makes the impact
particularly visible."

---

## Environment

```bash
kind get kubeconfig --name kubeflow > /tmp/kubeflow-config
export KUBECONFIG=/tmp/kubeflow-config

kubectl get deploy -n kubeflow notebook-controller-deployment \
  -o jsonpath='{.spec.template.spec.containers[*].image}'
# ghcr.io/kubeflow/notebooks/notebook-controller:v1.11.0
```

- Namespace: `kubeflow-user-example-com` (has the `default-editor` ServiceAccount).
- 3-node Kind cluster (32 CPU / 128Gi each).
- The `Notebook` CRD's `spec.template.spec` is a plain PodSpec; `status` **is** a
  subresource (`{"status":{}}`).

## Reproduction

A lightweight `busybox` image reproduces the bug identically — the driver is the
probe-less pod condition plus scale, not Jupyter. (Real `jupyter-scipy` Notebooks were
tested as a control and behave the same; see below.)

Manifest per Notebook:

```yaml
apiVersion: kubeflow.org/v1
kind: Notebook
metadata:
  name: repro-<N>
  namespace: kubeflow-user-example-com
spec:
  template:
    spec:
      containers:
        - name: repro-<N>
          image: busybox:1.36
          command: ["sh", "-c", "sleep infinity"]
          ports:
            - { containerPort: 8888, name: notebook-port, protocol: TCP }
          resources:
            requests: { cpu: "10m", memory: "16Mi" }
```

Create ~100:

```bash
NS=kubeflow-user-example-com
for i in $(seq 0 99); do
  cat <<EOF | kubectl apply -f -
apiVersion: kubeflow.org/v1
kind: Notebook
metadata: { name: repro-$i, namespace: $NS }
spec:
  template:
    spec:
      containers:
        - name: repro-$i
          image: busybox:1.36
          command: ["sh", "-c", "sleep infinity"]
          ports: [{ containerPort: 8888, name: notebook-port, protocol: TCP }]
          resources: { requests: { cpu: "10m", memory: "16Mi" } }
EOF
done
```

## Observations

### Symptom counters (this run, ~93 Notebooks running, no user edits)

```bash
# Reconciles per 30s
kubectl logs -n kubeflow deploy/notebook-controller-deployment -c manager --since=30s \
  | grep -c "Reconciliation loop started"

# Needless Notebook status writes per 30s
kubectl logs -n kubeflow deploy/notebook-controller-deployment -c manager --since=30s \
  | grep -c "Updating Notebook CR Status"

# Aggregate resourceVersion climb across the fleet
kubectl get notebooks -n $NS -o jsonpath='{range .items[*]}{.metadata.resourceVersion}{" "}{end}' \
  | python3 -c "import sys;print(sum(int(x) for x in sys.stdin.read().split()))"
```

| Metric (idle, no edits)                  | 10 Notebooks | ~93 Notebooks |
| ---------------------------------------- | ------------ | ------------- |
| `Reconciliation loop started` / 30s      | **0**        | **600** (~20/s) |
| `Updating Notebook CR Status` / 30s      | 0            | **486**       |
| `Updating StatefulSet` / 60s             | 0            | **484**       |
| Distinct Notebooks reconciled / 20s      | 0            | **90**        |
| Σ `resourceVersion` delta / 30s          | **0**        | **+79,206**   |
| Single idle Notebook `resourceVersion`   | frozen       | climbing      |
| Single idle Notebook `Ready.lastProbeTime` | frozen (e.g. `16:00:49Z` while clock ran to `16:06`) | advances each second (`16:32:20Z → 16:32:51Z`) |

At 10 Notebooks the fleet is silent within ~80s of stabilizing. At ~100 it reconciles
continuously and forever with no user changes — the reproduced bug.

### Direct proof that each reconcile does a needless write (works at any scale)

Poking a pod (forcing one Pod-watch event) triggers one reconcile and one status write:

```
poke 1 (16:11:26): Ready.lastProbeTime=16:11:23Z  rv=25858
poke 2 (16:11:29): Ready.lastProbeTime=16:11:26Z  rv=25894
poke 3 (16:11:32): Ready.lastProbeTime=16:11:29Z  rv=25928
```

### Nuance: StatefulSet *revision* numbers did **not** increase here

The PR also lists "StatefulSet revision numbers increase." On this cluster that specific
symptom did **not** reproduce (busybox and `1.0Gi` `jupyter-scipy` Notebooks alike):

```
repro-50: sts generation=1, controllerrevisions=1
real-0:   sts generation=1, controllerrevisions=1
```

`CopyStatefulSetFields` returns `true` every reconcile (so `r.Update(sts)` is called — 484
times / 60s), but writing the un-defaulted spec re-defaults to the **same** stored bytes,
so the apiserver treats it as a no-op: `generation` stays 1 and no new `ControllerRevision`
is cut. The wasteful API call is real (and is what the fix's SHA-256 template-hash +
semantic comparison eliminates), but revision growth needs a field whose re-defaulting is
non-idempotent, which neither spec here triggers. Worth clarifying in the PR description.

### Control: real Jupyter Notebooks behave identically

3 × `ghcr.io/kubeflow/kubeflow/notebook-servers/jupyter-scipy:v1.11.0` (0.5 CPU / 1.0Gi)
were created. Once stable they were just as silent as busybox (Σ resourceVersion delta 0
over 90s, 0 reconcile logs) — confirming the driver is **scale**, not the image.

## Cleanup

```bash
kubectl delete notebooks -n kubeflow-user-example-com \
  $(kubectl get notebooks -n kubeflow-user-example-com -o name | grep -E 'repro-|real-')
```
Deleting the Notebooks cascades to their StatefulSets / Services / pods; the controller log
goes quiet afterward.

</details>

Issue #1112, opened 2026-05-20
[TASK] Duplicate useWorkspaceKinds calls in WorkspaceForm


thaorell:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

On load, the WorkspaceKind Form component calls `listWorkspaceKinds` multiple times (see screenshot). We need to discover a way to load the form more efficiently without the duplicate calls

<img width="1193" height="679" alt="Image" src="https://github.com/user-attachments/assets/28b27a5c-3a1a-4c75-bdb5-9794c77c1952" />

### Acceptance Criteria

- [ ] WorkspaceForm loads with fewer (ideally 1) call to the API
- [ ] Tests pass

PR #1344, opened 2026-08-21
fix: stabilize workspace kind ordering

thisis-Shitanshu:
## What

Sort WorkspaceKinds by `displayName` before rendering them in the Create Workspace form, using `name` as a deterministic tie-breaker.

## Why

The WorkspaceKinds API does not guarantee a presentation order, so the order of cards on the Create Workspace page could change between page refreshes.

The frontend owns presentation sorting for WorkspaceKinds, so the selection list should apply a deterministic default order before rendering.

## Testing

- Added unit coverage for ordering WorkspaceKinds by `displayName`
- Added coverage for deterministic ordering when display names are equal
- 454/454 Jest tests pass
- Create Workspace Cypress suite passes 73/73 tests
- Manually verified across repeated page refreshes that the UI consistently renders:
  - JupyterLab
  - RStudio
  - VS Code

  even when the WorkspaceKinds API returns the items in different orders

The full mocked Cypress run has one unrelated failure in `workspaces/secretsEdit.cy.ts`. The same failure is reproducible on a clean `origin/notebooks-v2` checkout.

## Related

related: #336
related: #1112

#336 documents the decision to keep paging, sorting, and filtering in the frontend.

#1112 separately tracks duplicate WorkspaceKind requests and is not addressed by this change.

### DIFF
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/kind/WorkspaceFormKindList.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/kind/WorkspaceFormKindList.tsx
@@ -49,9 +49,17 @@ export const WorkspaceFormKindList: React.FunctionComponent<WorkspaceFormKindLis
   const { filterValues, setFilter, clearAllFilters } =
     useToolbarFilters<KindFilterKey>(filterConfig);
 
+  const sortedWorkspaceKinds = useMemo(
+    () =>
+      [...allWorkspaceKinds].sort(
+        (a, b) => a.displayName.localeCompare(b.displayName) || a.name.localeCompare(b.name),
+      ),
+    [allWorkspaceKinds],
+  );
+
   const filteredWorkspaceKinds = useMemo(
-    () => applyFilters(allWorkspaceKinds, filterValues, filterableProperties),
-    [allWorkspaceKinds, filterValues],
+    () => applyFilters(sortedWorkspaceKinds, filterValues, filterableProperties),
+    [sortedWorkspaceKinds, filterValues],
   );
 
   const onChange = useCallback(
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/kind/__tests__/WorkspaceFormKindList.spec.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/kind/__tests__/WorkspaceFormKindList.spec.tsx
@@ -1,0 +1,80 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+import { buildMockWorkspaceKind } from '~/shared/mock/mockBuilder';
+import { WorkspaceFormKindList } from '~/app/pages/Workspaces/Form/kind/WorkspaceFormKindList';
+
+// For this test, we're using mock WorkspaceKindImage, because image loading and API context are unrelated to the behavior we are testing.
+
+jest.mock('~/app/components/WorkspaceKindImage', () => ({
+  __esModule: true,
+  default: ({ children }: { children: (src: string) => React.ReactNode }) =>
+    children('mock-logo.svg'),
+}));
+
+describe('WorkspaceFormKindList', () => {
+  it('renders workspace kinds sorted by display name', () => {
+    const vscode = buildMockWorkspaceKind({
+      name: 'codeserver',
+      displayName: 'VS Code',
+    });
+
+    const jupyter = buildMockWorkspaceKind({
+      name: 'jupyterlab',
+      displayName: 'JupyterLab',
+    });
+
+    const rStudio = buildMockWorkspaceKind({
+      name: 'rstudio',
+      displayName: 'RStudio',
+    });
+
+    render(
+      <WorkspaceFormKindList
+        allWorkspaceKinds={[vscode, jupyter, rStudio]}
+        selectedKind={undefined}
+        onSelect={jest.fn()}
+        isSelectionDisabled={false}
+      />,
+    );
+
+    const cards = screen.getAllByTestId(/^kind-card-/);
+
+    // We aren't testing .sort() itself. We're testing the user-visible guarantee.
+
+    expect(cards.map((card) => card.getAttribute('data-testid'))).toEqual([
+      'kind-card-jupyterlab',
+      'kind-card-rstudio',
+      'kind-card-codeserver',
+    ]);
+  });
+
+  it('uses name as a tie-breaker when display names are equal', () => {
+    const second = buildMockWorkspaceKind({
+      name: 'jupyter-z',
+      displayName: 'JupyterLab',
+    });
+
+    const first = buildMockWorkspaceKind({
+      name: 'jupyter-a',
+      displayName: 'JupyterLab',
+    });
+
+    render(
+      <WorkspaceFormKindList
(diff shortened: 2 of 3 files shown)


thisis-Shitanshu:
@christian-heusel @andyatmiami  PR is up: #1344. Thanks again for the guidance and for helping me validate the direction before implementation. I kept the change focused on deterministic frontend ordering and added regression coverage for the default sort and tie-breaker. Would appreciate your review when you get a chance!

workspaces/frontend/src/app/pages/Workspaces/Form/kind/WorkspaceFormKindList.tsx
```diff
@@ -49,9 +49,17 @@ export const WorkspaceFormKindList: React.FunctionComponent<WorkspaceFormKindLis
   const { filterValues, setFilter, clearAllFilters } =
     useToolbarFilters<KindFilterKey>(filterConfig);
 
+  const sortedWorkspaceKinds = useMemo(
+    () =>
+      [...allWorkspaceKinds].sort(
+        (a, b) => a.displayName.localeCompare(b.displayName) || a.name.localeCompare(b.name),
```
christian-heusel, line 55: What is your reasoning for using `localeCompare()` here? I personally would be happy to have just a regular sort here (but people who are more frontend-versed might have different opinions) 😊

workspaces/frontend/src/app/pages/Workspaces/Form/kind/WorkspaceFormKindList.tsx
christian-heusel, line 1: Why did you opt to add the fix add the presentation layer instead of the data layer? 😊 the `useWorkspaceKinds()` hook is also used in other parts of the code, did you audit them if they suffer from the same issue? Some have other sorting mechanisms, but might be sensible to look at 😊

workspaces/frontend/src/app/pages/Workspaces/Form/kind/__tests__/WorkspaceFormKindList.spec.tsx
```diff
@@ -0,0 +1,82 @@
+// cspell:ignore rstudio
```
christian-heusel: Please move this into the global cspell config file: `workspaces/frontend/config/cspell-ignore-words.txt`

christian-heusel (changes requested):
@thisis-Shitanshu great job on your first PR to this repository 🎉 

I briefly reviewed your changes, see the comments below 😊

workspaces/frontend/src/app/pages/Workspaces/Form/kind/WorkspaceFormKindList.tsx
thisis-Shitanshu: Thanks for taking the time to review my PR, Chris! 😊

So here I used `localeCompare()` mainly because `displayName` is user-facing text, and I noticed the existing string sorting in `WorkspaceKinds.tsx` and `WorkspaceTable.tsx` already uses the same pattern, so I followed that convention.

It also keeps the comparator concise and easy to read, which I think is a small developer-experience and maintainability benefit compared with spelling out the string comparisons manually.

There isn't a specific locale-sensitive requirement here though, so I don't feel strongly about it, and I'm happy to simplify it if we'd prefer a different convention.

workspaces/frontend/src/app/pages/Workspaces/Form/kind/WorkspaceFormKindList.tsx
thisis-Shitanshu: Very good point. My initial reason for keeping this in `WorkspaceFormKindList` came partly from the earlier discussion with @andyatmiami about whether the PatternFly container should own the ordering. We found that `Gallery` itself doesn't own ordering, and the adjacent Image and PodConfig list components also prepare their ordering before rendering, so I treated this as frontend list/presentation preparation.

After your comment, I audited all production consumers of `useWorkspaceKinds()`:

- `WorkspaceFormKindSelection` passes the result into the `WorkspaceKind` card list, which is the flow affected by this issue.
- `WorkspaceKinds.tsx` applies its own sorting based on the active table sort.
- `WorkspaceTable.tsx` only uses the result to build the kind/logo lookup.
- `useWorkspaceFormData.ts` uses it for lookup, so ordering is irrelevant there.

So currently only the **Create Workspace** flow depends on this default ordering, while one consumer replaces the ordering and the other two are order-insensitive. There will be a small redundant sort for `WorkspaceKinds.tsx` if we move this into the hook, but here that cost is negligible.

This still seems consistent with the earlier discussion with Andy because the ordering remains in the frontend data-preparation layer rather than being delegated to PatternFly.

I also noticed the duplicate `useWorkspaceKinds()` / API requests again while auditing this. That's already tracked separately in #1112, so I'd keep that out of this PR and will look at it independently.

workspaces/frontend/src/app/pages/Workspaces/Form/kind/__tests__/WorkspaceFormKindList.spec.tsx
thisis-Shitanshu: Done, moved rstudio into the global `workspaces/frontend/config/cspell-ignore-words.txt` file and removed the file-level cspell ignore. Thanks for pointing me to that, I missed it 😄

workspaces/frontend/src/app/pages/Workspaces/Form/kind/WorkspaceFormKindList.tsx
christian-heusel: If there's already a precedent for using it im fine with that, let's keep it ✅

thaorell: I would say just sorting by `displayName` is enough, since it is a required field and the user does not see `name` on the UI

thaorell (changes requested):
a minor comment below

workspaces/frontend/src/app/pages/Workspaces/Form/kind/WorkspaceFormKindList.tsx
thisis-Shitanshu: @thaorell, I used `name` only as a deterministic tie-breaker rather than as something user-facing. Since `displayName` is required but, as far as I can tell, not necessarily unique, two WorkspaceKinds with the same `displayName` would compare equal and retain their incoming API order. Since the API ordering is what was varying in the original issue, the `name` tie-breaker makes that case deterministic as well.

If duplicate `displayName`s aren't something we need to account for here though, I'm happy to simplify it to just `displayName` 😊

MattBetancourt: Resolved in commit `c3867dd3505953be581513e894ddc931148f60fe`. Section 6.1 color legend has been updated to precisely match the color tokens drawn by the diagnostic render code: cyan for datums/centerline, green for door cutout, yellow for windshields, red for headlights/taillights, and amber for anticlimber.

thaorell (approved):

andyatmiami (approved):
/lgtm
/approve

thanks for review @thaorell ... changes are simple enough (and well documented/tested) such that I feel comfortable approving these changes even though they are `frontend`-related.

PR #1348, opened 2026-08-24
refactor: replace deprecated Result.Requeue in controller reconcilers

christian-heusel:
`controller-runtime` deprecates `reconcile.Result.Requeue` in favor of
`RequeueAfter` / returning an error. The dependency bump in #1306 moves
the controller to `controller-runtime v0.24.1`, where `staticcheck`
flags **every** `Result{Requeue: true}` usage as `SA1019`:

    This setting is deprecated as it causes confusion and there is no
    good reason to use it. When waiting for an external event to happen,
    either the duration until it is supposed to happen or an appropriate
    poll interval should be used, rather than an interval emitted by a
    ratelimiter whose purpose it is to control retry on error.
    Deprecated: Use `RequeueAfter` instead.

## Approach

The reconciler used `Requeue: true` in two distinct situations, handled differently:

| Situation | Count | Old | New | Rationale |
|---|---|---|---|---|
| `IsConflict(err)` retry after an update | 14 | `return Result{Requeue: true}, nil` | `return Result{}, err` | A conflict *is* an error; returning it lets the workqueue rate-limiter handle retry-on-error — exactly what the deprecation note says that limiter is for. Preserves the identical rate-limited/backoff behavior. |
| Stale cache (owned ServiceAccount not yet cached) | 2 | `return Result{Requeue: true}, nil` | `return ctrl.Result{RequeueAfter: requeueAfterStaleCache}, nil` | We already `Owns(&ServiceAccount{})`, so the owned-object watch will requeue us — the existing code comments already say so. No explicit requeue needed. |

`mergeReconcileResult()` no longer inspects the deprecated `.Requeue` field and merges purely on `RequeueAfter`; its unit test was updated to match.

## Options considered

| Option | Requeue timing | Logging | Notes |
|---|---|---|---|
| **A. return the error** (this PR) | immediate, rate-limited backoff — **identical** to `Requeue: true` | conflict now logged at **error** level by controller-runtime | Matches the deprecation's stated rationale; smallest behavioral change to *timing*. |
| B. `RequeueAfter: <small const>` | fixed delay, **no** exponential backoff | stays quiet (returns `nil`) | Changes retry cadence; needs a somewhat arbitrary constant. |
| C. `//nolint:staticcheck` | unchanged | unchanged | Keeps the deprecated API; just silences the linter. |

## Discussion points

1. **Logging noise (main one).** With option A, benign optimistic-concurrency conflicts (previously logged only at `V(2)` and returned as `nil`) now surface via controller-runtime's `"Reconciler error"` at **error** level (and bump the error metric). Is that acceptable, or do we prefer option B (`RequeueAfter`) to keep conflicts quiet at the cost of changing the retry cadence?
2. **Stale-cache sites.** Is relying solely on the `Owns(&ServiceAccount{})` watch to requeue acceptable, or should we rather keep an explicit short `RequeueAfter` as a safety net?
3. **Scope.** Should this also convert `k8s.io/apimachinery/pkg/util/httpstream` → `k8s.io/streaming/pkg/httpstream` (the other `SA1019` the bump surfaces), or keep that as a separate PR?

related: #1306

### DIFF
--- a/workspaces/controller/internal/controller/workspace_controller.go
+++ b/workspaces/controller/internal/controller/workspace_controller.go
@@ -63,6 +63,11 @@ const (
 	// pod template constants
 	workspacePodTemplateContainerName = "main"
 
+	// requeue delay when the local cache is known to be stale (fixed delay,
+	// since a stale cache is not write contention, so no exponential backoff
+	// is needed)
+	requeueAfterStaleCache = 250 * time.Millisecond
+
 	// lengths for resource names
 	generateNameSuffixLength    = 6
 	nameHashLength              = 8
@@ -292,9 +297,10 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 				existingServiceAccount := &corev1.ServiceAccount{}
 				if getErr := r.Get(ctx, client.ObjectKeyFromObject(serviceAccount), existingServiceAccount); getErr != nil {
 					if apierrors.IsNotFound(getErr) {
-						// the cache is stale, the watch on owned ServiceAccounts will requeue us
-						log.V(2).Info("ServiceAccount already exists but is not in the cache yet, relying on the owned-object watch to requeue")
-						return ctrl.Result{}, nil
+						// the cache is stale; requeue after a short delay instead of relying on the
+						// owned-object watch, which will not fire if the ServiceAccount is owned by another controller
+						log.V(2).Info("ServiceAccount already exists but is not in the cache yet, will requeue")
+						return ctrl.Result{RequeueAfter: requeueAfterStaleCache}, nil
 					}
 					log.Error(getErr, "unable to get existing ServiceAccount")
 					return ctrl.Result{}, getErr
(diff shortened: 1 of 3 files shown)


christian-heusel:
/cc @siyuanfoundation @thesuperzapper
Would be good to get some input from you on the discussion points & options given your experience in controller design & code 🤗 
As noted above, the code itself is not intended to merge (given this is a topic that I am not familiar with), I just wanted to have something as a conversation starter 😄

siyuanfoundation:
I think for `IsConflict(err) retry after an update`, it is pretty standard to do `return Result{}, err`, but for stale cache, "We already Owns(&ServiceAccount{}), so the owned-object watch will requeue us" is not necessarily true if the same name `ServiceAccount` is owned by some other controller, in that case it is better to use `RequeueAfter: <small const>` to have requeue guarantee. There is no need for exponential back off because stale cache is pretty common and not a write contention (you just wait)

siyuanfoundation:
the code LGTM.
Can you update the PR description? Because the solution for `Stale cache` has changed.

Thanks!

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -63,6 +63,10 @@ const (
 	// pod template constants
 	workspacePodTemplateContainerName = "main"
 
+	// requeue delay when the local cache is known to be stale (fixed delay, since a stale cache
+	// is not write contention, so no exponential backoff is needed)
+	requeueAfterStaleCache = 2 * time.Second
```
thesuperzapper: I think 2 seconds is a bit long when we are only waiting for the list-watch cache to update, lets start with 250ms (which is still 250ms longer than the current requeue), and see if that works.

thesuperzapper (changes requested):

workspaces/controller/internal/controller/workspace_controller.go
christian-heusel: Applied ✔️

PR #1350, opened 2026-08-24
fix: restore Connect button outline in MUI theme

### LINKED ISSUE #1345: "Connect" button has no outline in Kubeflow Community Distribution deployment
### Checks

- [x] I have searched the [existing issues](https://github.com/kubeflow/notebooks/issues).
- [x] My issue is related to one of the components in the [`kubeflow/notebooks`](https://github.com/kubeflow/notebooks) repository.

### Kubeflow Notebooks Version

v2.0.0-beta.0

### Kubeflow Platform

Manifests repository https://github.com/kubeflow/community-distribution/#install-with-a-single-command

### Kubernetes Distribution

KinD

### Kubernetes Version

```shell
Client Version: v1.36.3
Kustomize Version: v5.8.1
Server Version: v1.35.0
```

### Description

When working on https://github.com/kubeflow/community-distribution/pull/3584 I saw that there currently is a bug with how the connect button is displayed in the context of the Kubeflow Community distribution / when serving Kubeflow Notebooks v2 together with Dashboard.

Problem:
Image

Expected:

Image

### Relevant Logs

```shell

```

thisis-Shitanshu:
## What

Fixes the Workspaces action controls rendering differently when Notebooks v2 is embedded in the Kubeflow Community Dashboard.

Closes #1345

## Root cause

The Connect and row-action controls use PatternFly `MenuToggle` variants.

When comparing the same UI at the same viewport in standalone Notebooks v2 and Community Dashboard, the variant-specific PatternFly tokens were already correct in both environments.

The difference was in which CSS declarations won the cascade:

- standalone: the PatternFly variant mappings win
- Community Dashboard: generic `mod-arch-kubeflow` / MUI compatibility mappings win

For the secondary Connect toggle this changed the final border, background, and horizontal padding.

For the plain row-actions toggle this changed the background and horizontal padding.

### Before

localhost-_-workspaces-workspaces_Before

## Fix

- Add a common class to both Connect variants.
- Restore the existing PatternFly secondary semantic mappings for the Connect control.
- Scope the workspace row-actions toggle through its action cell and restore its existing PatternFly plain mappings.
- Reuse PatternFly theme variables rather than hard-coding colors, spacing, or dimensions.

The overrides remain scoped to the affected Workspaces controls.

### After

localhost-_-workspaces-workspaces_After

## Verification

Compared the same Workspace at a 1111px viewport in:

- standalone Notebooks v2 / Tilt
- Community Distribution baseline
- Community Distribution with this PR

Before, Community resolved:

- Connect border: `transparent`
- Connect padding: `10px`
- Connect background: `#fff`
- row-actions background: `#fff`
- row-actions padding: `10px`

After the fix, the Community controls match the standalone variant semantics:

- Connect width: `134.953125px`
- Connect height: `37px`
- Connect padding: `24px / 24px`
- Connect background: transparent
- Connect border: `#1976d2`
- row-actions: `37px × 37px`
- row-actions padding: `8px / 8px`
- row-actions background: transparent

Hover/focus behavior was also verified.

Checks:

- `npm run prettier:check`
- `npm run test:lint`
- `npm run test:type-check`
- `npm run test:unit` — 452 tests passed
- `npm run build`

### DIFF
--- a/workspaces/frontend/src/app/app.css
+++ b/workspaces/frontend/src/app/app.css
@@ -252,6 +252,33 @@ body,
   margin-left: 0;
 }
 
+/* Preserve secondary MenuToggle styling when using the Kubeflow MUI theme. */
+.mui-theme .kubeflow-connect-toggle.pf-m-secondary {
+  --pf-v6-c-menu-toggle--BackgroundColor: var(--pf-v6-c-menu-toggle--m-secondary--BackgroundColor);
+  --pf-v6-c-menu-toggle--PaddingInlineStart: var(
+    --pf-v6-c-menu-toggle--m-secondary--PaddingInlineStart
+  );
+  --pf-v6-c-menu-toggle--PaddingInlineEnd: var(
+    --pf-v6-c-menu-toggle--m-secondary--PaddingInlineEnd
+  );
+  --pf-v6-c-menu-toggle--BorderColor: var(--pf-v6-c-menu-toggle--m-secondary--BorderColor);
+  --pf-v6-c-menu-toggle--hover--BorderColor: var(
+    --pf-v6-c-menu-toggle--m-secondary--hover--BorderColor
+  );
+  --pf-v6-c-menu-toggle--expanded--BorderColor: var(
+    --pf-v6-c-menu-toggle--m-secondary--expanded--BorderColor
+  );
+}
+
+/* Preserve plain MenuToggle styling for workspace row actions when using the Kubeflow MUI theme. */
+.mui-theme .kubeflow-workspace-actions-cell .pf-v6-c-menu-toggle.pf-m-plain {
+  --pf-v6-c-menu-toggle--BackgroundColor: var(--pf-v6-c-menu-toggle--m-plain--BackgroundColor);
+  --pf-v6-c-menu-toggle--PaddingInlineStart: var(
+    --pf-v6-c-menu-toggle--m-plain--PaddingInlineStart
+  );
+  --pf-v6-c-menu-toggle--PaddingInlineEnd: var(--pf-v6-c-menu-toggle--m-plain--PaddingInlineEnd);
+}
+
 /* Single-endpoint "Connect" control: rendered as a MenuToggle (so it keeps the
    same size as the multi-endpoint dropdown), but its caret is rotated to point
    right to signal that clicking connects directly instead of opening a menu. */
--- a/workspaces/frontend/src/app/components/WorkspaceTable.tsx
+++ b/workspaces/frontend/src/app/components/WorkspaceTable.tsx
@@ -428,7 +428,12 @@ const WorkspaceTable = React.forwardRef<WorkspaceTableRef, WorkspaceTableProps>(
 
                           if (columnKey === 'actions') {
                             return (
-                              <Td isActionCell key="actions" data-testid="action-column">
+                              <Td
+                                isActionCell
+                                key="actions"
+                                data-testid="action-column"
+                                className="kubeflow-workspace-actions-cell"
+                              >
                                 <ActionsColumn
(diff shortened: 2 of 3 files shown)


christian-heusel (changes requested):
What you shared in your screenshot still looks different from the "expected" one in https://github.com/kubeflow/notebooks/issues/1345, could you recheck the differences? From a broad look the spacing / button layout still seems to be different 😅

thisis-Shitanshu:
Hi @christian-heusel!

Thanks for catching that. I rechecked the controls against standalone at the same viewport and found the spacing/layout difference came from the same CSS cascade issue as the missing outline.

The PatternFly variant tokens were already correct, but in the Community Dashboard the generic MUI compatibility mappings were winning instead. This affected the Connect border, background and horizontal padding, and the adjacent plain row-action toggle's background/padding as well.

I've updated the fix to restore the existing PatternFly variant mappings only for these Workspaces controls, without hard-coding colors or dimensions. I also verified the updated Community build against standalone, including hover/focus states, and all frontend checks pass (452 tests).

Would appreciate another look when you have time 😊

workspaces/frontend/src/app/app.css
```diff
@@ -252,6 +252,33 @@ body,
...
+  --pf-v6-c-menu-toggle--hover--BorderColor: var(
+    --pf-v6-c-menu-toggle--m-secondary--hover--BorderColor
+  );
+  --pf-v6-c-menu-toggle--expanded--BorderColor: var(
+    --pf-v6-c-menu-toggle--m-secondary--expanded--BorderColor
+  );
+}
+
+/* Preserve plain MenuToggle styling for workspace row actions when using the Kubeflow MUI theme. */
+.mui-theme .kubeflow-workspace-actions-cell .pf-v6-c-menu-toggle.pf-m-plain {
```
thaorell: `Td isActionCell` already renders PatternFly's built-in `pf-v6-c-table__action` class, so `.mui-theme .pf-v6-c-table__action .pf-v6-c-menu-toggle.pf-m-plain` should work without adding the new `kubeflow-workspace-actions-cell` class/prop. I'd keep the current approach as-is though — reusing the generic PF class would also apply this override to the other `isActionCell` tables (WorkspaceKinds, Secrets, Volumes, Tolerations, Activity Rules), which haven't been verified against the Community Dashboard theme in this PR. Just flagging the trade-off in case a broader fix is intended later; no change requested here.

thaorell (commented):
Minor comment

workspaces/frontend/src/app/app.css
thisis-Shitanshu: Thanks, this is helpful context. During the #1350 investigation, I compared the same Notebooks v2 baseline in standalone and in the Community Dashboard at the same viewport.

What I found was that the PatternFly variant-specific values were already correct in both environments, but the final values differed because the generic MUI `MenuToggle` mappings won the cascade in Community. For the secondary Connect toggle, that changed the border, background, and horizontal padding. I saw the same pattern on the adjacent plain row-action toggle: its plain variant values were correct, but Community resolved the final background and padding differently.

After restoring the variant-specific mappings, the Community controls matched the standalone behavior for the properties I measured, and I also verified hover/focus behavior.

That is why I kept `kubeflow-workspace-actions-cell` deliberately narrow here. This PR only reproduced and validated the issue on the Workspaces table; using the generic `pf-v6-c-table__action` boundary would extend the change to other tables, which I did not verify as part of this PR.

If we later audit those other surfaces and see the same PF/MUI cascade consistently, I agree the generic `pf-v6-c-table__action` boundary would be the better place for a broader fix. I can add that cross-theme audit as a follow-up task under #1395 so we can verify those surfaces systematically later.

thaorell (approved):
/lgtm
/approve

PR #788, opened 2025-12-08
feat: enable kueue integration by ensuing that labels added to Notebooks are propogated to the underlying StatefulSet

kannon92:
Apply stale https://github.com/kubeflow/kubeflow/pull/7674 to this v1 branch.

This was a clean pick of https://github.com/opendatahub-io/kubeflow/pull/646.

It looks like this was merged in opendatahub but I didn't see this patch contributed to kubeflow notebooks.

This should provide integration with Kueue based on https://github.com/kubernetes-sigs/kueue/issues/3878#issuecomment-3628061305.

PR #1351, opened 2026-08-24
feat: add configurable StatefulSet metadata to WorkspaceKind

christian-heusel:
## Motivation

This PR the first step toward **supporting [Kueue](https://kueue.sigs.k8s.io/) in Notebooks v2**.

cc @kannon92

Kueue gates admission of a workload via the `kueue.x-k8s.io/queue-name` label on the managed object — for a Notebook Workspace that object is the **StatefulSet** the controller generates. Today an admin can label/annotate the Workspace **Pods** (`spec.podTemplate.podMetadata`) but has no way to put a label on the StatefulSet itself, so a Workspace cannot be opted into a Kueue
LocalQueue.

This PR adds an admin-only `spec.podTemplate.statefulSetMetadata` (labels + annotations) that is applied to the generated StatefulSet's own metadata. Beyond Kueue it is also useful for cost allocation, GitOps/ArgoCD tracking, and org-wide labeling conventions.

The field is optional and omitted by default, so existing WorkspaceKinds are completely unaffected.

## Related work & roadmap

Label/metadata support for Notebooks v2, so workloads can integrate with systems
like Kueue:

| Where | `podMetadata` (Pods) | `statefulSetMetadata` (StatefulSet) |
|-------|----------------------|-------------------------------------|
| `WorkspaceKind.spec.podTemplate` | ✅ already exists | ✅ **added by this PR** |
| `WorkspaceKind.spec.podTemplate.options.podConfig[].spec` | ⬜ planned | ⬜ planned |

The per-`podConfig` variants (letting a specific pod config carry its own
metadata, e.g. routing only GPU pod configs to a queue) are planned as
follow-ups and are **not** part of this PR.

### References

- v1 support request: https://github.com/kubeflow/notebooks/pull/788
- Kueue upstream (docs for using Kueue with Notebooks v2):
  https://github.com/kubernetes-sigs/kueue/issues/3878#issuecomment-3628061305

## Design notes

- **Admin-only, WorkspaceKind-scoped.** Unlike `podMetadata`, there is no
  per-Workspace override and no per-`podConfig` variant (yet) — it is a single
  default sourced from the WorkspaceKind.
- **Controller labels stay authoritative.** Admin labels are merged with
  `labels.Merge` so `notebooks.kubeflow.org/workspace-name` always wins and
  cannot be overridden.
- **Reconcile fix included.** `copyLabelFields`/`copyAnnotationFields` only
  detected changed/removed keys, never purely *added* ones, so a label added to
  a WorkspaceKind would not propagate to already-running StatefulSets until some
  other field changed. Both helpers now also iterate the desired map. (These are
  shared with `CopyServiceFields`, which benefits from the same correction.)
- The backend create/update models already carry the field (they embed the
  controller's `WorkspaceKindPodTemplate` verbatim); this PR also surfaces it in
  the hand-written **read/list** response model so `GET /workspacekinds` returns
  it.

## Not included / follow-up

- Per-`podConfig` `podMetadata` / `statefulSetMetadata` support (see roadmap).
- Frontend `npm run generate:api` (regenerating `src/generated/` from the new
  swagger) — intentionally left as a separate follow-up.


workspaces/controller/api/v1beta1/workspacekind_types.go
```diff
@@ -331,6 +335,16 @@ type WorkspaceKindPodMetadata struct {
 	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
+type WorkspaceKindStatefulSetMetadata struct {
```
andyatmiami, line 340:
```suggestion
// WorkspaceKindStatefulSetMetadata defines labels and annotations applied to the Workspace StatefulSet.
type WorkspaceKindStatefulSetMetadata struct {
```

workspaces/controller/internal/controller/workspace_controller.go
andyatmiami, line 1: **Missing webhook validation for `statefulSetMetadata` labels and annotations** — `workspaces/controller/internal/webhook/workspacekind_webhook.go` (not changed — but should have been)
- didn't have a better way to call this out 😇 

The structurally identical `podMetadata` field has webhook validation that calls `v1validation.ValidateLabels` and `apivalidation.ValidateAnnotations` in both `ValidateCreate` and `ValidateUpdate`:

```go
// ValidateCreate
allErrs = append(allErrs, v.validatePodTemplatePodMetadata(workspaceKind)...)

// ValidateUpdate
if !equality.Semantic.DeepEqual(newWorkspaceKind.Spec.PodTemplate.PodMetadata, oldWorkspaceKind.Spec.PodTemplate.PodMetadata) {
    allErrs = append(allErrs, v.validatePodTemplatePodMetadata(newWorkspaceKind)...)
}
```

The new `statefulSetMetadata` has the same structure (`map[string]string` labels and annotations) but no corresponding validation. Without it, an admin can set invalid label keys (e.g., containing spaces) or invalid annotation values that the API server will reject when the controller tries to create/update the StatefulSet, putting the Workspace into an error state with no clear indication that the WorkspaceKind itself is misconfigured.

workspaces/backend/api/suite_test.go
andyatmiami, line 1: **No test coverage for nil `StatefulSetMetadata` (default values path)** — 

The CRD field `StatefulSetMetadata` is an optional pointer (`*WorkspaceKindStatefulSetMetadata`). The conversion in `funcs.go` correctly handles `nil` by initializing empty maps, but `NewExampleWorkspaceKind` always sets it to a non-nil value with populated maps. No test verifies the nil-to-empty-maps path.

workspaces/backend/api/workspacekinds_handler_test.go
```diff
@@ -146,6 +146,16 @@ var _ = Describe("WorkspaceKinds Handler", func() {
 				models.NewWorkspaceKindModelFromWorkspaceKind(a.Config, workspacekind2),
 			))
 
+			By("ensuring the statefulSetMetadata from the WorkspaceKind is surfaced in the response")
+			var item1 models.WorkspaceKindListItem
+			for _, wsk := range response.Data {
+				if wsk.Name == workspaceKind1Name {
+					item1 = wsk
+				}
+			}
```
andyatmiami, line 155: **Missing guard assertion in test lookup loop** 

If `workspaceKind1Name` doesn't match any item, `item1` stays zero-valued and subsequent assertions fail with a confusing empty-map message. Consider adding a failesafe a la:

```
Expect(item1.Name).To(Equal(workspaceKind1Name), "WorkspaceKind %q not found in response", workspaceKind1Name)
```

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -999,6 +999,15 @@ func generateStatefulSet(workspace *kubefloworgv1beta1.Workspace, workspaceKind
...
 	}
 
+	// generate statefulset metadata
+	// NOTE: statefulset metadata is only configurable at the WorkspaceKind level
+	stsAnnotations := make(map[string]string)
+	stsLabels := make(map[string]string)
+	if workspaceKind.Spec.PodTemplate.StatefulSetMetadata != nil {
+		maps.Copy(stsAnnotations, workspaceKind.Spec.PodTemplate.StatefulSetMetadata.Annotations)
+		maps.Copy(stsLabels, workspaceKind.Spec.PodTemplate.StatefulSetMetadata.Labels)
+	}
```
andyatmiami, line 1009:
```suggestion
    var stsAnnotations, stsLabels map[string]string
    if m := workspaceKind.Spec.PodTemplate.StatefulSetMetadata; m != nil     {
        stsAnnotations = maps.Clone(m.Annotations)
        stsLabels = maps.Clone(m.Labels)
    }
```

I think we can express this a little more compactly - albeit its certainly a "nit" comment

⚠️ coding in GH - beware me screwing up whitespace

workspaces/backend/internal/models/workspacekinds/funcs.go
```diff
@@ -39,6 +39,14 @@ func NewWorkspaceKindModelFromWorkspaceKind(cfg *config.EnvConfig, wsk *kubeflow
 		maps.Copy(podAnnotations, wsk.Spec.PodTemplate.PodMetadata.Annotations)
 	}
 
+	stsLabels := make(map[string]string)
+	stsAnnotations := make(map[string]string)
+	if wsk.Spec.PodTemplate.StatefulSetMetadata != nil {
+		// NOTE: we copy the maps to avoid creating a reference to the original maps.
+		maps.Copy(stsLabels, wsk.Spec.PodTemplate.StatefulSetMetadata.Labels)
+		maps.Copy(stsAnnotations, wsk.Spec.PodTemplate.StatefulSetMetadata.Annotations)
+	}
```
andyatmiami, line 48: See [this comment](https://github.com/kubeflow/notebooks/pull/1351/changes#r3883817952) for a "nit suggestion"

andyatmiami (changes requested):
@christian-heusel - thanks for getting this necessary PR raised.

minor nitpicks and suggestions - but nothing here that concerning (outside the alignment of the webhook validation)

let me know if you have any issue/disagreement with my comments. i'd expect we can get this merged after your next push 💯

workspaces/controller/internal/controller/workspace_controller.go
christian-heusel: The whole reason why I did it this way is that it is in line with the implementation for the podLabels / podAnnotations:
https://github.com/kubeflow/notebooks/blob/9e93648fd17e889caba8e2515f007ff550d73619/workspaces/backend/internal/models/workspacekinds/funcs.go#L34-L40

Do we want to keep things consistent or should I refactor this for the pod* sibling implemntations where it shows up aswell? 😊 
```
$ rg "maps.Copy"
[...]
workspaces/backend/internal/repositories/workspacekinds/repo.go
227:    maps.Copy(labels, namespace.Labels)

workspaces/controller/internal/controller/workspace_controller.go
991:            maps.Copy(podAnnotations, workspaceKind.Spec.PodTemplate.PodMetadata.Annotations)
992:            maps.Copy(podLabels, workspaceKind.Spec.PodTemplate.PodMetadata.Labels)
995:            maps.Copy(podAnnotations, workspace.Spec.PodTemplate.PodMetadata.Annotations)
996:            maps.Copy(podLabels, workspace.Spec.PodTemplate.PodMetadata.Labels)

workspaces/controller/internal/controller/workspace_activity.go
336:    maps.Copy(result, ns.Labels)

workspaces/backend/internal/models/workspacekinds/funcs.go
38:             maps.Copy(podLabels, wsk.Spec.PodTemplate.PodMetadata.Labels)
39:             maps.Copy(podAnnotations, wsk.Spec.PodTemplate.PodMetadata.Annotations)

workspaces/backend/internal/models/workspaces/common/funcs.go
35:             maps.Copy(podLabels, ws.Spec.PodTemplate.PodMetadata.Labels)
36:             maps.Copy(podAnnotations, ws.Spec.PodTemplate.PodMetadata.Annotations)
```

andyatmiami: we should definitely be consistent - sorry for (stupidly) overlooking the fact this comment had a "consistency impact"

I'm indifferent on outcome here - aside from your point on consistency... if you like this (marginally more) compact syntax and wanna upgrade all call sites... **great**

if you don't care and wanna leave this code **as-is** - so its consistent with codebase - **also great**

dealer's choice - thx for pushback/providing your justification

workspaces/controller/api/v1beta1/workspacekind_types.go
christian-heusel: Added, also added something to the pod-equivalent of this (which also didn't have a comment) 🤗

workspaces/controller/internal/controller/workspace_controller.go
christian-heusel: Nice catch, I adressed it in 2b9b0f36152a3f96d180dd053e22ce2892c01fdf ✅

workspaces/backend/api/suite_test.go
christian-heusel: Adressed in 7de3ba8796d9e62e3d172864eb9e18fe8f5d6d4b, thanks for pointing it out! ✅

workspaces/controller/internal/controller/workspace_controller.go
christian-heusel: Given the 🤖 analysis below I think I'll leave this as is since it unfortunately matches what I expected (no free lunch / cleanup here 😢) :

```
I checked every place this make + maps.Copy pattern appears, and Andy's maps.Clone form does not apply cleanly to all of them — so per your instruction, I've left it unimplemented. Here's why:

The critical difference: make(map...) always yields a non-nil empty map, whereas maps.Clone(nil) returns nil. That distinction matters at some callsites.

The pattern lives in 4 spots:

┌────────────────────────────────────┬────────────────┬───────────────────────────────────────────────────────────────────────────┐
│              Callsite              │  Clean with    │                                  Reason                                   │
│                                    │  maps.Clone?   │                                                                           │
├────────────────────────────────────┼────────────────┼───────────────────────────────────────────────────────────────────────────┤
│ controller                         │                │ single source; result feeds ObjectMeta.Annotations and labels.Merge(...), │
│ workspace_controller.go:1004 sts   │ ✅ yes         │  where nil vs empty is functionally identical to k8s                      │
│ metadata                           │                │                                                                           │
├────────────────────────────────────┼────────────────┼───────────────────────────────────────────────────────────────────────────┤
│ controller                         │                │ merges two sources (WorkspaceKind then Workspace override). Can't be one  │
│ workspace_controller.go:991 pod    │ ❌ no          │ maps.Clone; and maps.Clone of the first source could return nil, so the   │
│ metadata                           │                │ follow-up maps.Copy(nil, …) would panic                                   │
├────────────────────────────────────┼────────────────┼───────────────────────────────────────────────────────────────────────────┤
│ backend funcs.go:34 pod metadata   │ ❌ no          │ nil-vs-empty is observable: a nil map marshals to null instead of {} in   │
│                                    │                │ the API response                                                          │
├────────────────────────────────────┼────────────────┼───────────────────────────────────────────────────────────────────────────┤
│                                    │                │ same JSON null vs {} issue — and it would directly break the non-nil test │
│ backend funcs.go:42 sts metadata   │ ❌ no          │  we just added (Expect(...).NotTo(BeNil())), which existed because of     │
│                                    │                │ Andy's own earlier review comment                                         │
└────────────────────────────────────┴────────────────┴───────────────────────────────────────────────────────────────────────────┘

So the suggestion is a valid nit only for the one controller sts block Andy commented on. Applying it there alone would leave the sibling pod block right above it using the old style (inconsistent), and it can't be propagated to the other three without either a panic risk or changing observable JSON output and contradicting the nil-path coverage from the previous comment.
```

Good call to point this out though, I'll pay attention to it in the future if I come across this pattern or other things that can be simplified 🤗

christian-heusel:
Adressed the feedback, thanks for the review 🤗

andyatmiami (approved):
/lgtm
/approve

---

thanks @christian-heusel for this nice addition.  comfortable with change set as it faithfully follows the `WorkspaceKindPodMetadata` implementation ( **as it should** ).

---

testing methodology:
- [PR-1351-TEST-RESULTS.md](https://github.com/user-attachments/files/31745849/PR-1351-TEST-RESULTS.md)

PR #1352, opened 2026-08-25
fix: set lastActivity when it is never set and the probe returns false

siyuanfoundation:
related: #867

Fix the edge case Raised by Andy
```
edge case:

A workspace whose probe consistently reports has_activity: false (and never provides last_activity) will accumulate infinite running time without ever being paused, even though the probe is explicitly saying "no activity detected."

The intent of has_activity: false preserving existing lastActivity makes sense for the transition case: the workspace was active,  then becomes inactive — you want to pause based on when activity stopped, which is the last known lastActivity. But when
lastActivity was never set (stays at 0), the safety guard ("unknown, don't pause") creates a dead end: the probe keeps saying "no last_activity) will accumulate infinite running time without ever being paused, even though the probe is explicitly saying "no activity detected."
```

The fix set `lastActivity` to`lastRunningTime` when it is never set and the probe returns false


christian-heusel:
> /ok-to-test

@siyuanfoundation you shoudln't need to run this now that you're an org member 🤗

andyatmiami (approved):
/lgtm
/approve

---

Thanks @siyuanfoundation for the quick turnaround here... realize its an unlikely use case - but good to be safeguarded against nonetheless!

Implementation is clean, compact, well-documented, and resolves the reported issue.

---

testing methodology:
- [PR-1352-TEST-RESULTS.md](https://github.com/user-attachments/files/31555527/PR-1352-TEST-RESULTS.md)

PR #1354, opened 2026-08-27
docs: document frontend styling and integration

thisis-Shitanshu:
## What

Add frontend documentation describing the styling layers used by Kubeflow Workspaces and considerations when rendering the UI standalone versus through the Kubeflow integration.

The new section documents:

- PatternFly component and semantic styling
- `mod-arch-kubeflow` / MUI compatibility styling
- Workspaces-local compatibility overrides
- cross-theme debugging guidelines
- verification in standalone and Kubeflow-integrated environments

## Why

While investigating #1345 and validating the fix in #1350, I found that a component could have the correct PatternFly variant variables while still rendering differently between standalone Workspaces and the Central Dashboard integration because different CSS declarations won the cascade.

Documenting these styling layers and a small debugging workflow should give frontend contributors a clearer starting point for similar integration issues.

 related: #1345
 related: #1350


christian-heusel (approved):
LGTM; Lets see what the actual frontend reviewers have to say about it 🤗

PR #1358, opened 2026-08-28
feat: add Kind, Image Config, and Pod Config printer columns to Workspace CRD

christian-heusel:
## What this does

Enriches the `kubectl get workspaces` and k9s output by adding three `additionalPrinterColumns` to the `Workspace` CRD, surfacing the same information the frontend workspace table already shows:

- **Kind**: the `WorkspaceKind` (`.spec.kind`)
- **Image Config**: the selected imageConfig option id (`.spec.podTemplate.options.imageConfig`)
- **Pod Config**: the selected podConfig option id (`.spec.podTemplate.options.podConfig`)

The existing **State** column is kept. All columns show in the default output (no `-o wide` needed).

## Before

```
$ kubectl get workspaces
NAME                   STATE
jupyterlab-workspace   Running
test                   Running
```

## After

```
$ kubectl get workspaces
NAME                   KIND         IMAGE CONFIG                POD CONFIG   STATE
jupyterlab-workspace   jupyterlab   jupyter-scipy:v1.10.0       tiny_cpu     Running
test                   codeserver   codeserver-python:v1.10.0   tiny_cpu     Running
```

## Notes

- The last activity was intentionally omitted: it is stored as epoch milliseconds, which kubectl's `type=date` columns cannot render as a friendly relative time.

### DIFF
--- a/workspaces/controller/api/v1beta1/workspace_types.go
+++ b/workspaces/controller/api/v1beta1/workspace_types.go
@@ -378,6 +378,9 @@ const (
 
 // +kubebuilder:object:root=true
 // +kubebuilder:printcolumn:name="State",type="string",JSONPath=".status.state",description="The current state of the Workspace"
+// +kubebuilder:printcolumn:name="Kind",type="string",JSONPath=".spec.kind",description="The WorkspaceKind of the Workspace"
+// +kubebuilder:printcolumn:name="Image Config",type="string",JSONPath=".spec.podTemplate.options.imageConfig",description="The imageConfig option of the Workspace"
+// +kubebuilder:printcolumn:name="Pod Config",type="string",JSONPath=".spec.podTemplate.options.podConfig",description="The podConfig option of the Workspace"
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:shortName=ws
 
--- a/workspaces/controller/manifests/kustomize/base/crd/kubeflow.org_workspaces.yaml
+++ b/workspaces/controller/manifests/kustomize/base/crd/kubeflow.org_workspaces.yaml
@@ -21,6 +21,18 @@ spec:
       jsonPath: .status.state
       name: State
       type: string
+    - description: The WorkspaceKind of the Workspace
+      jsonPath: .spec.kind
+      name: Kind
+      type: string
+    - description: The imageConfig option of the Workspace
+      jsonPath: .spec.podTemplate.options.imageConfig
+      name: Image Config
+      type: string
+    - description: The podConfig option of the Workspace
+      jsonPath: .spec.podTemplate.options.podConfig
+      name: Pod Config
+      type: string
     name: v1beta1
     schema:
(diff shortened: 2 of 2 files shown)


workspaces/controller/api/v1beta1/workspace_types.go
thisis-Shitanshu, line 1: I also feel we need these outputs, Chris, especially after yesterday’s debugging session. I understand that `LastRunningTime`, `PauseTime`, and `LastUpdate` are stored as int, which cannot be rendered as dates. How about having a column for `.metadata.creationTimestamp`, which is already stored as a ts? 😊

christian-heusel: I think since the table view does not include created at we can also omit it here ... I had a lengthy (unfortunately private) chat with Andy regarding the timestamps though, refactoring them to be of type `metav1.Time` is more involved and we're currently unsure whether its desirable anyways so we'll just leave it as is 🤗

workspaces/controller/api/v1beta1/workspace_types.go
```diff
@@ -377,6 +377,9 @@ const (
 */
 
 // +kubebuilder:object:root=true
+// +kubebuilder:printcolumn:name="Kind",type="string",JSONPath=".spec.kind",description="The WorkspaceKind of the Workspace"
+// +kubebuilder:printcolumn:name="Image Config",type="string",JSONPath=".spec.podTemplate.options.imageConfig",description="The imageConfig option of the Workspace"
+// +kubebuilder:printcolumn:name="Pod Config",type="string",JSONPath=".spec.podTemplate.options.podConfig",description="The podConfig option of the Workspace"
```
andyatmiami, line 383: **Consider column ordering vs. `kubectl get` truncation** 🤔 

`ImageConfig`/`PodConfig` values are capped at 256 chars and `Kind` values can also be long. On narrow terminals, this pushes `State` — arguably the most operationally-important column — far right where it may be truncated. Consider putting `State` first (right after `NAME`, per the `kubectl get pods`-style convention where lifecycle status is the primary quick-glance column), then Kind/Image Config/Pod Config as descriptive context. 

Open to pushback here - but i'll raise the thought for your consideration.  otherwise this looks great/nice value add! 💯

andyatmiami (changes requested):
On fence about this - feel free to push back - but I'll surface it to be transparent..

workspaces/controller/api/v1beta1/workspace_types.go
christian-heusel: Sounds good, applied ✔️

andyatmiami (approved):
/lgtm
/merge

**yet again** - appreciate attention to detail and quality of life fixes @christian-heusel 

testing methodology:
- [PR-1358-TEST-RESULTS.md](https://github.com/user-attachments/files/31812222/PR-1358-TEST-RESULTS.md)

PR #1365, opened 2026-08-29
fix: handle contextual pod config defaults

### LINKED ISSUE #1353: bug: Dynamic filter rules on PodConfig cause flickering or incorrectly reveal hidden options
### Checks

- [x] I have searched the [existing issues](https://github.com/kubeflow/notebooks/issues).
- [x] My issue is related to one of the components in the [`kubeflow/notebooks`](https://github.com/kubeflow/notebooks) repository.

### Kubeflow Notebooks Version

v2.0.0-beta.0

### Kubeflow Platform

Kubeflow Distributions

### Kubernetes Distribution

GKE

### Kubernetes Version

```shell
Client Version: v1.36.1
Kustomize Version: v5.8.1
Server Version: v1.35.6-gke.1258000
```

### **What happened?**

When defining a `filterRule` on `WorkspaceKind` to hide non-GPU `podConfigs` when a CUDA image is selected:

1. **Infinite Re-render Loop / UI Flickering with `api.hide: true`:**
   When using both `ui.hide: true` and `api.hide: true` in the rule, selecting a CUDA image in the Spawner UI causes the form to flicker rapidly and get stuck in an infinite state update loop. The user is unable to select the remaining GPU pod config or proceed to create a workspace.

2. **Hidden PodConfigs Automatically Revealed with only `ui.hide: true`:**
   When removing `api.hide: true` (relying only on `ui.hide: true`), user can proceed to create the GPU workspace, but the default CPU pod configs are not hidden in the UI even though the user has not selected "Show hidden". The "Show hidden" filter is automatically enabled.

---

### **What did you expect to happen?**

1. Selecting a CUDA image should cleanly filter the pod configs so that only compatible GPU pod configs are shown (and selectable) without causing UI flickering or infinite re-renders.
2. If `ui.hide: true` is set on CPU configs for a CUDA image, those CPU configs should remain hidden by default unless the user explicitly checks "Show hidden".

---

### **How to reproduce it?**

1. add the following `filterRules` in `kubeflow-notebooks/workspaces/controller/manifests/kustomize/samples/jupyterlab_v1beta1_workspacekind.yaml`:

```yaml
    - scope: POD_CONFIG
      effect:
        ui:
          hide: true
        api:
          hide: true
      match:
        - matchImageConfig:
            selector:
              matchExpressions:
                - key: "cuda_version"
                  operator: Exists
        - matchPodConfig:
            selector:
              matchExpressions:
                - key: "gpu"
                  operator: DoesNotExist
```

2. Open the Workspace Spawner UI (`/workspaces/form/create`).
3. In Step 1, select the `jupyterlab` WorkspaceKind.
4. In Step 2, select the CUDA image (`jupyter-pytorch-cuda-full:v1.10.0`) and click **Next**.
5. **Observe Behavior 1**: The UI starts flickering indefinitely and pod config cards cannot be selected.

Image

6. Remove `api.hide: true` from the `filterRules` and reload the form.
7. Select the CUDA image and proceed to Step 3.
8. **Observe Behavior 2**: "Tiny CPU" is shown in the list by default even though the rule specifies `ui.hide: true` and the user did not check "Show hidden".
---

#### 1. Infinite Loop in `WorkspaceForm.tsx` (Triggered when `api.hide: true`)
- In `WorkspaceForm.tsx`:
  - **Effect 1 (Defaulting)** sets `podConfig` to `allValuesData.podConfig.default` (`tiny_cpu`) whenever `!data.podConfig`:
    ```tsx
    useEffect(() => {
      if (!allValuesData || !allValuesLoaded || !data.kind) return;
      if (!data.podConfig && allValuesData.podConfig.default) {
        setData('podConfig', allValuesData.podConfig.default);
      }
    }, [allValuesData, allValuesLoaded, data.kind, data.imageConfig, data.podConfig, setData]);
    ```
  - **Effect 2 (Validation)** clears `data.podConfig` when `filteredValuesData` arrives and the selected pod config is no longer valid:
    ```tsx
    useEffect(() => {
      if (!filteredValuesLoaded || !filteredValuesData || !data.podConfig) return;
      const podConfigOptions = filteredValuesData.podConfig.values ?? [];
      const isStillValid = podConfigOptions.some((pc) => pc.id === data.podConfig);
      if (!isStillValid) {
        setData('podConfig', undefined);
      }
    }, [filteredValuesData, filteredValuesLoaded, data.podConfig, setData]);
    ```
- When `api.hide: true` removes `tiny_cpu` from `filteredValuesData`, Effect 2 sets `data.podConfig = undefined`. Because `data.podConfig` is in Effect 1's dependencies, Effect 1 immediately sets it back to `"tiny_cpu"`, which re-triggers Effect 2, creating an endless ping-pong re-render loop (`undefined` $\leftrightarrow$ `"tiny_cpu"`).

#### 2. Auto-enabling "Show hidden" in `filterDefaults.ts` (Triggered when only `ui.hide: true`)
- In `WorkspaceFormPodConfigSelection.tsx` and `filterDefaults.ts`:
  ```ts
  export const computeDefaultFilterValues = (options, defaultId) => {
    const defaultOption = options.find((opt) => opt.id === defaultId);
    return {
      showHidden: defaultOption?.hidden ?? false,
      showRedirected: defaultOption?.redirect !== undefined,
    };
  };
  ```
- Because the backend returns the static default ID (`tiny_cpu`), and `tiny_cpu` has `hidden: true` due to `ui.hide`, `computeDefaultFilterValues` sets `showHidden: true` by default.
- Additionally, `selectedPodConfig` is initially set to `tiny_cpu` (which is `hidden: true`), forcing `showHidden = true`.

#### 3. Backend `/listvalues` returns static default
- In `workspaces/backend/internal/models/workspacekinds/podtemplate/options/funcs.go`, `/listvalues` statically returns `wsk.Spec.PodTemplate.Options.PodConfig.Spawner.Default` without validating if that default option was hidden or removed by filter rules in the requested context.

### Relevant Logs

```shell

```

thisis-Shitanshu:
Fixes #1353

# What changed

PodConfig defaults are now evaluated against the selected image context before being automatically selected.

This prevents a configured default from remaining selected when dynamic filter rules make it hidden or remove it from the contextual `listValues` response.

The form now:

- resolves the effective PodConfig default using both unfiltered and image-contextual values
- preserves statically hidden defaults and their existing `Show hidden` behavior
- clears an auto-selected default when it becomes invalid for the selected image
- preserves explicit user selections separately from auto-selected defaults
- waits for contextual PodConfig values before rendering the PodConfig selection step

# Root cause

The form previously applied the PodConfig default from the unfiltered `listValues` response before image-context filtering was available.

With dynamic filter rules, the configured default could later become hidden or be removed while still remaining selected.

With `ui.hide`, this caused hidden PodConfigs to be revealed automatically.

With `ui.hide + api.hide`, the form could repeatedly apply and clear the default, resulting in flickering and a `Maximum update depth exceeded` error.

# Testing

## Manually reproduced both cases from #1353 before the fix

- `ui.hide`: hidden CPU PodConfigs were automatically revealed
1345_5_show_hidden

- `ui.hide + api.hide`: reproduced flickering and `Maximum update depth exceeded`

## Verified after the fix

- dynamically hidden defaults are no longer auto-selected
- `Show hidden` remains off for image-context-hidden defaults

- API-hidden defaults no longer cause the select/clear loop
- no flickering or maximum update-depth error
Screenshot from 2026-08-29 22-31-07
Screenshot from 2026-08-29 22-33-16

- compatible GPU PodConfig remains selectable
- switching from CUDA back to the default SciPy image restores the valid default
- statically hidden and redirected default behavior remains unchanged

## Automated validation

- Jest: 457 tests passing
- targeted Cypress suites: 87 tests passing
- added regression coverage for an image-context-hidden PodConfig default
- TypeScript checks passing
- ESLint passing
- Prettier passing
- `git diff --check` clean

# Related implementation history

While tracing this issue, I found a few earlier changes that are closely related to the behavior fixed here:

- #948 / #957 introduced the hidden and redirected option UX for workspace image and PodConfig selection.
- #691 / #972 introduced the backend `listValues` API used to return image and PodConfig options for compatibility filtering.
- #861 / #1098 integrated `listValues` into `WorkspaceForm` using separate unfiltered (`allValuesData`) and image-contextual (`filteredValuesData`) responses. It also documented a previous default/deselect loop, which is closely related to the failure mode in #1353.
- #1144 / #1254 extended the workspace form's hidden/redirect handling and PodConfig revalidation when image context changes.

### DIFF
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/utils/filterDefaults.ts
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/utils/filterDefaults.ts
@@ -23,3 +23,38 @@ export const computeDefaultFilterValues = <T extends OptionWithHiddenAndRedirect
     showRedirected: defaultOption.redirect !== undefined,
   };
 };
+
+type OptionWithHidden = {
+  id: string;
+  hidden: boolean;
+};
+
+/**
+ * Resolves whether the configured default remains valid in the current context.
+ *
+ * Preserves defaults that are already hidden in the unfiltered configuration,
+ * but rejects defaults that become hidden or are removed by contextual filtering.
+ */
+export const resolveContextualDefaultId = <T extends OptionWithHidden>(
+  allOptions: T[],
+  contextualOptions: T[],
+  defaultId?: string,
+): string | undefined => {
+  if (!defaultId) {
+    return undefined;
+  }
+
+  const configuredDefault = allOptions.find((option) => option.id === defaultId);
+  const contextualDefault = contextualOptions.find((option) => option.id === defaultId);
+
+  if (!configuredDefault || !contextualDefault) {
+    return undefined;
+  }
+
+  // It was visible in the unfiltered configuration, but became hidden in the current context.
+  if (!configuredDefault.hidden && contextualDefault.hidden) {
+    return undefined;
+  }
+
+  return defaultId;
+};
--- a/workspaces/frontend/src/__tests__/cypress/cypress/tests/mocked/workspaces/createWorkspace.cy.ts
+++ b/workspaces/frontend/src/__tests__/cypress/cypress/tests/mocked/workspaces/createWorkspace.cy.ts
@@ -1630,6 +1630,75 @@ describe('Create workspace', () => {
       createWorkspace.findPodConfigCard(bigGpuPod.id).should('be.visible');
     });
 
+    it('should not select a pod config default hidden by image context', () => {
+      const hiddenSmallCpuPod: OptionsPodConfigValue = {
+        ...smallCpuPod,
+        hidden: true,
+      };
+
+      const kind = buildMockWorkspaceKind({
+        podTemplate: {
+          ...mockWorkspaceKind.podTemplate,
+          options: {
+            imageConfig: {
+              default: cpuImage.id,
+              values: [cpuImage, cudaImage],
+            },
+            podConfig: {
+              default: smallCpuPod.id,
+              values: [smallCpuPod, bigGpuPod],
+            },
+          },
+        },
+      });
+
+      cy.interceptApi(
+        'GET /api/:apiVersion/workspacekinds',
+        { path: { apiVersion: NOTEBOOKS_API_VERSION } },
+        mockModArchResponse([kind]),
+      ).as('getWorkspaceKindsListValues');
+
+      cy.intercept(
+        'POST',
+        `**/workspacekinds/${kind.name}/podtemplate/options/listvalues`,
+        (req) => {
+          const imageId = req.body?.data?.context?.imageConfig?.id;
+          const isCudaContext = imageId === cudaImage.id;
+
+          req.reply({
+            data: {
+              imageConfig: {
+                default: cpuImage.id,
+                values: imageId
+                  ? [cpuImage, cudaImage].filter((image) => image.id === imageId)
+                  : [cpuImage, cudaImage],
+              },
+              podConfig: {
+                default: smallCpuPod.id,
+                values: isCudaContext ? [hiddenSmallCpuPod, bigGpuPod] : [smallCpuPod, bigGpuPod],
+              },
+            },
+          });
+        },
+      );
+
+      createWorkspace.visit();
+      cy.wait('@getWorkspaceKindsListValues');
+
+      createWorkspace.selectKind(kind.name);
+      createWorkspace.clickNext();
+
+      createWorkspace.selectImage(cudaImage.id);
+      createWorkspace.clickNext();
+
+      createWorkspace.assertExtraFilterNotChecked('showHidden');
+      createWorkspace.findPodConfigCard(smallCpuPod.id).should('not.exist');
+      createWorkspace.findPodConfigCard(bigGpuPod.id).should('be.visible');
+
+      // The contextually hidden default must not remain selected.
+      cy.findByTestId('next-button').should('be.disabled');
+    });
+
     it('should show only CPU pod configs when listValues returns no GPU options', () => {
       const kind = buildMockWorkspaceKind({
         podTemplate: {
--- a/workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
+++ b/workspaces/frontend/src/app/pages/Workspaces/Form/WorkspaceForm.tsx
@@ -50,6 +50,7 @@ import { submitFormData } from '~/app/pages/Workspaces/Form/submitHelper';
 import { WorkspaceFormSummaryPanel } from '~/app/pages/Workspaces/Form/WorkspaceFormSummaryPanel';
 import { WorkspaceFormRedirectConfirmModal } from '~/app/pages/Workspaces/Form/WorkspaceFormRedirectConfirmModal';
 import { validateName } from './helpers';
+import { resolveContextualDefaultId } from './utils/filterDefaults';
 
 enum WorkspaceFormSteps {
   KindSelection,
@@ -104,6 +105,24 @@ const WorkspaceForm: React.FC = () => {
       imageId: data.imageConfig,
     });
 
+  const effectivePodConfigDefaultId = useMemo(() => {
+    if (
+      !data.imageConfig ||
+      !allValuesLoaded ||
+      !filteredValuesLoaded ||
+      !allValuesData ||
+      !filteredValuesData
+    ) {
+      return undefined;
+    }
+
+    return resolveContextualDefaultId(
+      allValuesData.podConfig.values ?? [],
+      filteredValuesData.podConfig.values ?? [],
+      allValuesData.podConfig.default,
+    );
+  }, [data.imageConfig, allValuesLoaded, filteredValuesLoaded, allValuesData, filteredValuesData]);
+
   // Store original values for edit mode diff view
   const [originalData, setOriginalData] = useState<WorkspaceFormData | undefined>(undefined);
 
@@ -111,6 +130,7 @@ const WorkspaceForm: React.FC = () => {
   // Refs for filter control
   const imageFilterControlRef = useRef<ImageSelectionFilterHandle>(null);
   const podConfigFilterControlRef = useRef<PodConfigSelectionFilterHandle>(null);
+  const autoSelectedPodConfigRef = useRef<string | undefined>(undefined);
   useEffect(() => {
     if (!initialFormDataLoaded || mode === 'create') {
       return;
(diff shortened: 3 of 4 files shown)


thisis-Shitanshu:
Hi @siyuanfoundation, thanks again for the detailed reproduction in #1353. I reproduced both cases locally and have included a fix here.

When you have time, could you please test this PR against your original setup and let me know if it resolves the flickering and hidden PodConfig behavior? Thanks!

christian-heusel (commented):
/ok-to-test

siyuanfoundation:
I have tested the PR with tilt locally, and can confirm the issue in https://github.com/kubeflow/notebooks/issues/1353 is fixed. Thank you @thisis-Shitanshu Great work!

Can I ask one favor? Can you verify the following is true on your cluster? When I tried to edit the wsk through the UI, it showed no field of `filterRules` in the yaml file while `kubectl` definitely has that field.
image

workspaces/frontend/src/app/pages/Workspaces/Form/utils/filterDefaults.ts
```diff
@@ -23,3 +23,32 @@ export const computeDefaultFilterValues = <T extends OptionWithHiddenAndRedirect
     showRedirected: defaultOption.redirect !== undefined,
   };
 };
+
+type OptionWithHidden = {
+  id: string;
+  hidden: boolean;
+};
+
+export const resolveContextualDefaultId = <T extends OptionWithHidden>(
```
siyuanfoundation, line 38: nit: I am not sure the convention for the frontend is. It would be nice to have a function level documentation.

thisis-Shitanshu: Yep, added. Thanks for the suggestion!

thisis-Shitanshu:
Thanks @siyuanfoundation for testing the PR and confirming the fix!
I’ll look into [this](https://github.com/kubeflow/notebooks/pull/1365#issuecomment-5496385595) and share what I find 😊

christian-heusel:
> Can I ask one favor? Can you verify the following is true on your cluster? When I tried to edit the wsk through the UI, it showed no field of filterRules in the yaml file while kubectl definitely has that field.

I think this is because the endpoint does not yet return a field for it:
```
$ curl -sk -H "Kubeflow-Userid: admin" "https://localhost:8443/workspaces/api/v1/workspacekinds/jupyterlab" | python3 -c "import sys,json; d=json.load(sys.stdin)['data']; print('top-level keys:', list(d.keys())); print('has filterRules:', 'filterRules' in d)"
top-level keys: ['revision', 'spawner', 'podTemplate', 'activityRules']
has filterRules: False
```

See this code:

https://github.com/kubeflow/notebooks/blob/9e93648fd17e889caba8e2515f007ff550d73619/workspaces/backend/internal/models/workspacekinds/types_write.go#L34-L45

thisis-Shitanshu:
@siyuanfoundation I reproduced the same behavior: `filterRules` is present on the Kubernetes resource but absent from the WorkspaceKind GET response/UI YAML. I also verified that saving an unrelated edit through the UI preserves the existing `filterRules`.

Thanks @christian-heusel, spot on! Your explanation about the endpoint model matches what I'm seeing.

I’ll do some further diagnosis and create a separate PR for this rather than opening a new issue, since we now have two independent reproductions and a likely cause identified.

siyuanfoundation:
> @siyuanfoundation I reproduced the same behavior: `filterRules` is present on the Kubernetes resource but absent from the WorkspaceKind GET response/UI YAML. I also verified that saving an unrelated edit through the UI preserves the existing `filterRules`.
> 
> Thanks @christian-heusel, spot on! Your explanation about the endpoint model matches what I'm seeing.
> 
> I’ll do some further diagnosis and create a separate PR for this rather than opening a new issue, since we now have two independent reproductions and a likely cause identified.

Thanks @thisis-Shitanshu and @christian-heusel for confirming!
The editor part is a separate problem and tracked in another issue.

This PR LGTM. But I am not an expert in ts or frontend ^_^

workspaces/frontend/src/app/pages/Workspaces/Form/utils/filterDefaults.ts
```diff
@@ -23,3 +23,38 @@ export const computeDefaultFilterValues = <T extends OptionWithHiddenAndRedirect
...
+};
+
+/**
+ * Resolves whether the configured default remains valid in the current context.
+ *
+ * Preserves defaults that are already hidden in the unfiltered configuration,
+ * but rejects defaults that become hidden or are removed by contextual filtering.
+ */
+export const resolveContextualDefaultId = <T extends OptionWithHidden>(
+  allOptions: T[],
```
thaorell, line 39: `resolveContextualDefaultId(allOptions, contextualOptions, defaultId)` takes three positional arguments, which makes call sites easy to misorder (e.g. swapping `allOptions`/`contextualOptions`) since both are same-shaped arrays. Consider a named object param, e.g.  `resolveContextualDefaultId({ allOptions, contextualOptions, defaultId })`, for clarity and safer refactors.

workspaces/frontend/src/app/pages/Workspaces/Form/utils/filterDefaults.ts
```diff
@@ -23,3 +23,38 @@ export const computeDefaultFilterValues = <T extends OptionWithHiddenAndRedirect
     showRedirected: defaultOption.redirect !== undefined,
   };
 };
+
+type OptionWithHidden = {
```
thaorell, line 27: The new `OptionWithHidden` type (`{ id, hidden }`) largely duplicates the existing `OptionWithHiddenAndRedirect` type just above it. Consider having `resolveContextualDefaultId` accept `OptionWithHiddenAndRedirect` (or extracting a shared base type) instead of introducing a near-identical second type in the same file.

thaorell (commented):
Left a few comments below

workspaces/frontend/src/app/pages/Workspaces/Form/utils/filterDefaults.ts
thisis-Shitanshu: Thanks, good point! I updated `resolveContextualDefaultId` to use a named options object, so `allOptions` and `contextualOptions` are explicit at each call site.

workspaces/frontend/src/app/pages/Workspaces/Form/utils/filterDefaults.ts
thisis-Shitanshu: Agreed. Extracted `OptionWithHidden` as the shared base type and made `OptionWithHiddenAndRedirect` extend it, so the common fields are no longer duplicated.

thisis-Shitanshu:
Thanks, @thaorell, for giving your feedback 🤗 I rebased the change onto the current `notebooks-v2` and re-verified #1353 after the review updates. Both `ui.hide` and `ui.hide + api.hide cases` are still resolved, including the original flickering case and the hidden PodConfig behavior.

Issue #1376, opened 2026-09-03
[TASK] Watch Warning Events for owned Pod/StatefulSet


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

The `WorkspaceReconciler` today does not subscribe to the Kubernetes `Event` stream for its owned Pod and StatefulSet. Warning Events (`FailedScheduling`, `FailedMount`, `FailedCreatePodSandBox`, PVC binding failures, image pull backoffs) are already consumed by `generateWorkspaceState` — see the `lastStsWarningEvent` / `lastPodWarningEvent` blocks in `workspace_controller.go` — but only when reconciliation happens to run for another reason. Between triggers, `status.state` and `status.stateMessage` are stale.

This task wires up a Warning-Event watch on the Workspace controller, backed by a new controller-runtime field indexer keyed on the UIDs of the currently-owned Pod and STS, recorded in `Workspace` status. State generation logic itself is not changed — only the set of triggers that cause it to run.

#### Key Files to Modify

```
| File | Role |
| ---- | ---- |
| `workspaces/controller/api/v1beta1/workspace_types.go` | Add `UID` field to `WorkspacePodStatus`; add new `WorkspaceStatefulSetStatus` type; add `PodTemplateStatefulSet` field to `WorkspaceStatus` |
| `workspaces/controller/api/v1beta1/zz_generated.deepcopy.go` | Regenerated by `make generate` |
| `workspaces/controller/manifests/kustomize/base/crd/kubeflow.org_workspaces.yaml` | Regenerated by `make manifests` |
| `workspaces/controller/internal/helper/index.go` | Register a new multi-value field indexer on `Workspace` in `SetupManagerFieldIndexers`, keyed on the UIDs recorded in `status.podTemplatePod` and `status.podTemplateStatefulSet`; export the index key as a new `Index*` constant next to `IndexEventInvolvedObjectUidField` |
| `workspaces/controller/internal/controller/workspace_controller.go` | Add `generateWorkspaceStatefulSetStatus(sts *appsv1.StatefulSet)` helper mirroring `generateWorkspacePodStatus`; populate `status.PodTemplatePod.UID` and `status.PodTemplateStatefulSet` in `generateWorkspaceStatus`; add `Watches(&corev1.Event{}, ...)` in `SetupWithManager` with a Warning-only predicate and a mapper that uses the new indexer |
[...]

PR #1381, opened 2026-09-04
feat: Watch Warning Events for owned Pod/StatefulSet

siyuanfoundation:
closes: #1376

In addition to what's described in the task, this PR also adds a flag `--watch-warning-events` as the kill switch for the feature. 

Kubernetes Warning Events cannot be filtered by owner or namespace at the informer level, so watching them forces the controller to cache every Warning Event
  cluster-wide across all tenants. In an event storm (e.g., node pressure or image-pull failures spanning thousands of pods), this unpartitioned stream can easily
  blow through the controller's memory (currently default to 128M) limit, crashlooping the single replica and halting all workspace reconciliation; the `--watch-warning-events flag` is the only
  escape hatch that disables that shared informer entirely while still letting workspaces inspect their own events via direct, scoped API queries.


workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -1687,36 +1791,38 @@ func (r *WorkspaceReconciler) generateWorkspaceState(ctx context.Context, log lo
...
-		var lastPodWarningEvent *corev1.Event
-		if len(podEvents.Items) > 0 {
-			for i, event := range podEvents.Items {
-				if event.Type == corev1.EventTypeWarning {
-					//
-					// TODO: ensure this actually works when there are multiple Warning events for this object
-					//
-					if lastPodWarningEvent == nil || lastPodWarningEvent.LastTimestamp.Time.Before(event.LastTimestamp.Time) {
-						lastPodWarningEvent = &podEvents.Items[i]
+		if r.Config != nil && r.Config.WatchWarningEvents {
```
thesuperzapper, line 1794: Even if we are not watching warning events, we might as well use their info if we get to this point in the state generation?

That is, why are we concerned with `WatchWarningEvents` here?

workspaces/controller/internal/helper/index.go
```diff
@@ -58,15 +75,17 @@ func indexByWorkspaceOwner(mgr ctrl.Manager, obj client.Object) error {
 // SetupManagerFieldIndexers sets up field indexes on a controller-runtime manager
 func SetupManagerFieldIndexers(mgr ctrl.Manager, cfg *config.EnvConfig) error {
 
-	// Index Event by `involvedObject.uid`
-	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.Event{}, IndexEventInvolvedObjectUidField, func(rawObj client.Object) []string {
-		event := rawObj.(*corev1.Event)
-		if event.InvolvedObject.UID == "" {
-			return nil
+	// Index Event by `involvedObject.uid` (only when warning event watching is enabled)
+	if cfg != nil && cfg.WatchWarningEvents {
```
thesuperzapper, line 79: Similarly, we can still index if `WatchWarningEvents` is not true, even if we are not triggering reconcile.

This makes me think the config might be better named `RECONCILE_ON_OWNED_WARNING_EVENTS`, or similar (just giving an example).

workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -1571,36 +1673,38 @@ func (r *WorkspaceReconciler) generateWorkspaceState(ctx context.Context, log lo
...
-		var lastStsWarningEvent *corev1.Event
-		if len(statefulSetEvents.Items) > 0 {
-			for i, event := range statefulSetEvents.Items {
-				if event.Type == corev1.EventTypeWarning {
-					//
-					// TODO: ensure this actually works when there are multiple Warning events for this object
-					//
-					if lastStsWarningEvent == nil || lastStsWarningEvent.LastTimestamp.Time.Before(event.LastTimestamp.Time) {
-						lastStsWarningEvent = &statefulSetEvents.Items[i]
+		if r.Config != nil && r.Config.WatchWarningEvents {
```
thesuperzapper, line 1676: Same here.

workspaces/controller/internal/helper/index.go
siyuanfoundation: the `WatchWarningEvents` flag is meant to be a full stopgap, and the memory pressure is a higher risk than reconciliation churn. 

Event watch is cluster wide and not bound by namespaces or any selectors. In the event of event storm, the reconciliation churn is already mostly mitigated by the queue dedupe, but the cache can really blow up the controller. 

If the controller crashes from high memory, its initial startup phase must perform a full List of all active Warning events across the cluster to populate the cache. If the cluster is already in an event surge, the controller will OOM immediately on boot and enter a CrashLoop.

workspaces/controller/internal/controller/workspace_controller.go
siyuanfoundation: changed the state generation to use directly list if `WatchWarningEvents=false` and cache is not enabled.

PR #1382, opened 2026-09-04
chore: go 1.27 upgrade

christian-heusel:
## Summary

Bumps both Go workspaces (`controller` and `backend`) from Go 1.26.5 to **Go 1.27.1**, updates the Docker build stages to `golang:1.27`, and moves golangci-lint to **v2.13.2** so the linter can analyze the new language features. The version bump surfaced a few new linter findings, which are applied in the same PR so `make lint` stays green.

See each commit for a more detailed motivation for each particular change.

## Motivation

Staying current with Go patch releases keeps us on supported toolchains and picks up compiler, runtime, and standard-library improvements.  One improvement Go 1.27 brings is directly relevant here:

- **Faster JSON, for free.** In Go 1.27 the new `encoding/json/v2` implementation backs the standard `encoding/json` API **by default**, since any code compiled with 1.27 gets the benefits of the new implementation without code changes (the new v2 implementation is used in the background for v1 now). This is directly relevant to a controller/BFF that (de)serializes Kubernetes objects constantly. Per the Go team's writeup:

  > The Marshal performance of v2 is roughly at parity with v1. Sometimes it is slightly faster, but other times it is slightly slower. The Unmarshal performance of v2 is significantly faster than v1, with benchmarks demonstrating improvements of up to 10x.
  >
  > — [The Go Blog: "Experimenting with a new JSON API"](https://go.dev/blog/jsonv2-exp#performance-optimizations)

  So simply landing on Go 1.27 speeds up our existing JSON handling (most notably unmarshaling) with no further work needed. See the [Go 1.27 release notes on `encoding/json/v2`](https://go.dev/doc/go1.27#jsonv2) for more details.

Potentially we could also raise a follow-up PR regarding actually using the new jsonv2 implementation, but that is outside of the scope of this PR 😊 

Previously this update was blocked by https://github.com/kubeflow/notebooks/pull/1348 which is now merged 🎉


andyatmiami (approved):
/lgtm
/approve

---

yay new language features! thanks @christian-heusel !

existing GHA checks more or less do a nice job exercising and validating this work... additionally clicked around via our `tilt` setup to perform "common activities" across workspaces and workspacekinds - inspecting logs to ensure I didn't see any strange errors manifest that could be tied to go upgrade issues.

all looks good!

Issue #1294, opened 2026-07-31
[TASK] [Backend] Implement repository-level TTL cache for workspace resource usage


richabanker:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

This task adds a short-lived (~30–60s) in-memory TTL cache to `Repository.GetWorkspaceResourceUsage` in internal/repositories/metrics/repo.go. This protects the Kubernetes API server and aggregated Metrics Server from excessive load caused by frequent UI polling across multiple clients.

### Acceptance Criteria

- [ ] A bounded ~30–60s in-memory TTL cache keyed by (namespace, workspace) is implemented in `Repository.GetWorkspaceResourceUsage`
- [ ] Repeated calls within the TTL window return the cached response without querying the underlying Provider or Kubernetes API server
- [ ] Concurrency-safe read/write access (sync.RWMutex or equivalent cache library) is verified under parallel request load
- [ ] Bounded capacity with standard eviction (LRU/LFU) prevents unbounded memory growth in large clusters

andyatmiami:
/close

this work has been completed and merged

PR #1385, opened 2026-09-06
feat: implement repository-level TTL cache for workspace resource usage metrics

richabanker:
This PR implements a short-lived (30s) in-memory LRU TTL cache for workspace resource utilization in `MetricsRepository.GetWorkspaceResourceUsage`

 closes: [#1294](https://github.com/kubeflow/notebooks/issues/1294)

### DIFF
--- a/workspaces/backend/internal/repositories/metrics/repo.go
+++ b/workspaces/backend/internal/repositories/metrics/repo.go
@@ -18,6 +18,7 @@ package metrics
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
@@ -25,6 +26,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/cache"
 	metricsv1beta1 "k8s.io/metrics/pkg/apis/metrics/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -37,14 +39,29 @@ import (
 const (
 	// TTL for API availability checks.
 	apiAvailabilityTTL = 60 * time.Second
+
+	// TTL for cached workspace resource usage responses.
+	resourceUsageCacheTTL = 30 * time.Second
+
+	// Maximum entries in the resource usage LRU cache.
+	resourceUsageCacheMaxCapacity = 1000
 )
 
 // MetricsRepository exposes point-in-time workspace resource utilization, read from the
 // Kubernetes Metrics Server, to the API layer.
+//
+// It maintains two levels of in-memory caching to minimize cluster overhead:
+//  1. API Availability (apiAvailable): A memoized probe (60s TTL) that caches whether the
+//     Kubernetes Metrics API (metrics.k8s.io) is served in the cluster, avoiding repetitive
+//     RESTMapper discovery calls.
+//  2. Resource Usage Cache (usageCache): A TTL LRU cache (30s TTL, max 1000 entries) keyed by
+//     "<namespace>/<workspace>/<podUID>" that stores computed resource usage metrics to avoid
+//     hammering the Metrics Server with repeated PodMetrics queries during high-frequency polling.
 type MetricsRepository struct {
 	cfg          *config.EnvConfig
 	client       client.Client
 	apiAvailable func() bool
+	usageCache   *cache.LRUExpireCache
 }
 
 // NewMetricsRepository creates a MetricsRepository for accessing workspace metrics.
@@ -53,6 +70,7 @@ func NewMetricsRepository(cfg *config.EnvConfig, c client.Client) *MetricsReposi
 		cfg:          cfg,
 		client:       c,
 		apiAvailable: memoize(apiAvailabilityTTL, func() bool { return metricsAPIServed(c) }),
+		usageCache:   cache.NewLRUExpireCache(resourceUsageCacheMaxCapacity),
 	}
 }
 
@@ -83,6 +101,16 @@ func (r *MetricsRepository) GetWorkspaceResourceUsage(ctx context.Context, ns, w
 	// strict deployment guarantees, there will only ever be a maximum of one pod running
 	// at any given time. Therefore, we can safely just grab the first item in the list.
 	pod := &podList.Items[0]
+
+	cacheKey := fmt.Sprintf("%s/%s/%s", ns, workspace, pod.UID)
+	if r.usageCache != nil {
+		if val, ok := r.usageCache.Get(cacheKey); ok {
+			if cachedUsage, valid := val.(*models.WorkspaceResourceUsage); valid {
+				return cachedUsage, nil
+			}
+		}
+	}
+
 	if !r.apiAvailable() {
 		return models.NewWorkspaceResourceUsage(pod, nil), nil
 	}
@@ -92,14 +120,20 @@ func (r *MetricsRepository) GetWorkspaceResourceUsage(ctx context.Context, ns, w
 		if ctx.Err() != nil {
 			return nil, ctx.Err()
 		}
+		// Pod metrics not ready or unavailable; return resource limits/requests without caching nil metrics.
 		return models.NewWorkspaceResourceUsage(pod, nil), nil
 	}
 
-	usageByContainer := models.UsageForPod(podMetrics)
-	return models.NewWorkspaceResourceUsage(
-		pod,
-		usageByContainer,
-	), nil
+	if len(podMetrics.Containers) == 0 {
+		// Metrics scrape has not populated container metrics yet; return without caching so subsequent polls can retry.
+		return models.NewWorkspaceResourceUsage(pod, nil), nil
+	}
+
+	usage := models.NewWorkspaceResourceUsage(pod, models.UsageForPod(podMetrics))
+	if r.usageCache != nil {
+		r.usageCache.Add(cacheKey, usage, resourceUsageCacheTTL)
+	}
+	return usage, nil
 }
 
 // memoize caches the result of the probe for ttl.
--- a/workspaces/backend/internal/repositories/metrics/repo_test.go
+++ b/workspaces/backend/internal/repositories/metrics/repo_test.go
@@ -19,6 +19,7 @@ package metrics
 import (
 	"context"
 	"errors"
+	"fmt"
 	"testing"
 	"time"
 
@@ -36,6 +37,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/cache"
 	metricsv1beta1 "k8s.io/metrics/pkg/apis/metrics/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
@@ -77,7 +80,7 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
 				&corev1.PodList{Items: []corev1.Pod{*pod}},
 			).Build()
 
-		repo := newTestMetricsRepository(client, true)
+		repo := newTestMetricsRepository(client, true, resourceUsageCacheMaxCapacity)
 		got, err := repo.GetWorkspaceResourceUsage(ctx, "default", "test-workspace")
 		expected := models.ContainerResourceUsage{
 			MetricsFromMetricsServer: &models.MetricsFromMetricsServer{
@@ -112,7 +115,7 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
 				&corev1.PodList{Items: []corev1.Pod{*pod}},
 			).Build()
 
-		repo := newTestMetricsRepository(client, true)
+		repo := newTestMetricsRepository(client, true, resourceUsageCacheMaxCapacity)
 		got, err := repo.GetWorkspaceResourceUsage(ctx, "default", "test-workspace")
 		expected := models.ContainerResourceUsage{
 			MetricsFromMetricsServer: nil,
@@ -136,7 +139,7 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
 			WithLists(&corev1.PodList{Items: []corev1.Pod{}}).
 			Build()
 
-		repo := newTestMetricsRepository(client, true)
+		repo := newTestMetricsRepository(client, true, resourceUsageCacheMaxCapacity)
 		got, err := repo.GetWorkspaceResourceUsage(ctx, "default", "no-such-workspace")
 
 		Expect(err).To(MatchError(repoCommon.ErrWorkspaceNotFound))
@@ -151,7 +154,7 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
 				&corev1.PodList{Items: []corev1.Pod{}},
 			).Build()
 
-		repo := newTestMetricsRepository(client, true)
+		repo := newTestMetricsRepository(client, true, resourceUsageCacheMaxCapacity)
 		got, err := repo.GetWorkspaceResourceUsage(ctx, "default", "test-workspace")
 
 		Expect(err).To(MatchError(repoCommon.ErrWorkspacePodNotRunning))
@@ -169,7 +172,7 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
 			WithLists(&corev1.PodList{Items: []corev1.Pod{*pod}}).
 			Build()
 
-		repo := newTestMetricsRepository(client, false)
+		repo := newTestMetricsRepository(client, false, resourceUsageCacheMaxCapacity)
 		got, err := repo.GetWorkspaceResourceUsage(ctx, "default", "test-workspace")
 		expected := models.ContainerResourceUsage{
 			MetricsFromMetricsServer: nil,
@@ -205,7 +208,7 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
 			}).
 			Build()
 
-		repo := newTestMetricsRepository(client, true)
+		repo := newTestMetricsRepository(client, true, resourceUsageCacheMaxCapacity)
 		got, err := repo.GetWorkspaceResourceUsage(ctx, "default", "test-workspace")
 		expected := models.ContainerResourceUsage{
 			MetricsFromMetricsServer: nil,
@@ -241,7 +244,7 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
 			}).
 			Build()
 
-		repo := newTestMetricsRepository(client, true)
+		repo := newTestMetricsRepository(client, true, resourceUsageCacheMaxCapacity)
 		got, err := repo.GetWorkspaceResourceUsage(ctx, "default", "test-workspace")
 		expected := models.ContainerResourceUsage{
 			MetricsFromMetricsServer: nil,
@@ -277,7 +280,7 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
 			}).
 			Build()
 
-		repo := newTestMetricsRepository(client, true)
+		repo := newTestMetricsRepository(client, true, resourceUsageCacheMaxCapacity)
 		got, err := repo.GetWorkspaceResourceUsage(ctx, "default", "test-workspace")
 		expected := models.ContainerResourceUsage{
 			MetricsFromMetricsServer: nil,
@@ -292,37 +295,6 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
 		Expect(got).NotTo(BeNil())
 		Expect(got.Containers).To(HaveLen(1))
 		Expect(got.Containers["container-4"]).To(BeComparableTo(expected))
-	})
-
-	It("omits metricsFromMetricsServer when availability could not be determined", func() {
-		pod := workspacePod("pod-5", "container-5", corev1.ResourceList{
-			corev1.ResourceCPU: resource.MustParse("100m"),
-		})
-
-		client := fake.NewClientBuilder().
-			WithScheme(scheme).
-			WithObjects(testWorkspaceCR()).
-			WithLists(&corev1.PodList{Items: []corev1.Pod{*pod}}).
-			Build()
-
-		repo := &MetricsRepository{
-			client:       client,
-			apiAvailable: func() bool { return false },
-		}
-		got, err := repo.GetWorkspaceResourceUsage(ctx, "default", "test-workspace")
-		expected := models.ContainerResourceUsage{
-			MetricsFromMetricsServer: nil,
-			Resources: corev1.ResourceRequirements{
-				Requests: corev1.ResourceList{
-					corev1.ResourceCPU: resource.MustParse("100m"),
-				},
-			},
-		}
-
-		Expect(err).NotTo(HaveOccurred())
-		Expect(got).NotTo(BeNil())
-		Expect(got.Containers).To(HaveLen(1))
-		Expect(got.Containers["container-5"]).To(BeComparableTo(expected))
 	})
 
 	It("propagates context cancellation error when context is canceled while getting metrics", func() {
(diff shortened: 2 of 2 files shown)


christian-heusel (commented):
/ok-to-test

workspaces/backend/internal/repositories/metrics/repo.go
```diff
@@ -79,27 +88,35 @@ func (r *MetricsRepository) GetWorkspaceResourceUsage(ctx context.Context, ns, w
...
+			usage = models.NewWorkspaceResourceUsage(pod, usageByContainer)
 		}
-		return models.NewWorkspaceResourceUsage(pod, nil), nil
 	}
 
-	usageByContainer := models.UsageForPod(podMetrics)
-	return models.NewWorkspaceResourceUsage(
-		pod,
-		usageByContainer,
-	), nil
```
christian-heusel: I think we could potentially unify the two branches the `if`-statement to make things more redable:
```suggestion
	usage := models.NewWorkspaceResourceUsage(pod, nil)
	if r.apiAvailable() {
		podMetrics := &metricsv1beta1.PodMetrics{}
		switch err := r.client.Get(ctx, client.ObjectKey{Namespace: ns, Name: pod.Name}, podMetrics); {
		case err == nil:
			usage = models.NewWorkspaceResourceUsage(pod, models.UsageForPod(podMetrics))
		case ctx.Err() != nil:
			return nil, ctx.Err()
		}
	}
```

I hope I missed no side-effects that would block this refactor, feel free to push back as neccessary! 😅

workspaces/backend/internal/repositories/metrics/repo_test.go
```diff
@@ -352,6 +353,147 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
...
+		It("returns cached response on repeated calls without querying client again", func() {
+			pod := workspacePod("pod-cached", "container-cached", corev1.ResourceList{
+				corev1.ResourceCPU: resource.MustParse("100m"),
+			})
+			metrics := workspacePodMetrics("pod-cached", "container-cached", corev1.ResourceList{
+				corev1.ResourceCPU: resource.MustParse("75m"),
+			})
+
+			podMetricsGetCount := 0
+			client := fake.NewClientBuilder().
```
christian-heusel, line 338: The defintion of `client` here shadows the client package above leading to the following CI error:

```
Error: internal/repositories/metrics/repo_test.go:405:23: client.ObjectKey is not a type (typecheck)
```
https://github.com/kubeflow/notebooks/actions/runs/34016297130/job/102059905793?pr=1385

workspaces/backend/internal/repositories/metrics/repo_test.go
christian-heusel: This is missing the `usageCache` as part of the construction, resulting in a panic:
```suggestion
		repo := newTestMetricsRepository(client, false)
```

workspaces/backend/internal/repositories/metrics/repo_test.go
```diff
@@ -439,10 +581,15 @@ func testWorkspaceCR() *kubefloworgv1beta1.Workspace {
 	}
 }
 
-func newTestMetricsRepository(c client.Client, apiAvailable bool) *MetricsRepository {
+func newTestMetricsRepository(c client.Client, apiAvailable bool, cacheCapacity ...int) *MetricsRepository {
+	capacity := resourceUsageCacheMaxCapacity
+	if len(cacheCapacity) > 0 && cacheCapacity[0] > 0 {
+		capacity = cacheCapacity[0]
+	}
```
christian-heusel: Why does this function accept a variadic `cacheCapacity ...int` if we ever only read the first parameter? To me this seems like an odd way to specify an optional parameter, is this a usual thing to do in go?

Intuitively, I would rather expect another overload or just explicitly setting the capacity in the call sites 🤔

workspaces/backend/internal/repositories/metrics/repo.go
christian-heusel, line 102: I think we could potentially unify the two branches the `if`-statement to make things more redable:
```suggestion
	usage := models.NewWorkspaceResourceUsage(pod, nil)
	if r.apiAvailable() {
		podMetrics := &metricsv1beta1.PodMetrics{}
		switch err := r.client.Get(ctx, client.ObjectKey{Namespace: ns, Name: pod.Name}, podMetrics); {
		case err == nil:
			usage = models.NewWorkspaceResourceUsage(pod, models.UsageForPod(podMetrics))
		case ctx.Err() != nil:
			return nil, ctx.Err()
		}
	}
```

I hope I missed no side-effects that would block this refactor, feel free to push back as neccessary! 😅

workspaces/backend/internal/repositories/metrics/repo.go
```diff
@@ -79,27 +88,35 @@ func (r *MetricsRepository) GetWorkspaceResourceUsage(ctx context.Context, ns, w
...
+	r.usageCache.Add(cacheKey, usage, resourceUsageCacheTTL)
```
christian-heusel: Should we also cache the result if the metricsServer backend is not available? 

🤖 -based review:

> When the metrics API is unavailable or the `PodMetrics` GET fails transiently (503 during
metrics-server restart, RBAC-propagation 403, momentary NotFound), a `metrics=nil` result is
cached and served for the full 30s — this is **negative caching newly introduced by this PR**.
If the metrics server recovers 1s later, callers still get metrics-less responses for ~29s,
because the success-path recompute is short-circuited. The existing test at `repo_test.go:226`
("omits metricsFromMetricsServer when PodMetrics returns ServiceUnavailable") confirms this
degraded path is real.
>
> **Fix:** only cache the fully-populated result, or cache degraded results with a much shorter
TTL (e.g. gate the `Add` on whether metrics are present).

workspaces/backend/internal/repositories/metrics/repo_test.go
```diff
@@ -352,6 +353,147 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
...
+				Build()
+
+			repo := newTestMetricsRepository(client, true)
+			cacheKey := client.ObjectKey{Namespace: "default", Name: "test-workspace"}.String()
+			repo.usageCache.Add(cacheKey, &models.WorkspaceResourceUsage{}, time.Minute)
+
+			got, err := repo.GetWorkspaceResourceUsage(ctx, "default", "test-workspace")
+			Expect(err).To(MatchError(repoCommon.ErrWorkspacePodNotRunning))
+			Expect(got).To(BeNil())
+		})
```
christian-heusel: 🤖 -based review result:

> The empty-pod-list check (`repo.go:87-89`) returns `ErrWorkspacePodNotRunning` **before** the cache
is consulted (`repo.go:91-96`), so the seeded entry is never read — the test passes identically
with or without the `Add`, giving false confidence a cache-bypass path is exercised.
> 
> **Fix:** drop the misleading `Add` and rename to reflect that the check precedes the cache, or
assert the invariant that matters (e.g. the seeded entry is still present and untouched after the
call).

christian-heusel (changes requested):
Hey @richabanker, thanks a lot for this PR 🤗 

Just for my understanding, we now have 2-layer cache right?
Pod -> MetricsServer -> repositoryCache

Maybe we should explain this hierarchy briefly with a comment on `MetricsRepository`? 😊

christian-heusel:
Also it seems like the multi-line suggestions that I posted above are broken, dunno what happened there 🤔
Just beware that they can't be applied directly 😅

workspaces/backend/internal/repositories/metrics/repo_test.go
richabanker: Fixed

workspaces/backend/internal/repositories/metrics/repo_test.go
richabanker: That was just me being lazy... updated all call-sites to now specify the cacheCapacity param

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: hmm, I dont like the use of switch for case err == nil, particularly because a more go-idiomatic way is to test for `err != nil`

Moreover we do already check `if !r.apiAvailable()` and return early on L#114. I think I prefer the early returns over the switch statements.. WDYT ?

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: the bot based review is correct, we actually dont want to cache the result when metricsServer is unavailable for transient reasons.. so now updated the code to just return the pod resource limits/requests without caching nil metrics. Only when we do get a successful result from metrics-server do we update the cache.

workspaces/backend/internal/repositories/metrics/repo_test.go
richabanker: uh right, that test is meaningless. Should have not pushed that one, thanks for the catch! Removed now.

richabanker:
> Maybe we should explain this hierarchy briefly with a comment on `MetricsRepository`? 😊

Good idea, added!

workspaces/backend/internal/repositories/metrics/repo_test.go
```diff
@@ -294,37 +297,6 @@ var _ = Describe("MetricsRepository.GetWorkspaceResourceUsage", func() {
 		Expect(got.Containers["container-4"]).To(BeComparableTo(expected))
 	})
 
-	It("omits metricsFromMetricsServer when availability could not be determined", func() {
```
richabanker, line 297: Removed since its redundant to the existing "omits metricsFromMetricsServer when PodMetrics is missing" test

workspaces/backend/internal/repositories/metrics/repo.go
andyatmiami, line 119: **Thundering herd on cache miss (no request coalescing) — Metrics Server DoS risk**

At TTL expiry (or cold start) for any hot workspace, every concurrent caller sees a miss and issues its own `Get(PodMetrics)`. In an enterprise deployment where a single workspace's usage endpoint is polled by:

- multiple browser tabs / users viewing the same workspace,
- (future?) the workspace list view fanning out one call per visible row,
- programmatic API consumers (scripts, CI, custom tooling),

…the effective miss-time QPS to the Metrics Server is `N_pollers × 1` per 30-second window — not `1`. That is exactly the "hammering" the cache claim to prevent (per the struct doc comment).

Multiply this across every workspace whose TTL expires within the same wall-clock second and the Metrics Server sees synchronized bursts (all pollers on the same 30s cadence align). This is textbook cache stampede.

_PROPOSED MITIGATION_

Wrap the miss path in `golang.org/x/sync/singleflight` keyed by `cacheKey`. Every duplicate concurrent miss for the same key then piggy-backs on the one in-flight fetch. This reduces worst-case Metrics Server load from `O(concurrent_pollers × workspaces)` to `O(workspaces)`.

`singleflight` is part of the official `golang.org/x/sync` sub-repo (maintained by the Go team, used inside `k8s.io/client-go`, `etcd`, gRPC-Go, Prometheus). It is the canonical Go pattern for this problem — not a fringe pick. The one genuine gotcha is context cancellation: `Do` runs the fetch on the **leader**'s goroutine, so if the leader's request is cancelled, every coalesced waiter receives `context.Canceled` even if their own request contexts are still alive. Mitigate this by using **`DoChan`** (so waiters can `select` on their own `ctx.Done()`) combined with a **detached context inside the fetch closure** (so a cancelled leader does not abort the shared K8s call). This is the same pattern client-go's discovery client uses.

andyatmiami, line 153: **`memoize()` serializes every metrics request through a single mutex — probe held under lock**

Two compounding problems:

1. **Every miss path calls `r.apiAvailable()`**, which acquires `mu`. Even on the fast path (cache is warm), every concurrent caller serializes on this mutex. Under high fan-out that turns `apiAvailable` into a global lock along the hot path.
2. **The probe (`RESTMapper().RESTMapping(...)`) runs while holding `mu`.** RESTMapper can trigger discovery — an out-of-process API call — which may block for hundreds of ms during API-server hiccups. Every metrics request in flight at that moment blocks until the probe returns. Combined with (1), one slow probe stalls the entire endpoint for all workspaces simultaneously.

_PROPOSED MITIGATION_

Switch to an `atomic.Pointer[struct{val bool; checkedAt time.Time}]` or `atomic.Int64` (packed `checkedAtUnix|val`) for the read fast-path, and use `sync.Once` / `singleflight` to guarantee only one goroutine runs the probe at TTL expiry while others read the previous value. Never hold a lock across the probe call.

The "correct" (?) pattern here is **stale-while-revalidate**: the fast path is a single lock-free atomic load, only the cold-start caller ever blocks, and TTL expiry triggers a background refresh while concurrent callers continue to receive the last known value. This is the same shape as client-go's discovery cache and etcd's leaseship checks — it is not a novel invention.

andyatmiami, line 162: **Transient probe failure is cached as "unavailable" for 60 s — silent metrics blackout**

`memoize` treats success and failure identically — a single failed `RESTMapping` call (API-server 5xx during rolling upgrade, network blip, CRD-install window) pins `apiAvailable() == false` for a full 60 seconds. During that window every workspace usage response omits live metrics (returns `NewWorkspaceResourceUsage(pod, nil)`) with no error and no log. At enterprise scale this manifests as a fleet-wide "usage numbers disappeared" dashboard incident with no signal in the request logs.

**This finding is orthogonal to [prior comment](https://github.com/kubeflow/notebooks/pull/1385/changes#r3992900368)** The other suggestion makes concurrent access lock-free and stampede-safe, but the atomic snapshot still caches whatever the probe returned for the same fixed TTL regardless of *why* it returned it. The blackout persists after #2 lands, exactly as bad as before.

_PROPOSED FIX_

Widen the probe signature from `func() bool` to `func() (bool, error)` and treat `error != nil` as "publish with a short negative TTL." Do NOT branch on the boolean alone.

`metricsAPIServed`'s current `return err == nil` collapses two semantically distinct outcomes into `false`:

- **Confirmed absent.** RESTMapper responded with `meta.NoKindMatchError` — the metrics API is genuinely not installed. This is a *legitimate steady state* (bare Kubeflow installs frequently run without metrics-server) and deserves the full positive TTL.
- **Transient probe failure.** The RESTMapper call errored for a reason unrelated to metrics-API presence. This is what we want to fix.

A "short TTL on any `false`" fix would short-cache the legitimate-absence case, re-running RESTMapper every few seconds forever on clusters that never had metrics-server — the opposite of what a TTL is for. The `(bool, error)` variant lets the probe expose the discrimination the underlying Kubernetes API already provides via `meta.IsNoMatchError`.

workspaces/backend/internal/repositories/metrics/repo.go
```diff
@@ -92,14 +120,20 @@ func (r *MetricsRepository) GetWorkspaceResourceUsage(ctx context.Context, ns, w
...
 
-	usageByContainer := models.UsageForPod(podMetrics)
-	return models.NewWorkspaceResourceUsage(
-		pod,
-		usageByContainer,
-	), nil
+	if len(podMetrics.Containers) == 0 {
+		// Metrics scrape has not populated container metrics yet; return without caching so subsequent polls can retry.
+		return models.NewWorkspaceResourceUsage(pod, nil), nil
+	}
```
andyatmiami, line 130: **Negative results (pod-metrics-not-ready) are intentionally not cached — sustained hammering during workspace warm-up**

The stated rationale ("allow subsequent polls to pick up metrics") is defensible for correctness, but at enterprise scale the failure mode is:

- A tenant kicks off a batch of workspaces (e.g., a class of 200 students starts JupyterLab).
- For 30–120 s each pod is running but `metricsv1beta1.PodMetrics` isn't yet populated (Metrics Server scrape lag is typically 15–60 s).
- Every 5 s poll from every dashboard pays a full `Get(PodMetrics)` per pod, on every request, for the whole warm-up window.

That is precisely the request storm the cache is supposed to shed.

_PROPOSED MITIGATION_

[Intertwined with prior suggestion](https://github.com/kubeflow/notebooks/pull/1385/changes#r3992883762)

Both non-happy-path branches gain an `Add` call with a short TTL, using the same `cacheKey` as the positive path so the LRU accounts for negative and positive entries under one budget. When metrics arrive, the next miss simply overwrites the negative entry with a positive one (the LRU's `Add` semantics).

workspaces/backend/internal/repositories/metrics/repo.go
```diff
@@ -37,14 +39,29 @@ import (
 const (
 	// TTL for API availability checks.
 	apiAvailabilityTTL = 60 * time.Second
+
+	// TTL for cached workspace resource usage responses.
+	resourceUsageCacheTTL = 30 * time.Second
+
+	// Maximum entries in the resource usage LRU cache.
+	resourceUsageCacheMaxCapacity = 1000
```
andyatmiami, line 47: **Is `resourceUsageCacheMaxCapacity = 1000` the right ceiling for enterprise deployments?**

 Rough per-entry footprint:

- **Cache key:** `<ns>/<workspace>/<uid>` ≈ 100 bytes (namespace + workspace name + 36-byte UUID + separators).
- **Value:** `*WorkspaceResourceUsage`. Contains one `map[string]ContainerResourceUsage` with typically 1–3 entries. Each `ContainerResourceUsage` embeds `corev1.ResourceRequirements` (two `ResourceList` maps of `resource.Quantity` — each `Quantity` is ~40 B with an inner `big.Int`) plus a small `MetricsFromMetricsServer` pointer target.
- **LRU book-keeping:** doubly-linked-list node + map entry ≈ 50 bytes.

Total per entry: **~1 KB** for a typical 2-container workspace pod, ~1.5 KB worst-case for pods with more containers or richer resource specs.

Extrapolated:

| Capacity | Steady-state heap (typical) | Steady-state heap (worst) |
|----------|-----------------------------|---------------------------|
| 1 000 | ~1 MB | ~1.5 MB |
| 10 000 | ~10 MB | ~15 MB |
| 100 000 | ~100 MB | ~150 MB |

I feel like 10000 might be better suited as a default 🤷

workspaces/backend/internal/repositories/metrics/repo.go
andyatmiami, line 1: ℹ️ This comment spans the K8s reads (`Workspace.Get` on 82-88, `Pod.List` on 90-94) and the cache check on 106-112

**`Workspace.Get` + `Pod.List` execute on every request — cache-hit path still touches K8s**

Every request — hit or miss — still runs `Workspace.Get` + `Pod.List` before the cache lookup. These reads come from the controller-runtime informer cache and never leave the process, but each one still:

- deep-copies a `Workspace` CR out of the informer store (~1 KB), and
- deep-copies a `PodList` plus at least one `Pod` (~10 KB — Pod carries embedded spec, status, containers, volumes, etc.).

At sustained enterprise polling load the allocation cost dominates the cache benefit:

```
| Requests/sec | Workspace Get bytes/sec | Pod List bytes/sec | Total alloc/sec |
|---|---|---|---|
| 500  | 500 KB | 5 MB | 5.5 MB |
| 5 000 | 5 MB | 50 MB | 55 MB |
| 50 000 | 50 MB | 500 MB | 550 MB |
```

The bottom row (5 000 workspaces × 10 pollers × 1 poll/sec) is exactly the regime the cache exists to help. Instead the fast path is producing ~550 MB/sec of short-lived allocations that land in young-generation heap, get scanned during GC mark, and get collected on the next cycle. Real CPU, real STW impact on a request path with millisecond latency budgets. The cached value itself (`*WorkspaceResourceUsage`, ~1 KB) is dwarfed by the per-request informer copies that happen before the cache is even consulted.

_PROPOSED MITIGATION_

Change the cache key from `ns/workspace/podUID` to `ns/workspace`, and move all three reads (`Workspace.Get`, `Pod.List`, `PodMetrics.Get`) inside `fetchAndCacheUsage`. The cached value gains a `podUID` field retained for observability but no longer part of the lookup identity. The two paths become:

```
Cache HIT: return cached usage. Zero K8s reads. Zero informer copies.
Cache MISS: singleflight → Workspace.Get + Pod.List + PodMetrics.Get → cache.
```

The hit path collapses to one `LRUExpireCache.Get` + one type assertion. That is what "cache saves work" is supposed to look like.

**Trade-off — pod-recreation staleness for up to `resourceUsageCacheTTL`.**

Losing pod-UID from the cache key means losing the automatic invalidation that pod-recreation currently provides. Two scenarios to be explicit about:

- **Pod OOMKilled / evicted / restarted by StatefulSet.** Cache serves the previous pod's usage numbers for up to 30 s. Same magnitude as any other TTL-bounded staleness. Users see "usage transitioned from X to Y over 30 s" rather than a hard reset — not user-visible as a bug.
- **User edits workspace resource requests/limits → pod recreated with new spec.** Cache serves the *old* `Resources` block (requests/limits) for up to 30 s. **This is the one legitimate user-visible regression.** A user clicking "edit workspace, save" and immediately opening the resource page might see old requests until TTL expiry. Recoverable via refresh after 30 s.

The net trade-off: **~550 MB/sec of hot-path allocation eliminated in exchange for ≤ 30 s of stale requests/limits after user edits.** For an enterprise deployment where edit-then-immediately-check is a rare interaction and QPS is high, the trade is worth taking. For a tiny deployment the allocation cost is invisible, so nothing is lost either — the change is opt-out-safe at the small end.

andyatmiami, line 119: **No per-call deadline in `GetWorkspaceResourceUsage` (pre-existing — surfaced because the cache change amplifies its blast radius)**

The repository accepts the caller's `ctx` unchanged and passes it to three K8s reads:

```go
r.client.Get(ctx, ..., ws)             // informer-cached — cheap
r.client.List(ctx, podList, ...)       // informer-cached — cheap
r.client.Get(ctx, ..., podMetrics)     // Metrics Server round-trip — the risk
```

The first two hit controller-runtime's local informer cache and are essentially free. The third — the `PodMetrics` Get — is a live HTTP call to the Metrics Server via the aggregated API. If Metrics Server is wedged (deploy rollout, network partition, its own upstream stall), controller-runtime's default REST client timeout is on the order of tens of seconds and the caller's goroutine sits pinned until the caller cancels or the transport gives up.

_PROPOSED FIX_

Wrap only the Metrics Server call — not the informer reads, and not any control-flow that runs before the miss path. - with its own `context.WithTimeout`
- It has to be strictly less than `resourceUsageCacheTTL` (30 s) so a slow upstream cannot cause the next poll to arrive before the previous one has returned.
    - Proposed: 5s?

andyatmiami (changes requested):
@richabanker 

Sorry for this review on a few accounts:
- it took longer than I'd like to get back to you
- there are walls (plural) of text
- I am re-litigating past decisions on prior PR

I took a (potentially absurdly) close this at this PR really honing in on what would make an **ideal, robust, enterprise cache** ... and admittedly I could be over-engineering/critiquing the shit out of this 😇 

While building this caching solution to scale **FOR METRICS SERVER** could reasonably be debatable - I also kinda like the idea of having this **patttern** in our codebase to act as a reference for other use cases that undeniably warrant this level of optimization.

I went ahead and [opened a "reference branch"](https://github.com/kubeflow/notebooks/compare/notebooks-v2...andyatmiami:kubeflow-notebooks:pr-1385-cache-review-reference?expand=1) that attempts to address the points raised in this PR review just so you can kinda "see" an end state I am envisioning.  A couple notes on this reference branch:
- heavily claude-coded but I reviewed it as I would any contributor's PR - so I stand by the logic
- related to it being claude driven - there is an absurd amount of code comments... but this is moreso to give you context around the motivation/purpose of the deviations from your PR

Now, please note I am **NOT** dictating "use this reference branch" !!! I merely want to leverage this to act as a "discussion tool" and we can collectively discuss and decide what (if anything) makes sense here.

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: From godoc for `golang.org/x/sync/singleflight` 
```
If a duplicate comes in, the duplicate caller waits for the original to complete and receives the same results
```
So it only merges requests that arrive while a fetch is already running. One call to Get(PodMetrics) should take O(ms) to finish ideally, after which every call gets a hit. It seems unlikely that multiple browser tabs, or the future workspace list view would all have the TTL expiring at the exact same second? Even if so, then a single call to Metrics API needs to succeed after which every subsequent call will hit the cache. Also about the "future workspace list view" - that would be a different workspace (or key) for the call right? So singleflight doesnt really help there since singleflight is keyed by cacheKey?

About multiple workspaces:
>  Multiply this across every workspace whose TTL expires within the same wall-clock second 

How likely is this scenario? 

Just worried that this maybe an overkill without an actual reason for making the codebase complex (like dealing with the `context.Canceled` issue you stated) ...

richabanker: > global lock along the hot path.

But the mutex is only used to compare two times, which shouldnt cause contention..? At expiry the probe() does run under the lock, but that's at most once per TTL, and the RESTMapper serves it from [its in-memory cache](https://github.com/kubernetes-sigs/controller-runtime/blob/6ab2188a1fb14a8c4837b418bc0a20049682cde4/pkg/client/apiutil/restmapper.go#L123) when the metrics API is installed. For the case when its not, it hits the API server (which should be once per TTL)

 Also, memoize() tracks 2 things together: 1. the cached result and 2. when it was checked. Sticking to mutex keeps them in sync and makes sure only 1 goroutine runs the probe when the TTL expires (while other concurrent requests wait for the mu. Once the first goroutine finishes, the waiting requests should just be able to use the results from this call that just finished). If we switch to atomics, that could in-fact cause each goroutine to call its own probe() ?

richabanker: But... maybe one fix we should still make is adding a timeout/ctx to the probe() to avoid having a hung API server hold the lock for prolonged time..

richabanker: It looks like even client-go's memcacheclient refreshes its cache under a lock, [ref](https://github.com/kubernetes/kubernetes/blob/e5f62fbfa6d9e9924a4211ef20bbef72db4a104a/staging/src/k8s.io/client-go/discovery/cached/memory/memcache.go#L112-L119), not seeing any use of atomics here... could you point to what you were referring to when you say "This is the same shape as client-go's discovery cache " ?

richabanker: oh I did raise concerns about swallowing the error previously [here](https://github.com/kubeflow/notebooks/pull/1312#discussion_r3833889654) and [here](https://github.com/kubeflow/notebooks/pull/1312#discussion_r3839931655)

I admit I was always in favor of surfacing the error up somewhere atleast in logs. I'll address that along with adding a shorter TTL in case of errors (except in the case when metrics API returns a 404 i.e. when metrics-server is not installed)

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Yeah the idea here was to not cache not-yet-ready results. But of we are already considering a shorter TTL for the err returned when making the call to Metrics Server, maybe we can adopt the same pattern even here.

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Sure I dont have a strong attachment to the cache capacity, can bump it up to 10000

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Hmm several issues if we remove podUID from the cacheKey:

1. Deleted/paused workspaces - if between 2 calls, the workspace gets deleted and we blindly return the cache populated from before, its wrong
2. Deleted and then recreated workspace with same name - returning a stale cache response in this scenario is misleading since the usage metrics can vary between the 2 pods and if the new workspace was created with updated requests/limits, returning a stale cache response in that case is also wrong

richabanker: This one makes sense, I'll bound the PodMetrics `Get` with a 5s timeout. Upon timeout it can degrade to returning requests/limits with no metrics (cached for shorter TTL), and only cancellation of the request context still propagates as an error.

richabanker: ok made the following changes for the probe() function:

Instead of using `(RESTMapper().RESTMapping(...))` I replaced that with:

```go
d.RESTClient().Get().AbsPath("/apis", "metrics.k8s.io", "v1beta1").Do(ctx).Error()
```
wrapped in an independent context with a dedicated 5s timeout (repo.discoveryTimeout). This eliminates unbounded stalls and goroutine accumulation when the API server is wedged or slow to respond.

This did involve plumbing the discovery client from workspaces/backend/internal/repositories/repositories.go into `NewMetricsRepository()`, so lmk if that looks fine

richabanker: Pushed the change to widen memoize's signature to func() (bool, error) so it now accepts `probe func() (bool, error)` and calculates dynamic TTL via `ttlFunc func(error) time.Duration` which decides the TTL for the cache entry by inspecting the err received after the discovery endpoint is queried for "/apis/metrics.k8s.io/v1beta1"

- **err == nil:** API is available; cached for the full apiAvailabilityTTL (60s)
- **apierrors.IsNotFound(err):** Confirmed absent (legitimate steady state for clusters without metrics-server); cached for the full apiAvailabilityTTL (60s) so we avoid re-probing every few seconds
- **Transient errors (timeouts, 5xx, network drops):** Logs an explicit structured warning `(r.logger.Warn("metrics API availability probe failed; will retry after transient cooldown", ...))` and returns a short apiAvailabilityTransientTTL (10s)

I did have to plumb the logger into the metrics repository, which also seemed a bit out-of-scope than what we had discussed before, so appreciate any feedback on that

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Update: pushed the cache capacity to 10000

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Pushed the changes for implementing negative caching with a short TTL. Both fallback paths (when `Get(PodMetrics)` fails/is unavailable, or when `len(podMetrics.Containers) == 0` because metrics scrape is still pending) now cache the fallback `models.NewWorkspaceResourceUsage(pod, nil)` for 5 seconds

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Addressed now in the latest commit - `PodMetrics` fetch from metrics-server is wrapped with a 5s per-call deadline `queryCtx, cancel := context.WithTimeout(ctx, r.queryTimeout)` 

- Because `queryCtx` derives from `ctx`, an early caller cancellation terminates the in-flight request immediately and propagates `ctx.Err()` without caching
- If `queryCtx` times out due to a slow or wedged Metrics Server, the repository degrades gracefully to returning the declared pod requests/limits and stores the fallback in cache with negativeCacheTTL (5s)

andyatmiami: yep, totally agree with your analysis - thanks for entertaining my idea.. but we'll leave this unimplemented for good reason.

andyatmiami: > It looks like even client-go's memcacheclient refreshes its cache under a lock, [ref](https://github.com/kubernetes/kubernetes/blob/e5f62fbfa6d9e9924a4211ef20bbef72db4a104a/staging/src/k8s.io/client-go/discovery/cached/memory/memcache.go#L112-L119), not seeing any use of atomics here... could you point to what you were referring to when you say "This is the same shape as client-go's discovery cache " ?

my humblest apologies - i had been trying to scrape unvalidated/hallucinated statements from my AI-assisted analysis - and I missed this one... inexcusable - i will strive to _be better_ 😇

andyatmiami: Yep, I'll proceed to back off this suggestion - again, thanks for entertaining the thought and providing solid pushback.

andyatmiami: > which also seemed a bit out-of-scope than what we had discussed before

yep, but upon further reflection I feel this is right for the metrics repo...  i did set an agent loose as i reviewed/thought on this to then try to start to craft best practices/guidance about logging in repos
- https://gist.github.com/andyatmiami/759044b25fe598909deab998d89300af
    - ⚠️ not vetted/reviewed by me yet.. this is moreso just a "working copy of notes and thoughts"
    
but that longer term "guidance" wholly out of scope to this PR...  

tl;dr - i'm in agreement with what you did here

andyatmiami: > This did involve plumbing the discovery client from workspaces/backend/internal/repositories/repositories.go into NewMetricsRepository(), so lmk if that looks fine

Yep, this seems reasonable/unavoidable to me in order to provide `ctx` on the call

andyatmiami (approved):
/lgtm
/approve

---

thanks for _bearing with me_ on this one @richabanker while I got a little overzealous in exploring caching patterns 😇 

your decision making was spot on w.r.t. taking and implementing feedback vs. pushing back.

happy with where this code has landed and ready to merge now! 💯 

---

test methodology:
- [PR-1385-TEST-RESULTS.md](https://github.com/user-attachments/files/32346896/PR-1385-TEST-RESULTS.md)
    - as always, agent driven, human reviewed and approved

workspaces/backend/internal/repositories/metrics/repo.go
richabanker: Perhaps worth a separate discussion in the community meeting or adhoc in a separate issue, I'll open one up to address this. Thanks for kicking off the best practices doc right away on this one, I think it will be useful to have!

PR #1388, opened 2026-09-08
fix: block stored XSS via WorkspaceKind SVG assets

christian-heusel:
## Summary

The backend served WorkspaceKind icon/logo SVG assets from a ConfigMap with only a `Content-Type: image/svg+xml` header and no related `Content-Security-Policy`, no `X-Content-Type-Options` or no `Content-Disposition`. The WorkspaceKind details drawer additionally rendered the raw asset URL as a clickable `<a href={...} target="_blank">` link.

Because the asset endpoint is **same-origin** with the dashboard, clicking that link performed a top-level navigation to an `image/svg+xml` document, and any `<script>` embedded in the SVG executed on the dashboard origin **as the victim user**, which results in a stored XSS.

This PR tries to adress this issue at both layers:

- it hardens the backend response headers (the authoritative fix at the source)
- it removes the clickable link that provided the user-interaction vector

## Changes

`workspaces/backend/api/helpers.go` (`WriteSVG`)

Sets three headers on all served SVG assets:

- **`Content-Security-Policy: default-src 'none'; script-src 'none'; sandbox`**: blocks `<script>` execution and external resource loads inside the SVG document, while still allowing it to render as an image.
- **`X-Content-Type-Options: nosniff`**: prevents MIME sniffing to a scriptable type.
- **`Content-Disposition: attachment`**: a direct top-level navigation now downloads the file instead of rendering it as a document (the airtight kill for the click vector).

This does **not** break in-app icon/logo display: the frontend `fetch()`es the bytes and renders them via a data URL, and `fetch()` ignores `Content-Disposition`.

`workspaces/frontend/src/app/pages/WorkspaceKinds/details/WorkspaceKindDetailsOverview.tsx`

Replace both `<a href={...url} target="_blank" rel="noreferrer">` links (icon + logo) with plain `<span title={...url}>` text.

Beyond removing the click prerequisite for the ConfigMap case, this closes a class the backend headers **cannot** cover: the `url` field on a WorkspaceKind asset is returned verbatim and has no scheme validation, so the anchor could render an arbitrary `javascript:` or remote URL that the backend never serves, so `javascript:` in an `href` would execute on the dashboard origin regardless of any response header.

`workspaces/backend/api/workspacekind_assets_handler_test.go`

`assertAssetResponse` now asserts the three protective headers on every
successful SVG response, so both the icon and logo handler test tables verify
them.

## Security impact

- **Class:** Stored Cross-Site Scripting (SVG served as a same-origin document)
- **Preconditions:** attacker needs `workspacekinds` create/update (admin) plus `configmaps` create/update for a labeled ConfigMap; victim must be a workspaces admin and click the link.
- **Effect of fix:** the served SVG can no longer execute script or render as a top-level document, and the clickable link (including the unvalidated-URL path) is removed.

A more nuanced version (given that the preconditions above already grant arbitrary code execution via the used image reference) would only require configmap update access to then use this as part of a privilege escalation strategy to obtain admin credentials.

### DIFF
--- a/workspaces/backend/api/helpers.go
+++ b/workspaces/backend/api/helpers.go
@@ -58,10 +58,28 @@ func (a *App) WriteJSON(w http.ResponseWriter, status int, data any, headers htt
 }
 
 // WriteSVG writes an SVG response with the given status code, content, and headers.
+//
+// SVG is a scriptable document format: when a browser performs a top-level
+// navigation to an "image/svg+xml" response, any <script> inside the SVG
+// executes on this origin. Since the asset bytes come from a user-controlled
+// ConfigMap and are served unsanitized, we set defensive headers so a malicious
+// SVG cannot run as a same-origin document (stored XSS):
+//   - Content-Security-Policy denies everything: no scripts and no external
+//     resource loads (which also closes CSS-based exfiltration). The sandbox
+//     directive additionally forces a unique, script-disabled origin if the SVG
+//     is ever rendered as a document.
+//   - X-Content-Type-Options prevents MIME sniffing to a scriptable type.
+//   - Content-Disposition forces a download instead of inline rendering on
+//     direct navigation. The frontend fetch()es the bytes and renders them via
+//     a data URL, and fetch() ignores Content-Disposition, so in-app display is
+//     unaffected.
 func (a *App) WriteSVG(w http.ResponseWriter, status int, content []byte, headers http.Header) error {
 	maps.Copy(w.Header(), headers)
 
 	w.Header().Set("Content-Type", constants.MediaTypeSVG)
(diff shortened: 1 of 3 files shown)


christian-heusel:
/cc @andyatmiami @thesuperzapper
since this is only tangentially related to the frontend part of the codebase 😊

andyatmiami (approved):
/lgtm
/approve

---

appreciate this security hardening @christian-heusel 💯   While we lose the "hyperlink" on the source type... for now i'd rather error on the side of security and simplicity - so rendering as a `span` ticks both those boxes.  we can always revisit in future if folks disagree with this...

---

had fun playing in tilt and getting some "malicious" svgs involved... can confirm these evil svgs don't trigger any alerts/javascript/deviant browser behavior in the app

Image

but using those same svgs in a trivial "demo.html" page - i could confirm they were trying to do evil things!

Issue #1198, opened 2026-06-18
[TASK]  Correct controller reference management for Workspace owned resources


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

Fix how the Workspace controller manages controller references on owned resources (StatefulSet,
Service, VirtualService) to handle re-adoption and version skew correctly.

Today, if a Workspace is deleted and re-created with the same name, the existing owned resources
retain the old Workspace's controller reference. The reconciler's call to
`SetControllerReference` fails because a different controller is already set. The resources
become unmanageable — updates fail silently and the workspace enters an inconsistent state.

Additionally, the owner reference field indexes in `internal/helper/index.go` match on exact
`APIVersion` string (e.g., `kubeflow.org/v1beta1`). This will break if the CRD is ever promoted
to a new version, since the owner reference on existing resources will still reference the old
API version.

This task addresses three tightly coupled fixes:

#### 1. `ReplaceWorkspaceAsController` helper

Create a helper function that safely replaces the controller reference on an owned resource:
- If the resource is already controlled by the given Workspace, no-op
- If controlled by a different Workspace, remove the old controller reference and set the new one
- If controlled by a non-Workspace resource, return an error (refuse to steal from other controllers)

Location: `internal/helper/helper.go`

#### 2. Apply to all owned resource types in reconciliation

During reconciliation, when the controller finds an existing owned resource (StatefulSet, Service,
VirtualService), call `ReplaceWorkspaceAsController` before updating its fields. This ensures
the current Workspace is always the controller of its resources.

Also fix the `SetControllerReference` error handling for freshly generated resources. Currently,
failure to set the controller reference on a just-generated resource produces a user-visible
`WorkspaceStateError`. This should be an internal error (log + return error) since it indicates
a programming bug, not a user-actionable problem.

Location: `internal/controller/workspace_controller.go`

#### 3. Fix owner reference index matching

Change the owner reference index functions to compare by API group rather than exact API version
string. This ensures the indexes continue to work across CRD version promotions.

Before:
```go
if owner.APIVersion != kubefloworgv1beta1.GroupVersion.String() || owner.Kind != OwnerKindWorkspace {
```

After:
```go
ownerGV, err := schema.ParseGroupVersion(owner.APIVersion)
if err != nil {
    return nil
}
[...]

Issue #1196, opened 2026-06-18
[FEATURE] Workspace Owned Resource Safety


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Motivation

When a Workspace is deleted, it must clean up all resources it owns (StatefulSet, Service,
VirtualService). Kubernetes garbage collection handles this automatically via owner references
and controller references — but only if those references are correctly established and maintained,
and only if users don't bypass the cascade.

Today, two gaps exist:

1. Users can perform an "orphan delete" on a Workspace, which removes the Workspace CR but
   leaves its owned resources running in the cluster with no parent. There is no admission
   control preventing this, so orphaned StatefulSets, Services, and VirtualServices can
   accumulate silently — wasting resources and potentially causing conflicts if a new
   Workspace with the same name is created.

2. The controller does not handle re-adoption of owned resources when the controlling
   Workspace changes. This can happen in several scenarios: a Workspace is deleted and
   re-created with the same name, ownership needs to transfer between Workspaces, or a user
   (or another operator) manually creates a resource that collides with the naming convention
   the controller expects to own. In all cases, the existing resources retain a controller
   reference that doesn't point to the current Workspace. The reconciler cannot update them
   because `SetControllerReference` fails when a different controller already exists. The
   controller should deterministically assert ownership of any resource it expects to manage —
   not silently co-manage it or error out. Additionally, the owner reference index uses exact
   API version matching, which will break across CRD version upgrades.

### Acceptance Criteria

- [ ] Users cannot perform an orphan delete of a Workspace — the operation is rejected with a
      clear error message
- [ ] The controller correctly manages controller references on all owned resource types during
      reconciliation, including re-adoption when the controlling Workspace changes
- [ ] Owner reference matching is robust across CRD API version changes
- [ ] Unit tests cover the ownership and orphan prevention scenarios

siyuanfoundation:
There are a few other gaps we might need to address:
1. there is no finalizer for `Workspace`, so the order of deletion for the child resources are non-deterministic. If user needs to checkpoint before exit, they might lose access to the storage.
2. If an owned `ServiceAccount` is deleted and recreated, the mounted tokens will no longer be valid. Should we guard against this corner case?

PR #1393, opened 2026-09-09
feat: reject orphan deletes of Workspace resources via webhook

### LINKED ISSUE #1197: [TASK] Reject orphan deletes of Workspace resources via webhook
### Certification

### Description

Add validation to the Workspace mutating/validating webhook to reject any update that adds the
`orphan` deletion propagation policy. This prevents users from deleting a Workspace while leaving
its owned resources (StatefulSet, Service, VirtualService) running in the cluster.

Kubernetes supports orphan deletion via two mechanisms:
- Setting `metadata.finalizers` to include `orphan`
- Passing `propagationPolicy=Orphan` in the delete request

The webhook should reject updates that add the `orphan` finalizer. The delete-request propagation
policy may need a separate mechanism (validating admission on DELETE) — the implementer should
investigate which approach fully closes the orphan delete path and document any limitations.

#### Goals

- Webhook validation logic in `internal/webhook/workspace_webhook.go`
- Reject updates that add an `orphan` finalizer to a Workspace
- Clear error message explaining why orphan deletion is not permitted
- Unit tests for the validation logic
- Investigate whether the `propagationPolicy=Orphan` delete-request path also needs coverage

#### Key Files

| File | Changes |
|------|---------|
| `workspaces/controller/internal/webhook/workspace_webhook.go` | Add orphan finalizer rejection |

#### Reference

- Kubernetes orphan deletion: https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/#set-orphan-deletion-policy

### Acceptance Criteria

- [ ] Workspace update that adds an `orphan` finalizer is rejected by the webhook
- [ ] Error message clearly explains that orphan deletion is not permitted for Workspaces
- [ ] Existing legitimate finalizers (e.g., WorkspaceKind finalizer) are not affected
- [ ] Unit tests cover: orphan finalizer rejected, non-orphan finalizer allowed, update without
      finalizer change allowed

gangli113:
Closes #1197
Part of #1196

## Motivation

A `Workspace` owns a `StatefulSet`, a `Service`, and (when Istio is enabled) a
`VirtualService`. These are cleaned up by Kubernetes garbage collection via
`metadata.ownerReferences` when the Workspace is deleted.

An **orphan delete** deliberately skips that cascade, removing the Workspace but
leaving its owned resources running with no parent. This wastes cluster
resources indefinitely and causes collisions if a new Workspace is later created
with the same name, since the controller expects to own resources at those
names. Nothing currently prevents this.

## Changes

Kubernetes exposes orphan deletion through three request shapes. All three are
now rejected:

| Mechanism | Admission op | Validator |
|---|---|---|
| `metadata.finalizers` contains `orphan` | CREATE / UPDATE | `validateNoOrphanFinalizer` |
| `propagationPolicy: Orphan` | DELETE | `validateNoOrphanPropagationPolicy` |
| `orphanDependents: true` (deprecated) | DELETE | `validateNoOrphanPropagationPolicy` |

- **`workspace_webhook.go`** — added the two validators above, following the
  existing `[]*field.Error` convention of the other Workspace validators.
  Rejections are returned via `apierrors.NewInvalid`, consistent with
  `ValidateCreate`/`ValidateUpdate`.
- **`+kubebuilder:webhook` marker** — added `delete` to the verbs, so
  `ValidateDelete` is actually invoked (see the note below on why this is
  required).
- **`manifests.yaml`** — regenerated via `make manifests`; adds `DELETE` to the
  `vworkspace.kb.io` rule.

On UPDATE, the finalizer check only fires when the finalizer is *newly added*,
so a Workspace that somehow already carries it can still be patched to remove
it. Existing legitimate finalizers, including the WorkspaceKind protection
finalizer, are unaffected.

## DELETE coverage

The issue suggested rejecting updates that add an `orphan` finalizer. On its own
this is **not sufficient**.

When a client issues `DELETE ?propagationPolicy=Orphan`, it never writes to
`metadata.finalizers` — the API server's garbage collector adds the `orphan`
finalizer *itself*, internally, as part of processing the delete. No UPDATE
admission request carrying that finalizer is ever generated, so a webhook that
only inspects `metadata.finalizers` on CREATE/UPDATE sees nothing to reject and
the orphan delete succeeds.

This was confirmed against a real API server in `envtest`: at DELETE admission
time the object's finalizer list is empty (`[]string(nil)`), and the `orphan`
finalizer only appears afterwards, written by the API server.

### Why the deprecated `orphanDependents` field is also needed

`orphanDependents` is the pre-1.7 way of requesting orphan deletion. It is
deprecated in favour of `propagationPolicy`, but the API server **still honors
it**, and it is **not** normalized into `propagationPolicy` — it arrives at the
webhook as its own distinct field.

Modern `kubectl` does not expose this field, so it takes a raw API call to
reach — but any client using the dynamic client, raw REST, or an older client
library can set it.

### Commands that trigger an orphan delete

All of the following are now rejected:

```bash
# propagationPolicy=Orphan (single object)
kubectl delete workspace <name> -n <namespace> --cascade=orphan

# propagationPolicy=Orphan (deletecollection)
kubectl delete workspaces --all -n <namespace> --cascade=orphan

# deprecated orphanDependents=true (needs the raw API)
kubectl proxy --port=8001 &
curl -X DELETE \
  -H 'Content-Type: application/json' \
  -d '{"apiVersion":"meta.k8s.io/v1","kind":"DeleteOptions","orphanDependents":true}' \
  'http://127.0.0.1:8001/apis/kubeflow.org/v1beta1/namespaces/<namespace>/workspaces/<name>'
```

`deletecollection` was specifically checked because it was not obvious whether
admission fires per object there; it does, so it is covered by the same
validator.

Normal cascading deletes are unaffected:

```bash
kubectl delete workspace <name> -n <namespace>                        # allowed
kubectl delete workspace <name> -n <namespace> --cascade=background   # allowed
kubectl delete workspace <name> -n <namespace> --cascade=foreground   # allowed
```

### Unit / envtest

Verified in `envtest` against a real API server:

| Scenario | Expected |
|---|---|
| CREATE with `orphan` finalizer | rejected |
| UPDATE adding `orphan` finalizer | rejected |
| UPDATE adding a non-`orphan` finalizer | allowed |
| DELETE with `propagationPolicy=Orphan` | rejected, object still present |
| DELETE with deprecated `orphanDependents=true` | rejected, object still present |
| DELETE with `propagationPolicy=Background` | allowed |
| DELETE with no propagationPolicy | allowed |

```
Ran 7 of 111 Specs -- 7 Passed | 0 Failed
```

Full controller unit suite passes; `golangci-lint` reports 0 issues.

### Manual verification on GKE

Built and deployed the controller to a GKE cluster and exercised every path by
hand. All 11 checks behaved as expected.

Webhook configuration after deploy:

```console
$ kubectl get validatingwebhookconfiguration workspaces-validating-webhook-configuration \
    -o jsonpath='{range .webhooks[?(@.name=="vworkspace.kb.io")]}{.rules[*].operations}{end}'
["CREATE","UPDATE","DELETE"]
```

**1. Create a valid Workspace, and confirm the owned resources exist**

```console
$ kubectl apply -f workspace.yaml
workspace.kubeflow.org/orphan-demo created

$ kubectl get statefulset,service -n orphan-test
NAME                                    READY   AGE
statefulset.apps/ws-orphan-demo-dmbrw   0/1     15s

NAME                           TYPE        CLUSTER-IP      PORT(S)    AGE
service/ws-orphan-demo-8qxg2   ClusterIP   34.118.226.93   8888/TCP   15s

$ kubectl get statefulset -n orphan-test -o jsonpath='{.items[0].metadata.ownerReferences}'
[{"apiVersion":"kubeflow.org/v1beta1","blockOwnerDeletion":true,"controller":true,
  "kind":"Workspace","name":"orphan-demo","uid":"9b530823-..."}]
```

These are the resources an orphan delete would strand.

**2. CREATE with the `orphan` finalizer — rejected**

```console
$ kubectl apply -f workspace-with-orphan-finalizer.yaml
The Workspace "orphan-finalizer-create" is invalid: metadata.finalizers: Invalid
value: "orphan": orphan deletion is not permitted for Workspaces, because it would
leave the resources owned by the Workspace (StatefulSet, Service, VirtualService)
running in the cluster with no parent, use cascading deletion instead
```

**3. UPDATE adding the `orphan` finalizer — rejected**

```console
$ kubectl patch workspace orphan-demo -n orphan-test --type=merge \
    -p '{"metadata":{"finalizers":["orphan"]}}'
The Workspace "orphan-demo" is invalid: metadata.finalizers: Invalid value:
"orphan": orphan deletion is not permitted for Workspaces, ...
```

**4-5. UPDATE with a legitimate finalizer, and a normal spec change — allowed**

```console
$ kubectl patch workspace orphan-demo -n orphan-test --type=merge \
    -p '{"metadata":{"finalizers":["notebooks.kubeflow.org/test-finalizer"]}}'
workspace.kubeflow.org/orphan-demo patched

$ kubectl patch workspace orphan-demo -n orphan-test --type=merge -p '{"spec":{"paused":true}}'
workspace.kubeflow.org/orphan-demo patched
```

**6. DELETE with `--cascade=orphan` — rejected, nothing removed**

```console
$ kubectl delete workspace orphan-demo -n orphan-test --cascade=orphan
The Workspace "orphan-demo" is invalid: propagationPolicy: Invalid value: "Orphan":
orphan deletion is not permitted for Workspaces, ...

$ kubectl get workspace orphan-demo -n orphan-test \
    -o jsonpath='{.metadata.name} deletionTimestamp={.metadata.deletionTimestamp}'
orphan-demo deletionTimestamp=

$ kubectl get statefulset,service -n orphan-test --no-headers
statefulset.apps/ws-orphan-demo-dmbrw   0/1   54s
service/ws-orphan-demo-8qxg2   ClusterIP   34.118.226.93   8888/TCP   54s
```

No `deletionTimestamp`, and both owned resources survived.

**7. DELETE with the deprecated `orphanDependents=true` — rejected**

```console
$ kubectl proxy --port=8001 &
$ curl -s -X DELETE -H 'Content-Type: application/json' \
    -d '{"apiVersion":"meta.k8s.io/v1","kind":"DeleteOptions","orphanDependents":true}' \
    'http://127.0.0.1:8001/apis/kubeflow.org/v1beta1/namespaces/orphan-test/workspaces/orphan-demo'
{
  "kind": "Status",
  "status": "Failure",
  "message": "admission webhook \"vworkspace.kb.io\" denied the request:
    Workspace.kubeflow.org \"orphan-demo\" is invalid: orphanDependents: Invalid
    value: true: orphan deletion is not permitted for Workspaces, ...",
  "reason": "Invalid"
}
```

**8. `deletecollection` with `--cascade=orphan` — rejected**

```console
$ kubectl delete workspaces --all -n orphan-test --cascade=orphan
The Workspace "orphan-demo" is invalid: propagationPolicy: Invalid value: "Orphan":
orphan deletion is not permitted for Workspaces, ...
```

**9. DELETE with `--cascade=background` — allowed, children garbage-collected**

```console
$ kubectl delete workspace orphan-demo -n orphan-test --cascade=background
workspace.kubeflow.org "orphan-demo" deleted from orphan-test namespace

$ kubectl get statefulset,service -n orphan-test --no-headers
No resources found in orphan-test namespace.
```

This is the key non-regression check: normal cascading deletion still works and
the owned resources are cleaned up.

**10-11. DELETE with no flag, and with `--cascade=foreground` — allowed**

```console
$ kubectl delete workspace orphan-demo2 -n orphan-test
workspace.kubeflow.org "orphan-demo2" deleted from orphan-test namespace

$ kubectl delete workspace orphan-demo3 -n orphan-test --cascade=foreground
workspace.kubeflow.org "orphan-demo3" deleted from orphan-test namespace
```

> Note: after a rejected delete the `kubectl get workspace` STATUS column may read
> `Terminating`. That column is `.status.state` (the workspace **pod's** state),
> not Kubernetes deletion — `deletionTimestamp` and `finalizers` were both empty
> in every rejected case above.

## Limitations

- **Webhook availability now gates Workspace deletion.** The webhook uses
  `failurePolicy: Fail`, so extending it to DELETE means Workspaces cannot be
  deleted while the webhook is unreachable. This matches the existing
  `vworkspacekind.kb.io` behavior, which already covers DELETE under the same
  policy — but it is a **new** failure mode for Workspaces specifically.
  Operators recovering a broken cluster may need to remove the webhook
  configuration first. **Flagging this for explicit reviewer sign-off.**
- **Cluster-admin bypass remains possible.** Anyone able to edit or remove the
  `ValidatingWebhookConfiguration`, or write directly to etcd, can bypass this.
  Admission control is a guardrail against accidental misuse, not a security
  boundary.
- **Pre-existing orphans are not remediated.** This change is preventative only.
  Adoption of already-orphaned resources is the scope of #1198.
- **Deletion ordering of owned resources is still non-deterministic**, as the
  Workspace has no finalizer of its own. Raised separately on #1196; not
  addressed here.

### DIFF
--- a/workspaces/controller/internal/webhook/workspace_webhook_test.go
+++ b/workspaces/controller/internal/webhook/workspace_webhook_test.go
@@ -20,7 +20,9 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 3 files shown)


workspaces/controller/internal/webhook/workspace_webhook_test.go
```diff
@@ -299,4 +301,151 @@ var _ = Describe("Workspace Webhook", func() {
...
+		It("should reject creating a Workspace with the `orphan` finalizer", func() {
+			By("creating the Workspace")
+			workspace := NewExampleWorkspace(fmt.Sprintf("%s-create", workspaceName), namespaceName, workspaceKindName)
+			workspace.Finalizers = []string{metav1.FinalizerOrphanDependents}
+			err := k8sClient.Create(ctx, workspace)
+			Expect(err).NotTo(Succeed())
+			Expect(err.Error()).To(ContainSubstring("orphan deletion is not permitted for Workspaces"))
+		})
+
+		It("should reject an update which adds the `orphan` finalizer", func() {
```
siyuanfoundation, line 387: what happens if a ws already has a `orphan` finalizer somehow? Can you add a test case for it?

gangli113: Good catch, thanks! Added a test case covering Workspaces that already have an orphan finalizer (e.g., legacy resources created prior to this PR). So I think updates to these Workspaces should still be permitted, and the test verifies that behavior.

PR #1370, opened 2026-09-02
fix: expose WorkspaceKind filterRules in backend API

thisis-Shitanshu:
## What changed

Expose `filterRules` through the WorkspaceKind GET/PUT API.

The backend now:

- returns `filterRules` from GET
- persists changes through PUT
- supports clearing existing rules

Regression tests cover reading, editing, and clearing `filterRules`.

## Why

This follows up on the `filterRules` issue found while validating #1365.

The field existed in the WorkspaceKind CR, but the backend update model did not expose it, so it was missing from the WorkspaceKind Edit YAML.

Related discussion: https://github.com/kubeflow/notebooks/pull/1365#issuecomment-5496385595

## Frontend follow-up

This PR only contains the backend API change.

After this is merged, the frontend API client can be regenerated against the merged backend SHA, followed by the small frontend change in a separate PR.

Related: #1372

PR #1394, opened 2026-09-10
chore: regenerate frontend API client

thisis-Shitanshu:
## What

Regenerate the frontend API client and WorkspaceKind YAML schema using the merged backend Swagger from #1370.

This also includes accumulated generated API changes since the previous frontend Swagger pin.

## Generated API updates handled

- `filterRules` types/schema from #1370
- ServiceAccount / per-Workspace RBAC contract from #1311
- Workspace pod resources endpoint from #1312
- `statefulSetMetadata` from #1351

The handwritten changes are limited to keeping mocks/tests compatible with the regenerated API:
- updated ServiceAccount mocks to the current `clusterRoles` shape
- verified non-form ServiceAccount config is preserved during WorkspaceKind form conversion
- updated WorkspaceKind mocks for `statefulSetMetadata`
- added mocks for `getWorkspacePodTemplateResources`

No handwritten `filterRules` behavior change is included here. That will be handled separately.

## Validation

- `npm run test:type-check`
- `npm run test:lint`
- `npm run test:unit` — 47 suites, 489 tests passed
- `npm run build`
- `npm run test:cypress-ci` — 22 specs, 446 tests passed
- `git diff --check`


thaorell (approved):
/lgtm

workspaces/frontend/src/shared/mock/mockNotebookApis.ts
```diff
@@ -80,6 +80,11 @@ export const mockNotebookApisImpl = (): NotebookApis => ({
     }),
     getWorkspacePodTemplateLogsBatch: async (_namespace, _workspaceName, query) =>
       buildMockWorkspaceLogs(Math.min(query?.tailLines ?? 20, 50)),
+    getWorkspacePodTemplateResources: async () => ({
```
andyatmiami, line 84: **Nit: extract a  buildMockWorkspaceResourceUsage  helper to match the mock-builder convention**

Every other read on the  workspaces  API in ` mockNotebookApis.ts`  delegates its  data  to a ` buildMock… ` helper in  `mockBuilder.ts ` ( `buildMockWorkspaceDetails` ,  `buildMockWorkspaceLogs` ,  `buildMockWorkspaceCreate` , etc.), so the envelope shape lives in exactly one place. This new  function  inlines ` { containers: {} }` , which drifts from that pattern and — because it's empty — gives no realistic seed for the follow-up UI work that this PR is scaffolding for.

Would you consider adding a builder alongside the others, then referencing it from  `mockNotebookApisImpl` ? Something like:

```
// workspaces/frontend/src/shared/mock/mockBuilder.ts
export const buildMockWorkspaceResourceUsage = (
  usage?: Partial<ResourcesWorkspaceResourceUsage>,
): ResourcesWorkspaceResourceUsage => ({
  containers: {
    main: {
      resources: {
        requests: { cpu: '500m', memory: '512Mi' },
        limits: { cpu: '2', memory: '2Gi' },
      },
      metricsFromMetricsServer: {
        timestamp: new Date().toISOString(),
        usage: { cpu: '750m', memory: '768Mi' },
      },
    },
  },
  ...usage,
});

// workspaces/frontend/src/shared/mock/mockNotebookApis.ts
getWorkspacePodTemplateResources: async () => ({
  data: buildMockWorkspaceResourceUsage(),
}),
```

Benefits:

• Keeps the "one envelope → one builder" convention consistent, so future contract drift only has to be fixed in one file.
• Exercises both the  resources  (configured requests/limits) and  metricsFromMetricsServer  (live sample, optional) branches of  ResourcesContainerResourceUsage  out of the box — which is the exact surface the follow-up work will render against.
• Downstream tests can  buildMockWorkspaceResourceUsage({ containers: {} })  to opt into the empty/pending-metrics case explicitly rather than the mock defaulting to it.

andyatmiami (changes requested):
@thisis-Shitanshu ... really appreciate you "scope-creeping yourself" and taking on this work to reduce tech debt and unblock your `filterRules` UI work..

it pains me to do this - as I am adding a comment on code that is **NOT** directly related to your ultimate goal - but things seen can sometimes then not be "unseen"!

I feel like its easy enough to implement my suggestion - such that the code is more easily ready to be "picked up and adapted" as we follow established conventions in our mock builds _right out the gate_...

but if you oppose this admittedly nit/scope-creepy suggestion.. please speak up and lets discuss!

workspaces/frontend/src/shared/mock/mockNotebookApis.ts
thisis-Shitanshu: Thanks for taking the time to review this and point me toward the existing mock-builder pattern. I went through #1223, #1241, and the frontend work in #1334 to understand how the resource-usage API is intended to be consumed.

I’ve updated this to use reusable resource-usage builders and have `mockNotebookApis` delegate to them, instead of returning an inline empty `containers` object. I also kept the builders composable so follow-up UI tests can represent cases such as unavailable metrics, additional containers, and different resource requirements.

thisis-Shitanshu:
This was a useful opportunity for me to understand how the generated API contract connects to the backend metrics work and the upcoming frontend implementation. Thanks again for the guidance @andyatmiami 🤗

andyatmiami (approved):
/lgtm
/approve

---

relying on @thaorell 's previous `lgtm` as `frontend` confirmation of appropriateness of this change.  in the most recent push - i see the builder pattern properly implemented.

as this is really a "scaffolding PR" to unblock future work by getting updated types in the codebase - the checks being exercised on the PR itself is sufficient to verify behavior.

Issue #1386, opened 2026-09-07
JupyterLab WSK hidden in the default sample setup


christian-heusel:
### Checks

- [x] I have searched the [existing issues](https://github.com/kubeflow/notebooks/issues).
- [x] My issue is related to one of the components in the [`kubeflow/notebooks`](https://github.com/kubeflow/notebooks) repository.

### Kubeflow Notebooks Version

v2.0.0-beta.1

### Kubeflow Platform

KCD

### Kubernetes Distribution

KinD

### Kubernetes Version

```shell
Client Version: v1.36.3
Kustomize Version: v5.8.1
Server Version: v1.35.
```

### Description

When deploying Kubeflow Workspaces version `v2.0.0-beta.1` by default the JupyterLab Workspace is hidden from the WS create flow:

<img width="2560" height="1278" alt="Image" src="https://github.com/user-attachments/assets/a960fd80-83cf-4a4a-84bf-22ec29265ec3" />

This is due to the filterRules that landed with #1357 and they way that our samples are set up:

https://github.com/kubeflow/notebooks/blob/18035691b8099f4bcdbb684c3d4b7ba047ab13c0/workspaces/controller/manifests/kustomize/samples/jupyterlab_v1beta1_workspacekind.yaml#L65-L76ry 

It's quite nice to see the feature working, however I could see that this could confuse people that want to try out the beta release. Also we've been stressing that those are only example configurations, however the JupyterLab sample is our most comprehensive one (showcasing redirects etc.), so we should most likely make that visible by default.

### Relevant Logs

```shell

```

christian-heusel:
cc @andyatmiami & @gangli13 who worked on the feature and @thisis-Shitanshu who was interested in picking this up 😊

thisis-Shitanshu:
First impression is that this looks like a sample configuration issue with a UX impact rather than a regression in #1357. The `WORKSPACE_KIND` filter rule seems to be working as intended, but with the current JupyterLab sample it causes the kind to be hidden in the default setup.

I agree with @christian-heusel that JupyterLab is our most comprehensive sample, so it makes sense for it to be visible by default. Unless something else comes up as we dig into it, I can update `jupyterlab_v1beta1_workspacekind.yaml` to reflect the intended Workspace create experience while keeping the filter-rule example clear.

/assign

christian-heusel:
@thisis-Shitanshu did you already get the chance to raise a PR for this? 😊

[...]

PR #1403, opened 2026-09-14
fix: samples - show JupyterLab WorkspaceKind by default

thisis-Shitanshu:
## Description

Updates the JupyterLab WorkspaceKind sample so it is visible by default in
the Workspace create flow.

The existing `WORKSPACE_KIND` filter used `NotIn`, which also matched the
default namespace when `workspace_team` was absent, causing JupyterLab to be
hidden.

This changes the example to use `In` instead. The filter-rule example remains
active, but becomes opt-in:

- default/unlabeled namespace → JupyterLab is visible
- `workspace_team=team_1` → JupyterLab is hidden

This keeps the sample demonstrating `WORKSPACE_KIND` filtering while making
JupyterLab available in the default sample setup.

closes:  #1386.

## Testing

- [x] `kubectl apply --dry-run=server` succeeds
- [x] JupyterLab is visible with no `workspace_team` namespace label
- [x] JupyterLab is hidden with `workspace_team=team_1`
- [x] JupyterLab becomes visible again after removing the label
- [x] RStudio and VS Code remain visible


christian-heusel (approved):
Looks good and simple enough, verified manually against a tilt cluster 🤗 

/ok-to-test
/lgtm

andyatmiami (approved):
/approve

---

simple, effective, minimal update to demonstrate the power of filter rules while also making our `tilt` environment intuitive _out of the box_

Issue #1371, opened 2026-09-02
Keep frontend generated API client in sync with backend Swagger


thisis-Shitanshu:
### Checks

- [x] I have searched the [existing issues](https://github.com/kubeflow/notebooks/issues).
- [x] My request is related to one of the components in the [`kubeflow/notebooks`](https://github.com/kubeflow/notebooks) repository.

## Problem

While working on #1370, I noticed that the frontend generated API client is still pinned to an older backend Swagger revision through:

`workspaces/frontend/scripts/swagger.version`

When I regenerated the client against the current backend Swagger, it pulled in substantial API drift unrelated to the `filterRules` change, including newer Workspace API endpoints and generated contract changes.

This is similar to #1314, which was split out from #1310 because regeneration also pulled accumulated backend API drift.

## Current behavior

Both sides already have deterministic generation and CI verification:

- the backend generates and verifies `workspaces/backend/openapi/swagger.json`
- the frontend generates its API client from the exact backend commit pinned in `workspaces/frontend/scripts/swagger.version`
- the frontend CI verifies that generated files match that pinned revision

The missing piece is synchronization between them.

Backend Swagger can advance while `swagger.version` remains pinned to an older commit. The drift then accumulates until a frontend feature needs a newer API contract, at which point regeneration can pull unrelated API changes into that feature PR.

## Related

- #1370
- #1314
- #1310

## Proposal

Add a dedicated frontend API synchronization workflow.

### Proposed GitHub workflow

| Stage | What happens | Edge case / boundary |
| --- | --- | --- |
| 1. Detect | Trigger only when `workspaces/backend/openapi/swagger.json` changes on `notebooks-v2` | Normal backend commits that do not change the API do nothing |
| 2. Resolve revision | Use the latest verified `notebooks-v2` commit SHA as the frontend Swagger revision | No LLM or decision engine chooses the SHA |
| 3. Sync pin | Update `workspaces/frontend/scripts/swagger.version` to that revision | We keep the existing SHA-pinning design instead of changing how generation works |
| 4. Generate | Run the existing `npm run generate:api` process | Uses the existing generator; no Dependabot or additional code-generation system is introduced |
| 5. Guard scope | Verify that only `swagger.version` and generated frontend API artifacts changed | If unexpected handwritten files change, the workflow stops |
[...]

PR #1369, opened 2026-09-01
fix: inherit permissions in reusable workflows

christian-heusel:
A reusable workflow may only request permissions its caller already grants, and whether it needs `packages: write` depends on the caller, only `ws-publish.yml` pushes images, and it declares `packages: write` itself.

Fixes #1368

PR #1409, opened 2026-09-15
ci: validate swagger version branch history

thisis-Shitanshu:
## What

Related to [#1371](https://github.com/kubeflow/notebooks/issues/1371#issuecomment-5537313962).

Add a check for updates to `workspaces/frontend/scripts/swagger.version`.

The validator checks that:

- the value is a full Git SHA
- the commit exists
- the commit is in the history of the branch the PR targets

**The workflow only runs when `swagger.version` changes.**

## Why

While looking into frontend/backend API synchronization in #1371, we discussed making sure `swagger.version` cannot reference a commit outside the target branch history.

This is especially useful because a commit may exist in the GitHub repository network without actually belonging to the branch being updated.

## Implementation note

For now this uses `pull_request`, so I see it mainly as a correctness check.

For stronger fork-security enforcement, `pull_request_target` would be preferable so the validation can come from trusted repository code. Since active v2 development is currently on `notebooks-v2` while `main` remains the default branch, I left a TODO to revisit this during the planned migration to `main`.

I also checked the release branches:

- `v2.0-beta-branch` satisfies the ancestry rule
- `v2.0-alpha-branch` has a historical cross-branch Swagger pin

So this check is currently scoped to `notebooks-v2` and `main`.

The validator is kept as a separate script so the same logic can also be reused by the API sync workflow if this approach looks good.

## Validation

- `bash -n`
- `shellcheck`
- `actionlint`
- Prettier
- `git diff --check`
- valid ancestor → passes
- existing out-of-branch commit → fails
- missing/malformed SHA → fails
- invalid target ref → fails


thisis-Shitanshu:
Hi @christian-heusel, I put together the ancestry check inspired by your input on #1371. For now, I kept it on `pull_request` and documented `pull_request_target` as the stronger follow-up around the `main` migration. When you have time, I’d appreciate your thoughts on this direction 🤗

.github/workflows/ws-frontend-swagger-version-test.yml
```diff
@@ -0,0 +1,35 @@
+name: Frontend Swagger Version Check
+
+permissions:
+  contents: read
+
+# TODO: Move this check to pull_request_target after the migration to main.
+# That will let the check use trusted repository code for stronger fork security.
```
christian-heusel, line 7: We shouldn't aim to use `pull_request_target` even after this migration due to the elevated nature in which it runs. See the discussion on 
- https://github.com/kubeflow/community-distribution/pull/3519
- https://github.com/kubeflow/community-distribution/pull/3431

christian-heusel (changes requested):
/ok-to-test

.github/workflows/ws-frontend-swagger-version-test.yml
thisis-Shitanshu: Thanks @christian-heusel, I went through both examples and I think I understand the concern better.

One reason I still wonder whether `pull_request_target` could make sense for this specific check after the `main` migration is that this workflow is intentionally very narrow:

- it only needs `contents: read`
- it does not need secrets or repository writes
- it does not need to build or execute code from the PR
- it only needs to read the proposed `swagger.version` SHA and validate it against trusted branch history

With the current `pull_request` approach, the check works, but the same PR can also modify `swagger.version` and the validator/workflow that checks it. In a large PR, or simply through human error during review, that could make a bad `swagger.version` update easier to miss.

My thinking is that a future `pull_request_target` version could keep the validator/workflow on the trusted base branch and treat the proposed `swagger.version` only as input data. We would explicitly keep permissions read-only and never check out or execute PR-controlled code.

This also feels narrower than the cases in the linked PRs, since we would not be modifying PR state or executing contributed code.

That said, if the project preference is to avoid `pull_request_target` even for this read-only case, I’m happy to remove the TODO and keep the current approach 🤗

PS: The permission changes around #1369 also made me think more explicitly about least privilege, which is why I’m considering whether a read-only `pull_request_target` check (after the future `main` migration) could be sufficiently constrained here.

.github/workflows/ws-frontend-swagger-version-test.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+
+      - name: Fetch target branch
+        run: |
+          git fetch origin \
+            "+refs/heads/${GITHUB_BASE_REF}:refs/remotes/origin/${GITHUB_BASE_REF}"
```
christian-heusel: Do we need this even when we're running `on_pull_request`?

Issue #1411, opened 2026-09-15
Workspace and WorkspaceKind updates fail with 500 on Kubernetes resourceVersion conflicts


amritraj7237:
### Checks

- [x] I have searched the existing issues.
- [x] My issue is related to one of the components in the `kubeflow/notebooks` repository.

### Kubeflow Notebooks Version

dev (notebooks-v2 branch, at commit 1a58debe)

### Kubeflow Platform

Not applicable. This came out of reading the code on notebooks-v2 rather than from a running install, so there's no platform to report.

### Kubernetes Distribution

Not applicable, same reason as above. The behaviour is in the backend's interaction with the API server, so it isn't specific to a distribution.

### Kubernetes Version

```shell
Not applicable. Reproduced against envtest with the 1.31.0 control plane
binaries rather than on a live cluster.
```

### Description

Editing a workspace through the backend API can come back as a 500 even when the request is valid, and the edit is silently dropped. The same thing happens when editing a WorkspaceKind.

I hit this while reading through the update path rather than from a cluster incident, but it's reproducible in tests and the code path is clear, so I wanted to write it up.

`UpdateWorkspace` in `workspaces/backend/internal/repositories/workspaces/repo.go` reads the Workspace through the manager's cached client and then calls `Update` with the `resourceVersion` it just read. In the meantime the controller keeps writing `status`: every activity probe refreshes `status.activity.lastProbe`, and because the Pod watch uses `ResourceVersionChangedPredicate`, any Pod event causes another status write. All of those move `resourceVersion` forward without changing `metadata.generation`.

The revision check the backend uses is generation based (`CalculateRevision` hashes `UID:Name:Generation`), so it looks at the caller's revision, concludes the spec hasn't changed under them, and lets the update proceed. Kubernetes then rejects the write on `resourceVersion`. Nothing in `UpdateWorkspaceHandler` checks for `apierrors.IsConflict`, so it lands on `serverErrorResponse` and the user gets "the server encountered a problem and could not process your request" with their change lost.

What makes it awkward to spot is that no second user is required. The controller's own background status writes are enough on their own, so it shows up as an occasional unexplained 500 rather than something reproducible on demand.

`UpdateWorkspaceKind` has the same structure. There the WorkspaceKind controller recomputes `status.workspaces` and the per-option metrics whenever any Workspace using that kind changes, so on a commonly used kind the object's `resourceVersion` is moving more or less constantly while an admin tries to edit it.

[...]

PR #1412, opened 2026-09-15
fix: retry workspace and workspacekind updates on resourceVersion conflicts

amritraj7237:
closes: #1411

## What's wrong

Updating a workspace through the backend API can fail with a 500 even when the request is completely valid, and the user's edit gets dropped.

`UpdateWorkspace` is a read-modify-write. It reads the Workspace through the manager's cached client and then calls `Update`, which sends back the `resourceVersion` it read. Meanwhile the controller is writing `status` on a regular basis: every activity probe updates `status.activity.lastProbe`, and the Pod watch uses `ResourceVersionChangedPredicate`, so any Pod event triggers another status write. Each of those bumps `resourceVersion` without touching `metadata.generation`.

That distinction is what makes this bite, because the revision check in the repository is built on generation:

```go
revisionInput := fmt.Sprintf("%s:%s:%d", meta.UID, meta.Name, meta.Generation)
```

So the application layer compares the caller's revision, sees the spec hasn't changed underneath them, and correctly lets the update through. Kubernetes then rejects that same write because `resourceVersion` has moved on. The handler has no branch for `apierrors.IsConflict`, so it falls through to `serverErrorResponse` and the user sees "the server encountered a problem and could not process your request".

No second user is needed for this. The controller's own background status writes are enough to trigger it.

There was already a TODO sitting on this code describing the situation, including the note that returning 409 would be wrong since the caller did nothing wrong.

`UpdateWorkspaceKind` has the same shape, with its own TODO. The WorkspaceKind controller recalculates `status.workspaces` and the per-option metrics whenever any Workspace using that kind changes, so a widely used kind has its `resourceVersion` moving constantly while an admin is trying to edit it.

## What changed

Both update paths are now wrapped in `retry.RetryOnConflict`.

The read inside the retry goes through an uncached reader (`mgr.GetAPIReader()`) instead of the cached client. This half is load-bearing rather than cosmetic: retrying against the cache would hand back the same stale object on every attempt, so the loop could never make progress. The cached client is still used everywhere else, including the PVC and secret lookups during validation.

A conflict that survives every retry is returned to the handler unwrapped. It is not mapped to a dedicated status code, so it falls through to 500 as it did before this change. With the retry in place that case is rare, but if you would rather it be an explicit 409 with a message that distinguishes it from the stale revision 409, say the word and I will add it.

The revision check itself is unchanged and still returns 409 when the caller genuinely is working from a stale spec. `ErrWorkspaceRevisionConflict` isn't a Kubernetes conflict, so `RetryOnConflict` returns it straight away instead of retrying it.

## Tests

`internal/repositories/workspaces` had no test file at all before this, so there's a new suite there, plus four more cases on the existing workspacekinds suite. Between them they cover the retry succeeding on a second attempt, an exhausted retry still surfacing as a Kubernetes conflict so the handler can map it, a stale caller revision being rejected with no write attempted, and the read going through the apiReader rather than the cached client.

I checked these actually catch the regression rather than just passing, by reverting each half of the fix on its own. Dropping the retry fails two of them, and pointing the read back at the cached client fails the apiReader one.

Full backend suite is green, including the 153 envtest specs under `api/`.

## Notes for reviewers

- An earlier revision of this PR returned 503 with a `Retry-After` header for an exhausted retry. That was wrong, since the service is not unavailable, and it has been removed following review.
- `apiReader` is threaded through `NewApp` and `NewRepositories`, so both constructor signatures changed. The envtest suite already used an uncached client, so it passes the same one for both.
- This PR is backend only. An earlier revision also touched `apiUtils.ts`, but that change has been reverted and the file is now identical to upstream.


workspaces/frontend/src/shared/api/apiUtils.ts
```diff
@@ -13,6 +13,7 @@ const HTTP_STATUS_MESSAGES: Record<number, string> = {
   415: 'Unsupported media type.',
   422: 'The request could not be processed. Please check your inputs.',
   500: 'An internal server error occurred. Please try again later.',
+  503: 'The resource was modified concurrently. Please try again.',
```
christian-heusel: In this generality this is not correct, since a 503 could happen for a lot of other reasons aswell 🤔

workspaces/backend/api/workspacekinds_handler.go
```diff
@@ -325,6 +325,7 @@ func (a *App) CreateWorkspaceKindHandler(w http.ResponseWriter, r *http.Request,
 //	@Failure		415		{object}	ErrorEnvelope			"Unsupported Media Type. Content-Type header is not correct."
 //	@Failure		422		{object}	ErrorEnvelope			"Unprocessable Entity. Validation error."
 //	@Failure		500		{object}	ErrorEnvelope			"Internal server error. An unexpected error occurred on the server."
+//	@Failure		503		{object}	ErrorEnvelope			"Service Unavailable. A concurrent write prevented the update, retry the request."
```
christian-heusel: A 503 is not the right return code here, since the service is not unavailable 🤔

christian-heusel (changes requested):
Thanks a lot for your contribution! 🤗 

Did you use AI in any capacity while creating these changes? I see that your commits do not carry any `Assisted-By: <...>` / `Co-authored-by: <...>` tags, so I wanted to make sure that you're aware of the [Kubeflow AI Policy](https://www.kubeflow.org/docs/about/ai_policy/). 🤗 🤖

Also did you hit the issue in a "real" test scenario or are you just looking for something to contribute to? 😊

amritraj7237:
thanks for the review and for flagging the policy, should have read it more carefully before openning this one

so yeah i did use AI but just to help track down the bug, the code itself is mine. been going through the notebooks-v2 codebase for teh past couple weeks to get familiar with it for GSoC and the assistant pointed me at the race in the update path after i described the symptom, but i dug into it and wrote the fix and tests myself. i understand the change fully and happy to walk through any part of it. let me know if you want me to add the Assisted-by trailer to the commit and ill ammend and force push

amritraj7237:
second question no real cluster incident. found it by reading the code and then reproduced it under envtest, which is what issue #1411 describes. short version: the backend reads the Workspace through the cached client and then updates it, while the controller keeps writing status in the background. those status writes move resourceVersion but not generation, and the revision check is generation based, so the check passes and the api server still rejects the write. nothing maps IsConflict in the handler so it comes back as a 500 and the edit is lost on the status codes youre right on both and ive fixed them. 503 was wrong, the service isnt unavailable, and putting a concurrency message in the generic status map in apiUtils.ts was worse since any unrelated 503 would show users a missleading reason. dropped the 503 handling entirely and the frontend file is back to unchanged so this is backend only now. with the retry in place an exhausted conflict is rare so it falls through to 500 as before. if youd rather it be an explicit 409 with a message that distinguishes it from the stale revision 409 just say the word and ill add it

christian-heusel (commented):
@amritraj7237 could you rebase your changes? 🤔 It seems like there currently are conflicts that would hinder a potential merge 😅

amritraj7237:
done rebased onto current notebooks-v2 and it shows mergeable again
the conflict was with #1385 the metrics TTL cache work that added a logger param to NewRepositories ok full confession the ai found the bug but i solved it i had added an apiReader param to the same function so we were both appending to the same signature resoltion just keeps both and its one line in app.go i also dropped the merge commmit that was on the branch so its back to a single commit
rebuilt and ran the full backend suite against the new base includng the envtest specs all green
also updated the PR descripton it was still describing the 503 and the apiUtils.ts change from before your review it now says an exhausted conflict falls through to 500 still happy to make that an explicit 409 distinguisable from the stale revision 409 if you prefer just let me know

PR #1414, opened 2026-09-16
fix: handle controller reference re-adoption and version skew for ws owned resources

siyuanfoundation:
closes: #1198
 
Safely replace controller references on owned resources (StatefulSet, Service, VirtualService) to handle re-adoption and CRD version skew correctly:

- Add ReplaceWorkspaceAsController helper in helper.go returning (bool, error) to replace controller reference from an old Workspace or older APIVersion to the current Workspace while rejecting non-Workspace controllers.
- Update owner reference indexing in index.go to match by API group instead of exact APIVersion string, allowing the controller to find owned resources across CRD version upgrades.
- Apply ReplaceWorkspaceAsController to StatefulSet, Service, and VirtualService reconciliation in workspace_controller.go and persist changes when controller reference is updated.
- Treat SetControllerReference failures on freshly generated VirtualService as internal errors rather than user-visible WorkspaceStateError.
- Add unit tests for ReplaceWorkspaceAsController and indexWorkspaceOwner.
- Add e2e test validating re-adoption across CRD version skew and verify owned resources are cleaned up on Workspace cascading delete. (for the e2e test, I did try to patch a different owner uid since orphan delete would not be allowed after https://github.com/kubeflow/notebooks/issues/1197, but then the patched ss was immediately removed by GC)


workspaces/controller/internal/controller/workspace_controller.go
```diff
@@ -478,10 +495,8 @@ func (r *WorkspaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			)
 		}
 		if err := ctrl.SetControllerReference(workspace, virtualsvc, r.Scheme); err != nil {
-			return r.updateWorkspaceState(ctx, log, workspace,
-				kubefloworgv1beta1.WorkspaceStateError,
-				fmt.Sprintf(stateMsgErrorSetControllerReference, "VirtualService", err.Error()),
-			)
+			log.Error(err, "INTERNAL ERROR: unable to set controller reference on VirtualService")
+			return ctrl.Result{}, err
```
andyatmiami, line 501: **Confirm the VirtualService `SetControllerReference` error handling is being aligned intentionally**

This appears intentional and correct — the sibling `SetControllerReference` sites for **StatefulSet** (line 348) and **Service** (line 424) already used the same `log.Error + return` pattern before this PR; VirtualService was the odd one out, and this PR aligns it. 

The call is on a freshly generated in-memory `virtualsvc` whose only realistic failure mode is a missing scheme registration — a program-level bug, not a user-actionable status condition.

workspaces/controller/internal/helper/helper.go
```diff
@@ -238,3 +246,45 @@ func CopyVirtualServiceFields(desired *istiov1.VirtualService, target *istiov1.V
...
+	// We do not use metav1.IsControlledBy(obj, workspace) because the reference should still be updated if only the version is different.
+	if currentController != nil && currentController.UID == workspace.GetUID() && currentController.APIVersion == kubefloworgv1beta1.GroupVersion.String() {
+		return false, nil
+	}
+
+	// the object is not controlled by the given workspace, we need to replace the controller reference
+	if currentController != nil {
+		// fail if the current controller is not a Workspace
+		ownerGV, err := schema.ParseGroupVersion(currentController.APIVersion)
+		if err != nil || (currentController.APIVersion != "" && ownerGV.Group != kubefloworgv1beta1.GroupVersion.Group) || currentController.Kind != OwnerKindWorkspace {
```
andyatmiami: **Consolidate owner-ref-is-a-Workspace matching into a shared predicate**

`ReplaceWorkspaceAsController` and `indexWorkspaceOwner` both individually implement the same "does this owner reference point to a Workspace in our API group?" check, but diverge on how they treat an owner ref with an empty `APIVersion`:

- `indexWorkspaceOwner` refuses it (parsed Group is `""` ≠ `"kubeflow.org"` → skipped, not indexed).
- `ReplaceWorkspaceAsController` adopts it (the `currentController.APIVersion != ""` short-circuit skips the Group comparison, so `Kind: "Workspace"` alone passes → controller reference is replaced).

Empty `APIVersion` is essentially unreachable on a live cluster (the API server rejects it as a required field), so this isn't a live-cluster correctness bug. But 
- (a) the two implementations drifted within a single PR
- (b) the exact code prescribed by issue #1198 for `index.go` matches the strict form (no `!= ""` guard)
- (c) the acceptance criterion "returns an error when the resource is controlled by a non-Workspace resource" argues for refusing unverifiable owner refs, not adopting them.

**Recommended fix:** extract one named predicate and call it from both sites. This collapses the divergence by construction and gives the "match by group, not by exact version" versioning invariant from #1198 a single home for future CRD promotions.

---

(feel free to adopt or not this form of the code - presenting purely to visualize the suggestion)

`index.go`:
```go
// isWorkspaceControllerRef reports whether the given owner reference points to a
// Workspace in the kubeflow.org API group. Version is intentionally ignored so
// this predicate is stable across CRD version promotions (see #1198).
//
// Callers are expected to nil-check ref beforehand.
func isWorkspaceControllerRef(ref *metav1.OwnerReference) bool {
    ownerGV, err := schema.ParseGroupVersion(ref.APIVersion)
    if err != nil {
        return false
    }
    return ownerGV.Group == kubefloworgv1beta1.GroupVersion.Group && ref.Kind == OwnerKindWorkspace
}
```

`indexWorkspaceOwner` becomes:
```go
func indexWorkspaceOwner(rawObj client.Object) []string {
    owner := metav1.GetControllerOf(rawObj)
    if owner == nil || !isWorkspaceControllerRef(owner) {
        return nil
    }
    return []string{owner.Name}
}
```

`ReplaceWorkspaceAsController` becomes:
```go
if currentController != nil {
    if !isWorkspaceControllerRef(currentController) {
        return false, fmt.Errorf("object %s/%s is controlled by %s/%s, which is not a Workspace",
            obj.GetNamespace(), obj.GetName(), currentController.Kind, currentController.Name)
    }
    // ... rest of removal logic
}
```

andyatmiami (changes requested):
@siyuanfoundation 

appreciate how you continue to keep _crankin' out_ solid `controller` improvements..

couple minor things here I flagged...   feel free to discuss if/as you want.

the "are you sure" comment is moreso about just _calling out_ this change - but I agree with how you aligned the behavior and appreciate your attention to detail 💯

christian-heusel (commented):
@siyuanfoundation could you rebase your changes once you apply the next round of review? The PR currently has a merge conflict 😊

workspaces/controller/internal/controller/workspace_controller.go
siyuanfoundation: yes, it was the only one setting the status on controller reference error, which is not really actionable on the user.

workspaces/controller/internal/helper/helper.go
siyuanfoundation: Consolidated in the 2nd commit

andyatmiami (approved):
/lgtm
/approve

---

thanks for yet another solid contribution @siyuanfoundation ...

code looks good and addresses core issue.  i _tried really hard_ to test all the invariants here - but as commented in the PR description - its very challenging!

I did manage to verify on tilt that pre-PR behavior around adoption (leaving stale owner UID) is correctly handled now on this code.  Other invariants I was not able to reproduce on `tilt` - but the tests being added provide solid evidence of the code working as intended.

PR #1417, opened 2026-09-17
fix: show friendly workspace kinds access message

Snehadas2005:
Non-admin users who land on the WorkspaceKind page (e.g. via a direct link, bookmark, or browser back/forward) currently see a broken/error state, because the page fires its data-fetching hooks before checking permissions, and those requests come back as 403s. This PR adds a permission gate to the page itself so that non-admins see a friendly, non-error message instead.

- Split `WorkspaceKinds` into two components: `WorkspaceKindsAdminView` (the existing table, filters, sorting, pagination, delete modal — all unchanged) and a new lightweight `WorkspaceKinds` wrapper that checks `user?.clusterAdmin` first.
- Non-admin (or not-yet-loaded) users now see a PatternFly `EmptyState` with a lock icon and the exact copy from the issue: *"WorkspaceKind management is restricted to administrators. Please contact your admin if you need changes to workspace configurations."*
- Because the check happens before `useWorkspaceKinds()` / `useWorkspaceCountPerKind()` are called, non-admins no longer trigger any workspace-kind API requests at all.
- Added `WorkspaceKinds.spec.tsx` covering: the restricted message renders for non-admins, the data hooks are never called for non-admins, and the same restricted view is shown when `user` is `null` (before user settings have loaded).

**Changes:**
image

**Notes for reviewers**
This supersedes the earlier, stalled PR #1158 for the same issue. The `WorkspaceKinds.tsx` file has grown substantially since that PR was opened (filtering, sorting, pagination, delete modal), so this isn't a rebase of that branch — it's the same approach (gate + split component) re-applied to the current file. The review feedback from that PR (exact copy match, gating the data hooks, covering the `user: null` edge case) is incorporated here from the start.

**Closes:** #1114


christian-heusel (approved):
Looks good to me regarding the end-product, lets see what the frontend reviewers have to say about the underlying code! 🤗

workspaces/frontend/src/app/pages/WorkspaceKinds/__tests__/WorkspaceKinds.spec.tsx
thisis-Shitanshu: The non-admin path is covered well here, including verifying that the data hooks are not called.

Since this wrapper now decides whether `WorkspaceKindsAdminView` mounts, could we also add a small positive-path test for `clusterAdmin: true`?

For example, it could verify that the normal admin path is entered or that the WorkspaceKind data hooks are invoked. That would document both sides of the new gate.

thisis-Shitanshu (commented):
Tested this locally with `clusterAdmin: false`. The restricted state works as intended for `/workspacekinds`, and I confirmed that the WorkspaceKind data requests no longer fire.

I left one small test-coverage suggestion inline.

thisis-Shitanshu:
While reviewing this PR, I noticed a related routing case and wanted to check the intended scope before asking Sneha to change anything.

I reproduced the PR locally with `clusterAdmin: false`.

The new gate works as expected for `/workspacekinds`: the friendly restricted-access state is shown, `WorkspaceKindsAdminView` does not mount, and I confirmed that the WorkspaceKind data requests no longer fire.

However, the other WorkspaceKind management routes remain directly addressable:

- `/workspacekinds/create` still renders the Create Workspace Kind form.
- `/workspacekinds/:kind/edit` still renders `WorkspaceKindForm` and calls `useWorkspaceKindByName(kind)`. With backend authorization enforcing non-admin access, that request could still return a 403 and fall into the existing `LoadError` state.
- need to look into `/workspacekinds/:kind/summary`. 

@paulovmr @thaorell, would you prefer #1417 to remain narrowly scoped to the list page as described by #1114, with Create/Edit handled in a follow-up, or would it make more sense to put the friendly access state behind a shared guard for the WorkspaceKind management routes?

Issue #633, opened 2025-10-02
[TASK] Agree on architecture/implementation details for secrets management APIs


andyatmiami:
### Certification

- [x] I certify I am an Epic Owner for Kubeflow Notebooks 2.0 and expected to create planning-related issues.

### Description

As a community, we need to agree on how the Secrets Management APIs will be designed and implemented.  An initial proposal has been worked on by @andyatmiami and @thaorell and written up as a Markdown file:
-  [nbv2-secrets-management-api.md](https://github.com/user-attachments/files/22892708/nbv2-secrets-management-api.md)

We need to review and revise this document so that the expectations on the Secrets API is well-defined and able to be used as a source of truth for implementation.

### Acceptance Criteria

- [ ] API Proposal for Secrets Management API is agreed upon by community

andyatmiami:
# API Proposal: Kubeflow Notebooks 2.0 Secrets Management APIs

**Tentative release date**: 2025 Q4

**Breaking Change?**: no



## Context

**User goal**:
Kubeflow Notebooks 2.0 users need to create, view, and update Kubernetes Secrets via a simplified REST API that abstracts away Kubernetes-native concepts.


**User Benefits**

* Consistent API contract across all Secret types.
* Sensitive values are never returned in API responses to reduce security risks.
* Clean, frontend-friendly responses that hide unnecessary Kubernetes internals.
* Clear usage of labels and annotations to enable "opting in" existing secrets in the namespace managed outside Notebooks 2.0


## User Experience

The API will provide full lifecycle management for Secrets: create, list, retrieve, and update.

### Create a Secret

**Setup**: A user wants to store credentials for an app in a new Secret.

**Steps**
1. User issues `POST /api/v1/secrets/{namespace}` with the secret contents.
2. Backend validates input, maps API request to k8s Secret schema, and creates a Kubernetes Secret.
3. Backend returns metadata (no sensitive values).

### List Secrets

**Setup**: A user wants to see all Secrets in a namespace.

**Steps**
1. User issues `GET /api/v1/secrets/{namespace}`.
2. Backend retrieves Kubernetes Secrets.
3. Backend returns list with metadata indicating if secret can be mounted on a workspace.

### Update a Secret

**Setup**: A user wants to update an existing Secret with new data.

**Steps**
1. User fetches Secret via GET.
2. User issues `PUT /api/v1/secrets/{namespace}/{name}`.
4. Secret is updated.


## API Summary

### Domain Concepts

[...]

PR #1425, opened 2026-09-22
fix: return 409 on secret update conflict

christian-heusel:
### Description

`PUT /api/v1/secrets/{namespace}/{name}` reads the secret, modifies it and then calls `Update`. If the secret is changed by someone else between the read and the write, the Kubernetes API rejects the `Update` with a `Conflict` error because the `resourceVersion` is stale. The handler had no branch for that error, so it fell through to `serverErrorResponse` and returned `500 Internal Server Error`.

The OpenAPI spec already declares `409 Conflict` for this endpoint, and the other write handlers (delete secret, delete PVC, delete workspace, delete WorkspaceKind) already map `apierrors.IsConflict` to `conflictResponse`. With a 500, clients can't tell a retryable concurrent-modification error from a real server failure.

### Changes

- `api/secrets_handler.go`: in `UpdateSecretHandler`, add an `apierrors.IsConflict(err)` branch that returns `conflictResponse` with the status causes, matching the existing handlers.

The repository (`SecretRepository.UpdateSecret`) already returns the Kubernetes error from `client.Update` unwrapped, so no repository change is needed. The swagger annotations already list 409, so `openapi/` is unchanged.


andyatmiami:
/hold

i think we are missing some context/nuance here I wanna take time to discuss

workspaces/backend/api/secrets_handler.go
```diff
@@ -336,6 +336,11 @@ func (a *App) UpdateSecretHandler(w http.ResponseWriter, r *http.Request, ps htt
 			a.failedValidationResponse(w, r, errMsgKubernetesValidation, nil, causes)
 			return
 		}
+		if apierrors.IsConflict(err) {
+			causes := helper.StatusCausesFromAPIStatus(err)
+			a.conflictResponse(w, r, err, causes)
+			return
+		}
```
christian-heusel, line 343: > Update Error Handling:
>
>    Should return 500 response in event the k8s Update client call fails in the backend due to a conflict (resourceVersion)
>    * cannot deterministically tell if the payload submitted by the user is what caused the conflict
>    * could be changes in labels on the secret, etc
>    * "retry" would simply be handled on the frontend

https://github.com/kubeflow/notebooks/issues/633#issuecomment-3412745787

